PR #75, opened 2016-01-17
Added UseTLS config option for both Admin and Phish servers

wjwoodson:


jordan-wright:
Nice!! I like this. Eventually, we'll probably want to force SSL for the admin server and generate the cert if one isn't given. But, this points us in the right direction.

I'll validate later and will merge.

jordan-wright:
Looks good! Thanks for getting this setup.

PR #87, opened 2016-01-21
Change Table Style & Sortable Columns

### LINKED ISSUE #74: Disable Sorting of Unsortable Columns
The columns that contain action buttons ("Delete", "Edit", etc.) have no need to have sort icons.

This can be done at the instantiation of the table, example from users.js:
```
targets = $("#targetsTable").dataTable({
    columnDefs: [
        { orderable: false, targets: -1 }
    ]
})
```

*Note: this is only if the column is the last one, I believe this can also be accomplished by adding a class to the columns that don't need sorting.

StrangeUSB:
- Changed table style with a couple small tweaks.
- Made the columns that aren't sortable, actions (Edit, Delete, etc.), not sortable.

Closes #74.

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -341,3 +341,34 @@
         float:none !important;
     }
 }
+
+/* Table Styling */
+.modal-content .dataTable tbody td {
+  font-size: 16px;/* Smaller font on modal tables */
+}
+.dataTables_info{
+  font-size: 15px;
+}
+
+/* Sort Icons */
+table.dataTable thead .sorting:after, table.dataTable thead .sorting_asc:after, table.dataTable thead .sorting_desc:after {
+  font-family: 'FontAwesome' !important;
+  position: relative !important;
+  display: initial !important;
+  top: initial!important;
+  right: initial!important;
+  left: 6px;
+  color: #1abc9c;
+}
+table.dataTable thead .sorting:after{
+  content: "\f0dc" !important;
+}
+table.dataTable thead .sorting_asc:after {
+  content: "\f0de" !important;
+  opacity: .8 !important;
+}
+table.dataTable thead .sorting_desc:after {
+  content: "\f0dd" !important;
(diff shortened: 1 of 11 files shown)


jordan-wright:
@StrangeUSB that check is probably there since I didn't know about ```destroy```. It was likely to prevent another initialization of a Datatable.

This looks good! I'll pull down the code and will test it out. If all looks good, I'll get it merged.

Great work!

StrangeUSB:
@jordan-wright Sounds good, double check the attachments table, we can actually probably remove that code now, so we don't do something twice.  I can submit another commit with that change.

StrangeUSB:
@jordan-wright Hold up on merging, small issue with the dismiss() functions.

StrangeUSB:
@jordan-wright should be good to pull down and test again.

jordan-wright:
Ok, will do.

jordan-wright:
@StrangeUSB Everything looks great! Quick question - do we need the ```destroy``` attr on the campaigns.js or dashboard.js?

StrangeUSB:
@jordan-wright No we shouldn't need it, the table isn't reload without a full page reload.  

My only concern could be the delete functions, which I don't have in my branch (should have done a pull), but I believe it should still be ok.  Double check that this function actually exists, https://github.com/gophish/gophish/commit/cff666119e54536ddfc736599c09fb5a9ed45474#diff-a4e69cbcca62aae00025a80410dd8e8dR61, I don't see it in the master branch.  

If that function exist then it will probably need the destroy attribute.

jordan-wright:
Looks like there's some conflicts. Resolving now...

jordan-wright:
Yikes - I ran an autoformat for all my JS on a previous commit. That's making these conflicts a nightmare. Standby - working out the kinks now.

jordan-wright:
Ok, should be good to go. Thanks man!

PR #93, opened 2016-01-21
Remove Column Width on Campaigns Table

StrangeUSB:
Removed column width on campaigns table, looks to fix the issue.


StrangeUSB:
Duplicate, ignore.

PR #169, opened 2016-02-21
Added storage of SMTP settings to database. Closes #78

### LINKED ISSUE #78: Store SMTP servers utilized in campaigns
Add DB table, API endpoints, GUI to manage SMTP server configuration. Update New Campaign modal to select from servers already configured

wjwoodson:
Added storage of SMTP configuration in the database. These can now be managed as independent entities on the 'Sending Profiles' page and are referenced like other objects (email templates, landing pages) when creating a campaign. Closes #78


jordan-wright:
Thanks! I'll give this a look this evening and will begin testing.

templates/campaign_results.html
```diff
@@ -13,8 +13,13 @@
                 </li>
                 <li><a href="/landing_pages">Landing Pages</a>
                 </li>
+                <li><a href="/sending_profiles">Sending Profiles</a>
+                </li>
                 <li><a href="/settings">Settings</a>
                 </li>
+                <li><hr></li>
+                <li><a href="/gophish_user_guide.pdf">User Guide</a>
```
jordan-wright: Eventually, I'm probably going to move these to a generic "Documentation" option with all the docs as links.

templates/sending_profiles.html
```diff
@@ -0,0 +1,143 @@
...
+            </div>
+        </div>
+        <div class="modal-footer">
+            <button type="button" data-dismiss="modal" class="btn btn-default">Cancel</button>
+            <button type="button" class="btn btn-primary" id="sendTestModalSubmit" onclick="sendTestEmail()"><i class="fa fa-envelope"></i> Send</button>
+        </div>
+    </div>
+  </div>
+</div> -->
+<!-- disable sendTestEmail functionality on sending profile page until update handling of /util/send_test_email -->
```
jordan-wright: Do we need to do this before merging?

db/migrations/20160221001538_0.1.2_store-smtp-settings.sql
```diff
@@ -0,0 +1,41 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE campaigns ADD COLUMN smtp_id bigint;
+UPDATE campaigns 
+	SET smtp_id = (SELECT smtp.smtp_id FROM smtp) 
+	WHERE campaigns.id = (
+		SELECT smtp.smtp_id
```
jordan-wright: Any reason this isn't ```SELECT campaigns.id```?

templates/sending_profiles.html
wjwoodson: This will require changes to `func API_Send_Test_Email` in api controller for default handling of empty template & landing page. Your call, can be worked as another issue before 0.1.2 release.

db/migrations/20160221001538_0.1.2_store-smtp-settings.sql
wjwoodson: From what I saw in testing the migrated ids are 1:1 as long as they were created through `/campaigns POST`

wjwoodson:
Working on `API_Send_Test_Email` to handle sending test email on sending profiles page. Hold off on merging, I should be able to get this added tomorrow.

Commits https://github.com/gophish/gophish/commit/ffb14b792757cb06253d84625d4632d5ef14366a https://github.com/gophish/gophish/commit/a3376a702fa7a5232cf2d1c83fde27dbf63d9573 . Broken send_test_email functionality in:
templates/sending_profiles.html
static/js/app/sending_profiles.html
controllers/api.go

jordan-wright:
Sounds good. That'll give me time to test things out a bit more.

jordan-wright:
Hey @wjwoodson, if you're still working with the templates, I made this to give a good example: https://play.golang.org/p/SfdqaZM_5k

You can find similar examples in ```worker.go```.

wjwoodson:
@jordan-wright thanks for the help on this. I think `API_Send_Test_Email` is good to go

jordan-wright:
Awesome!! I'll check it out this afternoon. If it looks good, I'll merge.

jordan-wright:
Looks like when I did the db upgrade, only one of my smtp profiles was migrated over. I know I've deleted a campaign before, so I'm checking to see if somehow that is the reason.

I'll update with more info soon.

jordan-wright:
So, it definitely looks like the deleted campaign was the reason I didn't get all the SMTP settings in there.

I wonder if there's a better way to create this new table? It seems like right now, if someone has say 40 campaigns created, they're going to have 40 SMTP objects with names like "1,2,3,4,5,etc.". That seems like a hassle to deal with.

At the very least, I'd like to name the SMTP profiles with the "host" field so that they're more recognizable. Ideally, I'd also like to resolve duplicates so that instead of 40 SMTP objects that are the same, we can use only the unique ones. Though, I don't have the SQL chops to make that happen easily.

wjwoodson:
@jordan-wright, the SQL should be fixed in  56c9ea0 to 
1. Import all* SMTP records
2. Names are set to 'Imported campaign via {{.Host}} from {{.FromAddress}}'
3. *Dedupe records based on user_id:host:from_address before insert

jordan-wright:
image

This looks *sick*. Great idea.

StrangeUSB:
Fix the color outline :)

jordan-wright:
Yeah, the issue is that the CSS for the outline is applied to the ```input``` element. I'm having difficulty finding the exact rule that is being applied, or else we could apply it either to the whole input group or to the input addon (that button) so that it's applied everywhere.

jordan-wright:
Still doing some final testing, but everything I can see looks good. I'll go ahead and merge.

**Excellent work on this!**

PR #192, opened 2016-03-07
Fix issue when tabbing to file input button

rcutmore:
This fixes a minor display issue when tabbing to a file input button where the button text would disappear. 

Affected buttons:
- _Bulk Import Users_ button when adding user group
- _Add Files_ button when adding email template

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -154,7 +154,6 @@
     right: 0;
     min-width: 100%;
     min-height: 100%;
-    font-size: 999px;
     text-align: right;
     filter: alpha(opacity=0);
     opacity: 0;


jordan-wright:
Interesting - thanks! I had no idea that was occurring. I'll test it this afternoon and will merge if all looks good.

jordan-wright:
Interesting... I couldn't replicate the issue, but hey, removing the font-size doesn't seem to hurt.

rcutmore:
Just a bit of follow up, it looks like this was an issue in Firefox but not Chrome. I'm not sure about Internet Explorer.

PR #229, opened 2016-04-05
Making tracker image fully invisible

ETeissonniere:
At the moment the tracker is a 1x1 image, changed the HTML code to make it fully invisible.

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -78,7 +78,7 @@ func processCampaign(c *models.Campaign) {
 			t,
 			c.URL + "?rid=" + t.RId,
 			c.URL + "/track?rid=" + t.RId,
-			"<img src='" + c.URL + "/track?rid=" + t.RId + "'/>",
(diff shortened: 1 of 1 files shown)


worker/worker.go
```diff
@@ -78,7 +78,7 @@ func processCampaign(c *models.Campaign) {
 			t,
 			c.URL + "?rid=" + t.RId,
 			c.URL + "/track?rid=" + t.RId,
-			"<img src='" + c.URL + "/track?rid=" + t.RId + "'/>",
+			"<img style="display: none" src='" + c.URL + "/track?rid=" + t.RId + "'/>",
```
rcutmore: I think single quotes are needed for the style attribute instead of double quotes: `"",`

jordan-wright: Definitely. You'll see the travis ci build fails with a syntax error when using double quotes. You'll need to use single quotes or escape them with a backslash "\".

DeveloppSoft:
Changed, thanks guys

jordan-wright:
Looks good - thanks @DeveloppSoft!

PR #233, opened 2016-04-06
Clean up templates

rcutmore:
This change makes the logo text for the login and register templates use the same font as other templates. If this was intended feel free to close this PR. It also removes a commented out CSS link from the templates.


jordan-wright:
Looks good - thanks @rcutmore!

PR #284, opened 2016-06-05
Clear input after adding group to campaign

rcutmore:
This is a small change to clear the group input when adding groups to a campaign, making it easier to add multiple groups.


jordan-wright:
Looks good - thanks!

PR #291, opened 2016-06-09
Fix group removal

### LINKED ISSUE #249: Deleting group from campaign
If I add two groups (haven't tested beyond that yet) and choose the trash can to delete one of them both are deleted. Minor but annoying. Note I am running 1.2 on Mac (thanks for fixing the release).

rcutmore:
This fixes #249 by making sure the group removal event handler is only set for a newly added row. Previously when adding a group the event handler would get set for all rows, overwriting the original event handlers for previously added rows.

There is code for copying a campaign [here](https://github.com/gophish/gophish/blob/master/static/js/app/campaigns.js#L205) that still needs to be updated; however, I noticed that it currently isn't working correctly and I don't want to make changes that I can't verify. If these changes look good to you I'll see if I can figure out what's going on with the campaign copy code and update it as well.

### DIFF
--- a/static/js/app/campaigns.js
+++ b/static/js/app/campaigns.js
@@ -302,24 +302,24 @@ $(document).ready(function() {
             errorFlash("Error fetching campaigns")
         })
     $("#groupForm").submit(function() {
-            // Add row to group table.
-            var newRow = groupTable.row.add([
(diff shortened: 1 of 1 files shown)


jordan-wright:
Looks great as always - thanks!

PR #297, opened 2016-06-13
Remove groups from campaign copy

### LINKED ISSUE #295: Campaign copy doesn't include groups
When copying a campaign the groups aren't included. An error is encountered [here](https://github.com/gophish/gophish/blob/master/static/js/app/campaigns.js#L205) when attempting to add groups to the new copy. From what I can see this is because campaign groups aren't currently being tracked in the database. It seems like a *campaign_groups* table would need to be added to the database to track this.

rcutmore:
This fixes #295 by removing the code that attempts to copy groups from a previous campaign to a new one.


jordan-wright:
Looks good to me - thanks!

Issue #18, opened 2015-02-21
Error Updating Target Information


jordan-wright:
Right now, editing groups (```PUT /api/groups/:id```) is implemented by iterating through the given targets to see if they exist in the group. If not, they are created. If there are targets in the group that are not given, they are created.

This does not allow us to actually edit the target details, which is a bug.

There are two implementations I can see happening:

* Using ```jinzhu/gorm``` to handle the many-2-many relationship (new feature added recently)
* Do a naive delete all / insert all operation. Might be a bit more IO, but would make it simple.

jordan-wright:
Moving this to 0.1.2.

PR #315, opened 2016-07-02
Add target update capability to API

rcutmore:
This is an idea for fixing #18.

Currently we check all new targets against existing targets to see if any have been removed. If an existing target is missing from the list of new targets it gets removed from the database. This PR changes _PutGroup_ to update a target's information in the database if it still exists during this check.

Note: the web front end still needs to be updated to allow editing. This just adds the capability to the back end.

Let me know if I missed anything or if there's a different approach you'd prefer.

### DIFF
--- a/models/group.go
+++ b/models/group.go
@@ -156,12 +156,15 @@ func PutGroup(g *Group) error {
 		for _, t := range ts {
 			if t.Email == nt.Email {
 				tExists = true
+				nt.Id = t.Id
 				break
 			}
 		}
-		// If the target is not in the db, we add it
+		// Add target if not in database, otherwise update target information.
 		if !tExists {
 			insertTargetIntoGroup(nt, g.Id)
+		} else {
+			UpdateTarget(nt)
 		}
 	}
 	err = db.Save(g).Error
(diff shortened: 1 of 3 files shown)


models/group.go
```diff
@@ -222,6 +223,20 @@ func insertTargetIntoGroup(t Target, gid int64) error {
 	return nil
 }
 
+// UpdateTarget updates the given target information in the database.
+func UpdateTarget(target Target) error {
+	targetInfo := map[string]interface{}{
```
jordan-wright: Does this need to be a `map[string]interface`? can it just be a `models.Target` instance?

models/models_test.go
```diff
@@ -63,6 +63,31 @@ func (s *ModelsSuite) TestPostGroupNoTargets(c *check.C) {
 	c.Assert(err, check.Equals, ErrNoTargetsSpecified)
 }
 
+func (s *ModelsSuite) TestPutGroup(c *check.C) {
```
jordan-wright: Thanks for adding tests! You're the hero we need, not the one we deserve 😄

jordan-wright:
Looking good! Thanks for taking this on. I've added one note regarding the need for a `map[string]interface`. If it turns out we need that, that's totally cool. I just figured it might work with a `models.Target` instance.

models/group.go
rcutmore: The docs show [here](http://jinzhu.me/gorm/curd.html#update) that a `map[string]interface` must be used to update all changed attributes, even empty ones. I tested it out using a `models.Target` instance and that seems to be the case.

I just added another test for this situation. If you update [this line](https://github.com/rcutmore/gophish/blob/92f4ca6a90197dda9c1289a68656c7b41ace57fc/models/group.go#L233) to `err := db.Model(&target).Where("id = ?", target.Id).Updates(target).Error` the new test will fail.

rcutmore:
I added another test related to updating empty attributes (it seems `map[string]interface` is needed for this case).

Just a side note, it looks like there's no clean up of the test database between tests. I'm new to Go and not familiar with this test framework yet. For the new test I just created a new group with new targets to avoid conflicts with other tests, but I'm not sure if there's any convention regarding clean up between tests to avoid this.

jordan-wright:
Thanks for knocking this out! I'll give it a shot over the next day or so and will merge if all looks good.

You're right - there currently isn't any database cleanup... I wasn't sure the best way to manage that when setting up the tests in the past. If you have any suggestions, let me know!

You're the man!

jordan-wright:
Dude, awesome work again - I dig it. Merged!

PR #320, opened 2016-07-09
Validate target input on web front end

rcutmore:
This cleans up the process for adding/editing targets on the web front end and prevents the user from adding multiple targets for a single email address.

When a user adds a new target the following takes place:
1. Email address is converted to lower case to help prevent duplicates.
2. The target data table is checked to see if an existing row contains given email address.
3. If email address already exists then the row is updated with user input, otherwise a new row is added.

Adding targets via the CSV upload runs through the same process as well.

If you are okay with this approach we may want to take into account email casing on the backend API to prevent issues as well?

### DIFF
--- a/static/js/app/users.js
+++ b/static/js/app/users.js
@@ -171,15 +171,34 @@ $(document).ready(function() {
     // Setup the event listeners
     // Handle manual additions
     $("#targetForm").submit(function() {
-        targets.DataTable()
-            .row.add([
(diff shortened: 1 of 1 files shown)


jordan-wright:
I dig it! Nice work. I'll test everything tomorrow and, if all looks good, I'll merge it in.

Thanks for knocking this out!

jordan-wright:
Merged! Thanks again for knocking this out - great work!

PR #324, opened 2016-07-15
Clean up between model tests

rcutmore:
This is an idea for cleaning up between tests. _gocheck_ offers `TearDownTest` and `TearDownSuite` for cleaning up either at the end of each test or at the end of the entire test suite. I implemented this for the model tests and cleaned up the existing tests a bit.

Some downsides:
- This needs to be set up and managed for each test suite. This PR only does so for _ModelsSuite_. One solution would be to create a separate clean up function that each `TearDownTest` function calls, but I wasn't sure where the best place to put that would be.
- This PR attempts to reset the _users_ table after each test but it doesn't handle the case of a test deleting the default user. I looked into clearing the table and adding the default user back, but the method for resetting the primary key seed was SQLite dependent. I'm sure there's an easy way that I'm not aware of.

### DIFF
--- a/models/models_test.go
+++ b/models/models_test.go
@@ -45,6 +45,15 @@ func (s *ModelsSuite) TestGetUser(c *check.C) {
 	c.Assert(u.Username, check.Equals, "admin")
 }
 
+func (s *ModelsSuite) TestPutUser(c *check.C) {
+	u, err := GetUser(1)
+	u.Username = "admin_changed"
+	err = PutUser(&u)
+	c.Assert(err, check.Equals, nil)
+	u, err = GetUser(1)
+	c.Assert(u.Username, check.Equals, "admin_changed")
+}
+
 func (s *ModelsSuite) TestGeneratedAPIKey(c *check.C) {
 	u, err := GetUser(1)
 	c.Assert(err, check.Equals, nil)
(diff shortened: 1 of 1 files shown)


rcutmore:
I pushed a few more group tests to fill out the test suite a bit more.

jordan-wright:
Hey man!

I dig it! Let me review this and I'll get back to you today or tomorrow.

As always, you're the man!

models/models_test.go
```diff
@@ -63,9 +86,84 @@ func (s *ModelsSuite) TestPostGroupNoTargets(c *check.C) {
...
+	group, err := GetGroupByName("Test Group", 1)
+	c.Assert(err, check.Equals, nil)
+	c.Assert(len(group.Targets), check.Equals, 1)
+	c.Assert(group.Name, check.Equals, "Test Group")
+	c.Assert(group.Targets[0].Email, check.Equals, "test@example.com")
+}
+
+func (s *ModelsSuite) TestGetGroupByNameNoGroups(c *check.C) {
+	group, err := GetGroupByName("Test Group", 1)
+	c.Assert(err, check.Not(check.Equals), nil)
```
jordan-wright: For tests like these, we don't need to do anything after `err != nil`, since Go provides no assurances for the data (like the "group" variable). I would recommend doing a check for something like err = models.WhateverErrorIsReturned to make sure we are getting the correct error.

rcutmore: Thanks, that makes sense.

It looks like `gorm.ErrRecordNotFound` is returned when the group doesn't exist. Do you think we should make a gophish-specific error for this case to return, something like `ErrGroupDoesNotExist`? If not, I'll update these tests to check for `gorm.ErrRecordNotFound`.

rcutmore:
I updated this to check for the expected error in the appropriate tests. If you think we should return a gophish-specific error for this case instead I can push another update.

jordan-wright:
Looks good - merged!

PR #370, opened 2016-09-10
Vulnerability fixes

svigne1:
This pull request comprises of the following
1. DOM - based XSS vulnerability fixes - escapeHTML function was not used in some of the places. I have added it to the places where it was missed.
2. When the project is used under HTTPS by setting "use_tls" property to true in the config.json file, secure flag is set to true for the cookies, and when used under HTTP, it is set to false.
3. Default expiration time for the cookie was 30 days. I have changed it to 5 days.
4. CSRF vulnerability & Insecure Direct object reference - I have duplicated the api routes separately for the browser. And i have modified all javascript files to use the /browser routes instead of /api. This way, the /api routes continue to be csrfexempt, while the /browser routes are now under csrf protection. And since mid.RequireLogin is used by the /browser routes, in contrast to the mid.RequireAPIKey used by the /api routes. The /browser routes deduce the identity of the user from the HTTP request through the "id" value that is set in the cookies rather than from the api_key being passed. This way, Insecure direct object reference also gets mitigated on the dashboard.

^^ Please NOTE that, this change doesn't touch the API layer at all.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -21,6 +21,8 @@ func init() {
 	gob.Register(&models.User{})
 	gob.Register(&models.Flash{})
 	Store.Options.HttpOnly = true
+	// This sets the maxAge to 5 days for all cookies
+	Store.MaxAge(86400 * 5)
 }
 
 // Store contains the session information for the request
--- a/gophish.go
+++ b/gophish.go
@@ -33,6 +33,7 @@ import (
 	"sync"
 
 	"github.com/NYTimes/gziphandler"
+	"github.com/gophish/gophish/auth"
 	"github.com/gophish/gophish/config"
 	"github.com/gophish/gophish/controllers"
 	"github.com/gophish/gophish/models"
@@ -53,6 +54,7 @@ func main() {
 	go func() {
 		defer wg.Done()
 		adminHandler := gziphandler.GzipHandler(controllers.CreateAdminRouter())
+		auth.Store.Options.Secure = config.Conf.AdminConf.UseTLS
 		if config.Conf.AdminConf.UseTLS { // use TLS for Admin web server if available
 			Logger.Printf("Starting admin server at https://%s\n", config.Conf.AdminConf.ListenURL)
 			Logger.Fatal(http.ListenAndServeTLS(config.Conf.AdminConf.ListenURL, config.Conf.AdminConf.CertPath, config.Conf.AdminConf.KeyPath,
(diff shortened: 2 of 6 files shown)


jordan-wright:
I would prefer to not have browser-specific routes if we can avoid it. What threat is this mitigating?

Since the web UI uses the API and requires a valid API key, any CSRF attacks will already depend on the attacker knowing the valid API key. At that point, CSRF doesn't really buy the attacker anything, since they can just hit the API directly. 

The only real threat I can see this protecting is the case where an attacker knows the admin's key, but doesn't have access (perhaps due to FW restrictions) to the dashboard. But, again, at that point, might as well just CSRF the API - not the browser.

Unless I'm missing something, I don't see the need to have the browser endpoints as they add overhead without really getting us anywhere.

I'll test all the other changes, but they look good to me. Just have some patience with me, since I'm working to get gophish up and running for Go 1.7 atm. :smile:

svigne1:
Hi,
That is exactly the case for which i need CSRF protection. We are going to have a firewall that will block the port the dashboard is hosted in. But an attacker can bypass firewalls through CSRF attacks. I thought of CSRF protecting the /api routes directly. But in the existing code CSRF protection was explicitly removed for the /api routes. 

Moreover, if we take the route of CSRF protecting the /api routes using the justinas/nosurf project, then, we need to pass in cookies whenever we want to hit the REST api.( Since the nosurf project always looks for a cookie named "csrf_token"). When hitting the api from the dashboard, we wont face any problems, since the browser takes care of sending the cookies for us. But for users hitting the REST api on their own, they need to write code for managing cookies. And if they are already writing code that reads and sends "csrf_token" cookies form the headers, then they might as well not write code to send the api_keys and can instead pass in the "gophish" cookie set by the gorilla/sessions project.

But the existing code, explicitly used a different auth mechanism for /api routes. It used mid. RequireAPIKey, which didnt look for "gophish" cookies. And used csrfHandler.ExemptGlob function so that the nosurf project didnt look for "csrf_token" cookies. 

And second thing is, these /browser routes which uses mid.RequireLogin also protects against insecure direct object reference, since it looks for the identity in the "gophish" cookie.
So, this was the reasoning for keeping separate routes for the browser. 

Please correct me if i am missing something here!

jordan-wright:
I would like to avoid adding the complexity that would come with trying to protect the API from CSRF attacks for now, since the attacker already needs to know your API key to feasibly perform the attack. Since API keys are generated when gophish is first installed (with no default), I consider this to be sufficient for our purposes.

Let's go ahead and remove the /browser routes for now, since they are only adding complexity without giving us any benefit.

Thanks for your feedback and for working on this!

svigne1:
Okay, I will move my commit one HEAD back

jordan-wright:
Looks good - thanks!

PR #382, opened 2016-09-27
Sending Campaign Emails in Parallel

svigne1:
The current implementation sends emails synchronously.

For every campaign email, 
1. the email content is first created
2. SMTP connection is opened 
3. the mail is sent.
4. then steps 1-3 are repeated again for the next campaign email after 1-3 are completed for the previous email

In this pull request,
Step 1 is put in Go-Routines. (A go-routine is started for each and every email which prepares the content of the mails concurrently)
Step 2 is modified to create a pool of SMTP connections (where every connection is a go-routine waiting on a buffered channel shared between them) where the number of connections can be set through config.json 
And Step 3 is accomplished when go-routines in Step 1 push the prepared object into the buffered channel for consumption by the SMTP pool.

This source code snippet from https://gopkg.in/gomail.v2 was referenced for this pull request
https://godoc.org/gopkg.in/gomail.v2#example-package--Daemon

### DIFF
--- a/config.json
+++ b/config.json
@@ -14,7 +14,8 @@
 	"smtp" : {
 		"host" : "smtp.example.com:25",
 		"user" : "username",
-		"pass" : "password"
+		"pass" : "password",
+		"pool_size" : 5
 	},
 	"db_path" : "gophish.db",
 	"migrations_path" : "db/migrations/"
--- a/config/config.go
+++ b/config/config.go
@@ -11,6 +11,7 @@ type SMTPServer struct {
 	Host     string `json:"host"`
 	User     string `json:"user"`
 	Password string `json:"password"`
+	PoolSize int    `json:"pool_size"`
 }
 
 // AdminServer represents the Admin server configuration details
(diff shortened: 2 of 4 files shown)


jordan-wright:
Hey @svigne1 - thanks for this!

I like this idea in theory, but to be honest right now I'm more concerned with handling proper mail queuing and retrying. I don't want to commit to any direction in the mail flow until I get those parts figured out.

I might move this to another branch so we at least have it available, but it might be a while before this can get moved into mainline.

Thanks again for sending this my way! Good stuff!

svigne1:
@jordan-wright okay, see what fits, and reach out to me if you have any suggestions or modifications that needs to be made in this approach of mail handling! 

From an article i read, i found that RabbitMQ needs to be incorporated instead of buffers, so that the memory doesn't get flooded when high volumes of mails are needed to be sent.

Let me read up more on that, and will add my modifications of this approach in this pull request.

jordan-wright:
Going ahead and closing this one, since I'm finishing up a refactor of our email sending infrastructure that'll wind up clashing pretty hard with this one.

That said, I appreciate you sending this over!

Issue #110, opened 2016-02-01
Add delays between email deliveries


grutz:
It would be nice to be able to spread out a campaign's e-mail delivery to not trip monitoring systems. Sending 100 e-mails, all with very similar payload data, within a 5-10 minute period is usually grounds for instant containment of any future mail.

Some options per campaign:
- Delay rand(x) seconds after group of y mails have been sent
- Delay rand(x) seconds after group of rand(y) mails have been sent
- Delay rand(x) seconds after each mail

jordan-wright:
Hey @grutz - this is a great idea! Let me poke around a bit to see if I can find a good way of implementing this on the back end.

rfdevere:
Seconded! Whilst on the subject further builds could have a randomised email subject (not truly random, from a pre formed list) this will help further limit basic spam email filtering programs.

jordan-wright:
Interesting thought! I'll see what I can come up with. I think right now you can use template values in the subject. So, I can at least do something like "{{.FirstName}} - Your password has been reset" and it will be dynamically generated.

I'll double check on that, and I'll see what I can do about the list of subjects. Thanks for the suggestion!

4144414D:
Ideally a feature such as emailing x number of people between a date range. 

i.e. emailing 100 people at random intervals of the course of a week.

rfdevere:
Winning idea^

mubix:
The randomized subject name should probably be it's own issue so that tracking of the email delay randomization can be better tracked

jordan-wright:
Agreed - I'll make a separate issue for this.

NoahJaehnert:
+1 on this. Currently using a different tool that has the same limitations - no delay/randomization in sending emails. So, we create separate campaigns and schedule them at different intervals/times. However, this makes reporting a nightmare. 

Would be great to have this randomization/delay feature available in GoPhish.

chkconfig:
this feature would be really cool. i'm unsure if we can do this at the moment but x emails to x accounts between x and x dates and in x and x hours would be amazing.

PR #389, opened 2016-10-07
Add random delay between emails sent

nrm21:
Hi,

I also wanted #110 so I decided instead of +1-ing it to a forum post, I thought I would do something about it.  Let me know if you think this is ok to start.

To test it out set use_delay to true in config (otherwise behavior of the program is the same as before).


nrm21:
PR will be redone in branch not named master.

PR #391, opened 2016-10-07
Random delay between sending emails

nrm21:
Hi,

I also wanted #110 so I decided instead of +1-ing it to a forum post, I thought I would do something about it. Let me know if you think this is ok to start.

To test it out set use_delay to true in config (otherwise behavior of the program is the same as before).

### DIFF
--- a/db/migrations/20160605210903_0.2_campaign_scheduling.sql
+++ b/db/migrations/20160605210903_0.2_campaign_scheduling.sql
@@ -2,8 +2,12 @@
 -- +goose Up
 -- SQL in section 'Up' is executed when this migration is applied
 ALTER TABLE "campaigns" ADD COLUMN "launch_date" DATETIME;
+ALTER TABLE "campaigns" ADD COLUMN "smtp_min_delay" VARCHAR(255);
+ALTER TABLE "campaigns" ADD COLUMN "smtp_max_delay" VARCHAR(255);
 
 UPDATE "campaigns" SET "launch_date" = "created_date";
+UPDATE "campaigns" SET "smtp_min_delay" = "1";
+UPDATE "campaigns" SET "smtp_max_delay" = "10";
 -- +goose Down
 -- SQL section 'Down' is executed when this migration is rolled back
 
--- a/db/migrations/20160605210903_0.2_campaign_scheduling.sql
+++ b/db/migrations/20160605210903_0.2_campaign_scheduling.sql
@@ -2,12 +2,8 @@
 -- +goose Up
 -- SQL in section 'Up' is executed when this migration is applied
 ALTER TABLE "campaigns" ADD COLUMN "launch_date" DATETIME;
-ALTER TABLE "campaigns" ADD COLUMN "smtp_min_delay" VARCHAR(255);
-ALTER TABLE "campaigns" ADD COLUMN "smtp_max_delay" VARCHAR(255);
 
 UPDATE "campaigns" SET "launch_date" = "created_date";
-UPDATE "campaigns" SET "smtp_min_delay" = "1";
-UPDATE "campaigns" SET "smtp_max_delay" = "10";
 -- +goose Down
 -- SQL section 'Down' is executed when this migration is rolled back
 
--- a/models/campaign.go
+++ b/models/campaign.go
@@ -25,8 +25,8 @@ type Campaign struct {
 	Events        []Event   `json:"timeline,omitemtpy"`
 	SMTPId        int64     `json:"-"`
 	SMTP          SMTP      `json:"smtp"`
-	SMTPMinDelay  int64     `json:"smtpmindelay"`
-	SMTPMaxDelay  int64     `json:"smtpmaxdelay"`
+	SMTPMinDelay  int       `json:"smtpmindelay"`
+	SMTPMaxDelay  int       `json:"smtpmaxdelay"`
 	URL           string    `json:"url"`
 }
 
--- a/models/campaign.go
+++ b/models/campaign.go
@@ -25,8 +25,8 @@ type Campaign struct {
 	Events        []Event   `json:"timeline,omitemtpy"`
 	SMTPId        int64     `json:"-"`
 	SMTP          SMTP      `json:"smtp"`
-	SMTPMinDelay  int       `json:"smtpmindelay"`
-	SMTPMaxDelay  int       `json:"smtpmaxdelay"`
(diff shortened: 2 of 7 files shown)


nrm21:
Actually... after reading the other PR request on "sending emails in parallel" I realize that this PR might make that PR even more necessary for those who are running multiple campaigns.  This is a blocking function (that's sort of the point of delay isn't it) that will only further delay and tie up the sending of emails depending on the langth of time we want to wait for.

Not an issue for me as I like to only run one campaign at a time but something to think about for others.

jordan-wright:
Hey @nrm21, thanks for sending this my way! So, with regards to the need for parallel email sending, keep in mind that every campaign is executed in its own goroutine right now.

As for the functionality, I'd like to see this configured per campaign or maybe per sending profile. Since we don't use the smtp entry in the config for anything right now, we'll want to avoid using it for this.

I think with some fiddling we can get this in a position to be merged. Thanks again for all your hard work on this!

nrm21:
Ok this is what I have so far but I might need your help getting the fields created in SQLLite.  The ALTER TABLE commands in the SQL file are ineffectual.

And TBH I'm not even really sure how you are doing it.  But when I created the fields in the campaigns.html it complained that "the campaigns table does not contain a field named ..."  So I suspect there's code that is scouring the page for field ID's and tries to automagically insert them into the SQL and since I wasn't able to make the column name it complains.

I did however download an SQLLiteeditor and add those columns manually just to see if it would complain and sure enough it fixed that.  So I know that's (partly at least) whats still wrong here.

jordan-wright:
Generally we handle database changes via migration files created with goose. Some detailed info can be found here: https://getgophish.com/blog/post/database-migrations-in-go/

worker/worker.go
```diff
@@ -217,6 +222,24 @@ func processCampaign(c *models.Campaign) {
 	}
 }
 
+// Creates a simple constant delay between sending each email message.  Later
```
jordan-wright: Golang code style suggests to use the function name to start the code documentation. So something like:
"SendingDelay creates a simple...". Also, go ahead and trim out the "Later..." since we don't want this in the official docs that are rendered. You can add this to the commit message.

templates/campaigns.html
```diff
@@ -83,6 +83,10 @@ <h4 class="modal-title" id="campaignModalLabel">New Campaign</h4>
               <input type="text" class="form-control" placeholder="http://192.168.1.1" id="url"/>
 	      <label class="control-label" for="url">Schedule: </label>
               <input type="text" class="form-control" id="launch_date"/>
+          <label class="control-label" for="senddelay">Min delay time: <i class="fa fa-question-circle" data-toggle="tooltip" data-placement="right" title="Minimum randomized delay time between sending of each campaign email"></i></label>
```
jordan-wright: I'd like the delay to be optional. We can have a checkbox that says something like "Randomize Sending Delay" or something that sounds better, heh.

There is code elsewhere (on mobile or I'd link it) that shows how to handle showing options only after a checkbox is shown.

jordan-wright: Also, I don't see you actually sending these - you'll need to modify campaigns.js to include grabbing these and sending them to the server.

models/campaign.go
```diff
@@ -255,6 +257,12 @@ func PostCampaign(c *Campaign, uid int64) error {
 	if c.LaunchDate.IsZero() {
 		c.LaunchDate = time.Now()
 	}
+	if c.SMTPMinDelay == 0 {
+		c.SMTPMinDelay = 1 // default to 1 second
+	}
+	if c.SMTPMaxDelay == 0 {
+		c.SMTPMaxDelay = 10 // default to 10 seconds (most wont want the delay feature anyway)
```
jordan-wright: We shouldnt introduce a default delay (see my other comments). Most people will want to have the emails streamed asap.

jordan-wright (changes requested):
Just an initial first glance.

db/migrations/20160605210903_0.2_campaign_scheduling.sql
```diff
@@ -2,8 +2,12 @@
 -- +goose Up
 -- SQL in section 'Up' is executed when this migration is applied
 ALTER TABLE "campaigns" ADD COLUMN "launch_date" DATETIME;
+ALTER TABLE "campaigns" ADD COLUMN "smtp_min_delay" VARCHAR(255);
```
nrm21: Done.

models/campaign.go
nrm21: Done, well... I set the min to 0 sec and the max to 1 sec (they cant be the same or the rand func will panic)

templates/campaigns.html
nrm21: Think, I'm going to figure this out another day... it's getting late now and I have to be somewhere tomorrow.  Maybe early next week or something.

nrm21: OK I may need some help here.  When I created the lines:
`
smtp_min_delay: $("#smtp_min_delay").val(),
smtp_max_delay: $("#smtp_max_delay").val(),
`

In the campaigns = {} structure in campaigns.js, then when I press the launch button I get:

Invalid JSON Structure

So I have traced that (I think) to the function 'API_Campaigns' in api.go but don't know what to do from there.  I have already defined the smtp delays in the campaign struct in go so I don't know why it complains.

Dorsidhion:
Hey nrm21, 

I was just wondering if this had been integrated as of yet? It's something that I'm after using in one of my campaigns. 

Cheers.

nrm21:
It hasn't, and TBH I sort of forgot it for a while.  I was fixing it up and ran into some trouble I couldn't solve and I think my Oct 11th response was part of that trouble.  I'm going to have to reread the code and my contrib code to refresh my memory because I kinda moved onto other features in other FOSS projects I'm working on.

Also at my job we haven't run a campaign in a month now and probably won't for another month or so (my company is small enough they only do periodic phishing campaigns not continuous).  Maybe this weekend I might get some time to code if there's not lots of chores around the house.

jordan-wright:
Hi everyone,

This is an older pull request, and since this was originally posted we've added the ability to specify a delay between emails. Making a random delay will be possible and will build on the same code that we used to add the initial delay.

Since things are in good shape, I'm going to close out this pull request. I do appreciate y'all getting the ball rolling and suggesting ideas on how this could have been done!

PR #394, opened 2016-10-11
Generate self-signed certificates for administration interface

ss23:
The documentation mentioned that at some point this might be automated, so I wrote an implementation.

I assume this is better placed in a controller, but I wasn't sure where was best (especially since it's not likely to be used anywhere else in the code), so I have just written it here for now.

ECDSA could easily be changed to RSA if requested.

I have refrained from writing similar code for the phishing interface as it is most likely going to not be productive to have a non-valid certificate on that interface.

### DIFF
--- a/gophish.go
+++ b/gophish.go
@@ -26,11 +26,20 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"errors"
 	"fmt"
 	"log"
+	"math/big"
 	"net/http"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/NYTimes/gziphandler"
 	"github.com/gophish/gophish/auth"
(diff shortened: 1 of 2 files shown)


jordan-wright:
This is an excellent idea - thanks, @ss23!

I'm finalizing a binary release atm (should be released within a couple of weeks or so), so if you don't mind I'm going to get this officially reviewed and merged shortly after that.

Thanks for sending this our way!

gophish.go
```diff
@@ -79,3 +92,70 @@ func main() {
...
+		return nil
+	}
+	if _, err := os.Stat(KeyPath); !os.IsNotExist(err) {
+		return nil
+	}
+
+	Logger.Printf("Creating new self-signed certificates for administration interface...\n")
+
+	// RSA would be a perfectly fine alternative
+	priv, err := ecdsa.GenerateKey(elliptic.P384(), rand.Reader)
```
jordan-wright: I'm admittedly not a crypto expert by any means, but any particular reason we chose `P384` as opposed to `p-521`?

jordan-wright (commented):
This is looking good! I like the idea a ton.

I've left a couple of comments and still have a bit of the code to check through. If you want to check out the comments, that would be awesome. Worst case, I can just merge this into a branch and handle some of it myself.

Thanks again for looking into this!

jordan-wright:
I have some spare time, so I'm going to go ahead and merge this to a branch and I'll clean it up based on the comments I left. 😄 

This is seriously good work and much appreciated. Thanks for this, @ss23!

PR #396, opened 2016-10-12
Support for sending emails with custom headers

### LINKED ISSUE #128: Support Custom Mail Headers
Mail headers are supported by the github.com/jordan-wright/email library, but aren't supported by gophish.

We can add these as a serialized field for templates.

ss23:
Adds an option during email template creation to specify any additional and/or custom email headers to be appended to the sent email.
These are parsed using the templating system to allow for variables to be used.

Closes #128

Example UI:
image

I'm not happy with the UI yet, so if anyone has improvements, just suggest them and I'll be happy to implement.

### DIFF
--- a/db/migrations/20161013000001_custom_email_headers.sql
+++ b/db/migrations/20161013000001_custom_email_headers.sql
@@ -1,0 +1,12 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+CREATE TABLE IF NOT EXISTS custom_headers(
+	id integer primary key autoincrement,
+	key varchar(255),
+	value varchar(255),
+	"template_id" bigint
+);
+-- +goose Down
+-- SQL section 'Down' is executed when this migration is rolled back
+DROP TABLE custom_headers;
--- a/models/campaign.go
+++ b/models/campaign.go
@@ -145,6 +145,11 @@ func (c *Campaign) getDetails() error {
 		Logger.Println(err)
 		return err
 	}
+	err = db.Where("template_id=?", c.Template.Id).Find(&c.Template.CustomHeaders).Error
+	if err != nil && err != gorm.ErrRecordNotFound {
+		Logger.Println(err)
+		return err
+	}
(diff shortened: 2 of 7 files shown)


jordan-wright:
This is a great idea, and a request that people have wanted for quite a while.

As for the UI, what do you think about doing key/val pairs? We could then have a dynamic "Add" button that adds a new row. Something similar can be found in adding members to a group.

ss23:
I've run into a minor implementation detail which I'm going to elaborate on here for the record:

The RFC as far as I can see does not define explicitly that mail headers have to be unique - [RFC 2822 Section 2.2](https://tools.ietf.org/html/rfc2822#section-2.2) - however [gomail has implemented headers as a map](https://github.com/go-gomail/gomail/blob/master/message.go#L103), thus only allowing a single header to be set per email message.

I imagine it's not likely this is going to be an issue for many people anyway, but in the interest of following the RFC closely, and ensuring we have an easy path to fix this if it ever becomes an issue in the future, I am implementing it such that you can have duplicate header keys.
If this does become an issue, adding some extra validation should be a simple job.

ss23:
I have updated the PR with a new UI

gophish-custom-headers-ux

Note this is a substantial change to the codebase compared to the original PR, and it has a lot of increased complexity because of this.
If this is merged, it would be a good idea to refactor how the Template's are instantiated, to avoid so much duplicated code such as in this PR.

jordan-wright:
Hey @ss23!

I want to apologize for taking so long on getting back to you - this is looking really good! I'll be creating a branch for this in the next couple of days and will merge this in. I'll then do exactly what you mentioned and refactor some of the ways template details are loaded to be more DRY-esque.

Thanks again for sending this over! This is exactly how I want this feature to look. Great work!

c0om0odo:
Any news on this one?

jordan-wright:
Merged into a custom branch. I'll clean up the code a bit, adjust the SQL migrations to match our new format (plus add a MySQL migration) and will merge into master.

Hopefully we'll be good to go today!

Thanks again for all your work on this, @ss23! I really like the way you approached this.

PR #414, opened 2016-10-27
Saving,Updating & Deleting Groups 

svigne1:
Since, the current implementation doesn't delete the users within a group when the group is deleted.. The users table (targets) over time starts to collect lots of junk data.

For example, Lets say, I create a group called A, and i misspelled a user's email id in it, and i by mistake save it.
Now, when i go and edit the group A, the misspelt user is not deleted. Instead, the new user with the correct email id is created. And when the group is deleted. Both the misspelt and correctly spelt users remain in the database.

And there is no way by which i can delete the misspelt user therefore the current approach leading towards a resource leak.
To avoid this, I think it is best to delete the users within a group whenever the group is deleted.( And so the changes in DeleteGroup function in this pull request)

And if this approach is taken up.. Then, While saving the group, in insertTargetIntoGroup function, the code that checks for whether or not an user already exists in the database can be removed, which will make inserting a group very fast.

Also, the criteria that all users in a given group should be unique can be achieved by sorting the array of users in the group locally in go and omitting the call for insertTargetIntoGroup function for all the duplicate entries. Also for efficiency each call to insertTargetIntoGroup can be put in go-routines.
But the current implementation has only one db connection opened as shown in the following line https://github.com/gophish/gophish/blob/master/models/models.go#L91, which i have changed by adding a db_pool variable in the config.json file, where you can modify the maximum number of db connections you want.

And for PutGroup function, taking all existing entries in the db and running a nested for loop against the new entries from UI slows down the process to an efficiency of O(n^2). Instead, deleting all rows of the group in db and adding them all again will give a performance of O(n) 

All these changes helps saving, updating and deleting very large datasets fast.

### DIFF
--- a/config.json
+++ b/config.json
@@ -17,5 +17,6 @@
 		"pass" : "password"
 	},
 	"db_path" : "gophish.db",
+	"db_pool" : 10,
 	"migrations_path" : "db/migrations/"
 }
--- a/config/config.go
+++ b/config/config.go
@@ -35,6 +35,7 @@ type Config struct {
 	PhishConf      PhishServer `json:"phish_server"`
 	SMTPConf       SMTPServer  `json:"smtp"`
 	DBPath         string      `json:"db_path"`
+	DBPool         int         `json:"db_pool"`
 	MigrationsPath string      `json:"migrations_path"`
 }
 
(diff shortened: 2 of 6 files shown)


svigne1:
I have modified TestPutGroup and TestPutGroupEmptyAttribute functions..
Because, Since i am using Go-routines and creating the targets again(during update). The resultant targets array may not contain the data in the given order. so, target[0] before update and after update may not be the same. In other words, the target array gets shuffled after update.

PR #421, opened 2016-11-08
Users&Groups Selective Processing

svigne1:
In Users&Groups, when a huge CSV file is being uploaded (lets say it has 50,000 records ),
then according to the following line https://github.com/gophish/gophish/blob/master/static/js/app/users.js#L97 , addTarget function is being called for each of those entries.

And This line in addTarget function https://github.com/gophish/gophish/blob/master/static/js/app/users.js#L143 , draws the table again and again, after adding each entry into the table.

This slows down the performance considerably. The browser hangs and we get alerts popping up asking for confirmation to kill the script.

In DataTables, there is an option called "server side processing"
https://datatables.net/examples/data_sources/server_side.html 

In this kind of processing, the datatable always stores only one page in the FrontEnd. Anytime a new page is requested, The server is polled. And the dataTable is updated with new results.

^^ In our case, the data is available only in the FrontEnd.
So, I have created a dummy API called /api/datatables/dummy, which when polled, always returns an empty Json.
And in the FrontEnd, I enabled server side processing on the datatable, and pointed the url to the dummy API. Upon receiving back this meaningless json in the FrontEnd, the ajax.dataSrc function is triggered. 
https://datatables.net/reference/option/ajax.dataSrc
In which, I add the local json i have stored in the global variable called PostData, to the server's Json
https://github.com/svigne1/gophish/blob/groups_ui/static/js/app/users.js#L100

This way, Selective processing of Huge DataSets, is accomplished on the frontend.

Additional functions like 
https://datatables.net/reference/option/infoCallback (to render the description of the table)

and removeDuplicates() which sorts PostData whenever its updated, and removes the duplicate items have been used
https://github.com/svigne1/gophish/blob/groups_ui/static/js/app/users.js#L53

Awaiting your feedback

### DIFF
--- a/controllers/api.go
+++ b/controllers/api.go
@@ -580,6 +580,10 @@ func API_Import_Site(w http.ResponseWriter, r *http.Request) {
 	return
 }
 
+func API_Datatable_Dummy(w http.ResponseWriter, r *http.Request) {
+	JSONResponse(w, models.Response{Success: true, Message: "Empty Json"}, http.StatusOK)
+}
+
 // API_Send_Test_Email sends a test email using the template name
 // and Target given.
 func API_Send_Test_Email(w http.ResponseWriter, r *http.Request) {
--- a/controllers/route.go
+++ b/controllers/route.go
@@ -63,6 +63,7 @@ func CreateAdminRouter() http.Handler {
 	api.HandleFunc("/import/group", API_Import_Group)
 	api.HandleFunc("/import/email", API_Import_Email)
 	api.HandleFunc("/import/site", API_Import_Site)
+	api.HandleFunc("/datatables/dummy", API_Datatable_Dummy)
 
 	// Setup static file serving
 	router.PathPrefix("/").Handler(http.FileServer(http.Dir("./static/")))
--- a/static/js/app/users.js
+++ b/static/js/app/users.js
@@ -1,15 +1,17 @@
 var groups = []
-var PostData = []
+var LocalData = []
+var Data_View = []
 var targetsTable ;
 var tableInfo = {};
 tableInfo.start=0;
-tableInfo.order=0;
+tableInfo.order=[{"column":0,"dir":"asc"}];
+tableInfo.search={"regex":false,"value":""};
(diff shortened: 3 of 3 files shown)


jordan-wright:
Ah, I should've looked at this earlier. I just made a small improvement by moving the draw operation out of the `addTarget` method so that it is now called after the targets are `add`ed to the table.

I like the idea of doing server-side processing and that's a really clever way to do it. The other option would be to leverage something like [this](https://datatables.net/extensions/scroller/examples/initialisation/large_js_source.html). What would be your thoughts on that? My concern with the fake server side processing would be handling things like sorting and searching. Do you know if this implementation supports that?

svigne1:
Hi Jordan, 

You have missed to include that improvement [here](https://github.com/gophish/gophish/blob/master/static/js/app/users.js#L79).. You could call `targets.DataTable().draw()` after line 80 and not inside the loop.

As for the client-side data source [approach](https://datatables.net/extensions/scroller/examples/initialisation/large_js_source.html),
I tried implementing it. It takes 20 seconds for 50k records to load. Whereas in my approach, it takes less than a second to load 100 thousand records.
The reason being, even in client-data-source approach, all the 50k records are required to be added to the table. And this loop iterating over 50k records and calling add function on the table takes time.Whereas, in server side processing, only 10 records are added to the table, and only 10 records are rendered for display, which boosts performance considerably.

So, Fake server side processing is the faster approach of the two.
I will add the searching and ordering capabilities soon, so that you will be able to merge it.
Awaiting your Feedback.

svigne1:
Added Ordering Capabilities.
Also `infoCallback` is not required. The default Description works fine.
I will add search capabilities soon

svigne1:
Hi Jordan,
I have added the Search Capability also. Now, the local server side processing, has both ordering and search capabilities. Let me look into this [approach](https://datatables.net/extensions/scroller/examples/initialisation/large_js_source.html) one more time,
In my local setup, the performance was very low. But the demo shown by datatables is loading pretty fast. Will get back to you after evaluating it again

PR #446, opened 2016-11-21
Client-Side Compression

svigne1:
Hi,
When Importing a huge CSV file via "Bulk Import CSV" option in Users&Groups, It takes a long time to upload it to the server.
To reduce network latency, zipping the files and uploading takes lesser time.
For example, a CSV file with 100 thousand records has size 5.3MB.
But when Compressed, it has size 137KB.

So, in this pull request, I am compressing the CSV file in the frontend using [zip.js](http://gildas-lormeau.github.io/zip.js/index.html) and decompressing the zip file in go using [archive/zip](https://golang.org/pkg/archive/zip/#example_Reader)

So, i have added 4 files :- zip.js, z-worker.js, deflate.js and inflate.js to the static/js/zip folder.
And edited the `ParseCsv` function in util.go, to unzip the files before parsing them.
This reduces latency considerably.. 
Hope this helps.

### DIFF
--- a/static/js/app/users.js
+++ b/static/js/app/users.js
@@ -1,5 +1,35 @@
 var groups = []
 
+function zipBlob(filename, blob, callback) {
+  // use a zip.BlobWriter object to write zipped data into a Blob object
+  zip.createWriter(new zip.BlobWriter("application/zip"), function(zipWriter) {
+    // use a BlobReader object to read the data stored into blob variable
+    zipWriter.add(filename, new zip.BlobReader(blob), function() {
+      // close the writer and calls callback function
+      zipWriter.close(callback);
+    });
+  }, onerror);
+}
+
+function unzipBlob(blob, callback) {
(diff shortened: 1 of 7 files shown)


svigne1:
This Pull Request is not needed. #446 solves the problem instead.

jordan-wright:
Hey @svigne1!

Sorry it took me so long to get back to you on some of these. I've been working on some of the more stability-related things with Gophish recently.

To be honest, I'm not sure this is an approach I'm interested in taking with Gophish. While there are some bandwidth gains that become apparent for very large CSV files, most people won't be uploading lists consisting of hundreds of thousands of records. Even then, the size of the CSV isn't our biggest problem 😄 

This also adds another step that would make it difficult to use this in the API. It would require users to upload ZIP files of their CSV's instead of just the CSV itself.

With all that in mind, I'm going to go ahead and close this one out for now. I appreciate you sending ideas like this to us. Keep the thoughts coming, and keep up the great work!

PR #457, opened 2016-11-29
Gzip Best Compression

svigne1:
Hi,
The current code in `Gophish.go` uses [GzipHandler](https://github.com/gophish/gophish/blob/master/gophish.go#L57) which passes [DefaultCompression](https://github.com/NYTimes/gziphandler/blob/master/gzip.go#L178) to `NewGzipLevelHandler` by default.

I have copy pasted the two lines inside `GzipHandler` and i am passing `BestCompression` instead of `DefaultCompression`.

This increases performance considerably. I tried inserting a csv with 5 lakh entries in it. It took 1.1 minutes to get back the Json response form the server.
But now with this change, it takes only 5.94 seconds.

Hope this helps.

### DIFF
--- a/gophish.go
+++ b/gophish.go
@@ -26,6 +26,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 import (
+	"compress/gzip"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


jordan-wright:
Works for me! I don't see any downsides to using best compression since Gophish shouldn't be high volume enough to feel the effects of extra processing.

Thanks for this!

PR #460, opened 2016-11-29
Groups Page - Putting Limit on Targets selected

svigne1:
Hi,
When we go to the groups page in the admin server. The [`load`](https://github.com/gophish/gophish/blob/master/static/js/app/users.js#L153) function in `users.js` makes a call to `API_Groups` which [fetches](https://github.com/gophish/gophish/blob/master/controllers/api.go#L152) all the groups from DB along with all the [targets](https://github.com/gophish/gophish/blob/master/models/group.go#L241) of each group

As a result, the Groups Page takes a long time to load waiting for group targets it doesn't need.Since only 4-5 Targets for each group is [displayed](https://github.com/gophish/gophish/blob/master/static/js/app/users.js#L173) in the Groups Page. Rest all are omitted by adding "...".

So, I have included a GetLimitedTargets function which fetches only 15 targets for each group.
And in the [`edit`](https://github.com/gophish/gophish/blob/master/static/js/app/users.js#L69) function, I am making a call to `api.groupId.get` to get all the targets of a group for display.

### DIFF
--- a/models/group.go
+++ b/models/group.go
@@ -63,7 +63,7 @@ func GetGroups(uid int64) ([]Group, error) {
 		return gs, err
 	}
 	for i, _ := range gs {
-		gs[i].Targets, err = GetTargets(gs[i].Id)
+		gs[i].Targets, err = GetLimitedTargets(gs[i].Id)
 		if err != nil {
 			Logger.Println(err)
 		}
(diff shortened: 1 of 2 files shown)


S0larflare:
I was halfway through saying that the issue is that this would make it impossible to search through the groups for a specific user...but it seems like you can't do that anyway (it only searches through the first couple of users which are shown for each group on the page), so carry on!

jordan-wright:
I'm likely going to change what this API endpoint returns, since right now it's way too much data. Instead, I'll likely just return the number of members in the group and display that instead of the weird abbreviated member list. That would make it much faster to load, and easier to find large/small groups and, when the group is edited, I'll make a separate call to get the group details.

svigne1:
Hi Jordan,
Yeah, Number of targets sounds better.. Abbreviated members list is not very useful.
The separate call to get group details, I have used the existing [`api.groupId.get`](https://github.com/gophish/gophish/blob/master/static/js/gophish.js#L85) endpoint to fetch all the targets of a group for display

PR #462, opened 2016-11-29
Scale Campaigns & Dashboard Page

svigne1:
Hi,
Like #460 , The dashboard page and campaign pages are very slow to load since [`getDetails`](https://github.com/gophish/gophish/blob/master/models/campaign.go#L124) function called by `API_Campaigns` collects all the information related to a campaign like `results`,`events`,`smtp`,`page` etc. 

But the dashboard page and the campaigns page don't require this information. The Campaigns page  requires only the name and status. And the Dashboard page requires success percentage in addition to  names and status.

i have written a `avgSuccess` function which calculates the average using sql.
Thereby reducing the load on JS to calculate the avg and increasing performance
Hope this helps

### DIFF
--- a/controllers/api.go
+++ b/controllers/api.go
@@ -64,11 +64,17 @@ func API_Reset(w http.ResponseWriter, r *http.Request) {
 func API_Campaigns(w http.ResponseWriter, r *http.Request) {
 	switch {
 	case r.Method == "GET":
-		cs, err := models.GetCampaigns(ctx.Get(r, "user_id").(int64))
+		cs, s, err := models.GetCampaigns(ctx.Get(r, "user_id").(int64))
 		if err != nil {
 			Logger.Println(err)
 		}
-		JSONResponse(w, cs, http.StatusOK)
+		JSONResponse(w, struct {
+			Campaigns []models.Campaign `json:"campaigns"`
+			Avg       []int64           `json:"avg"`
+		}{
+			Campaigns: cs,
+			Avg:       s,
+		}, http.StatusOK)
 	//POST: Create a new campaign and return it as JSON
 	case r.Method == "POST":
 		c := models.Campaign{}
--- a/models/campaign.go
+++ b/models/campaign.go
@@ -176,20 +176,30 @@ type Event struct {
 	Details    string    `json:"details"`
 }
 
-// GetCampaigns returns the campaigns owned by the given user.
-func GetCampaigns(uid int64) ([]Campaign, error) {
+func (c *Campaign) avgSuccess() int64 {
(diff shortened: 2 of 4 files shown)


jordan-wright:
I like the idea of making the overview of multiple campaigns more efficient and, you're right, we don't need all the data we're pulling back.

Instead of changing the API to return more data in a seemingly odd response format, I'd rather just change the `/api/campaigns/` endpoint to return campaign stats instead of the full campaign details.

I'll probably approach this by returning something like a `[]CampaignStats`, which would look something like:

```
{
    "name" : "Campaign Name",
    "created_date" : "[date]",
    "launch_date" : "[date]",
    "stats" : {
        "recipients" : 0,
        "error" : 0,
        "sent" : 0,
        "opened" : 0,
        "clicked" : 0,
        "phished" : 0
    }
}
```

That way, the API still returns clean objects, returns *way less data*, and the JS still has to do minimal work to get the numbers we care about.

svigne1:
Right now, all the stats are calculated by the front end. And i checked through the code. The only statistics it calculates is the success % for each campaign which it calculates as shown [here](https://github.com/gophish/gophish/blob/master/static/js/app/dashboard.js#L81).

So,in the json you posted, what is your plan for "stats" key?

And I changed the `/api/campaigns` endpoint to give only the statistics the UI was calculating, the success % of every campaign. Instead of giving out the success percentages as an array, i thought of calculating the average of them, which was what the UI was actually interested in and named the variable avg.

But i forgot this line of thought, and sent back the success percentages as an array itself with avg as still its name. Sorry!

jordan-wright:
I'm not as worried about the JS doing some division operations as I am making sure our API is clean. By introducing this extra array, were sending more data and, in my opinion, it makes the API seem disorganized.

The campaigns API endpoint should just return campaign objects. Now, my initial mistake was that I chose to return every aspect of every campaign when that is hit, which doesn't scale.

Instead, my plan is to return overview objects including the stats of each campaign (pulled via server side sql) and return that. When someone then hits /api/campaigns/:id, we will return all the information.

Sure, the JS will need to calculate the avgs, but a few frontend divisions shouldn't be the bottleneck we're experiencing.

jordan-wright:
Hey @svigne1!

Just a heads up, I implemented the campaign summary endpoints, which are now being used in the "Dashboard" and "Campaign" pages. Things should be a bit quicker now. You can see more details in #505.

I'm planning on extending this to other endpoints for objects (like groups, landing pages, email templates, etc.).

I'm going to go ahead and close this one out since we have basically the same functionality through different means. Thanks again for sending this our way!

PR #471, opened 2016-12-02
Attachment Bug Fix

svigne1:
Hi,
When the config is changed to point to a Mysql Db. Then I am facing problems when i try to send attachments of size > 49 KB.
So, if an image with size > 49 KB is saved in the email template and say a campaign is launched.
The Mail server throws "UNEXPECTED EOF" error. And in the mail you receive, you only get a cropped image.

Changing text to longtext fixes the problem for me.


jordan-wright:
Work for me! Good catch!

PR #482, opened 2016-12-07
User defined delay between emails

S0larflare:
So I had seen a few people saying that they would like a delay for sending emails as some systems might block a large amount of email coming in one go or some people might want to run a campaign slowly over a large time period.

With that in mind, I've done a few simple changes and added the ability to define a delay between emails per campaign. There is a field added into the new campaign screen which defines the delay, in seconds. This field is set to 0 by default, and should only allow numbers (so no decimals, letters or negative numbers allowed). Feedback more than welcome...

Also, I added an error output to api.go (this was something I'd added in to trace a JSON error I was having, so I don't think it necessarily needs to stay. I have also deleted a space in db/db_sqlite3/dbconf.yml as it was causing issues when using goose up/down. 

I haven't done this for mysql, as I am using sqlite and wouldn't have a means to test it, that said it should be about as trivial as replicating the column in mysql to allow for the change.

### DIFF
--- a/controllers/api.go
+++ b/controllers/api.go
@@ -75,11 +75,13 @@ func API_Campaigns(w http.ResponseWriter, r *http.Request) {
 		// Put the request into a campaign
 		err := json.NewDecoder(r.Body).Decode(&c)
 		if err != nil {
+			Logger.Println(err)
 			JSONResponse(w, models.Response{Success: false, Message: "Invalid JSON structure"}, http.StatusBadRequest)
 			return
 		}
 		err = models.PostCampaign(&c, ctx.Get(r, "user_id").(int64))
 		if err != nil {
+			Logger.Println("bob12345")
 			JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusBadRequest)
 			return
 		}
--- a/controllers/api.go
+++ b/controllers/api.go
@@ -81,7 +81,6 @@ func API_Campaigns(w http.ResponseWriter, r *http.Request) {
 		}
 		err = models.PostCampaign(&c, ctx.Get(r, "user_id").(int64))
 		if err != nil {
-			Logger.Println("bob12345")
 			JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusBadRequest)
 			return
 		}
--- a/db/db_sqlite3/dbconf.yml
+++ b/db/db_sqlite3/dbconf.yml
@@ -1,5 +1,5 @@
 production:
-    driver: sqlite3 
+    driver: sqlite3
     open: gophish.db
     dialect: sqlite3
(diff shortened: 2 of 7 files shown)


mkannan22:
I just built from source and do not currently see a delay button anywhere. Could you point me in the right direction of where it should be displaying. I tried New Campaign and can see Schedule, but no delay. Also checked Sending Profile with no luck. The version is 0.3-dev

S0larflare:
Its not a button, if you click on new campaign you should see a field which you cam specify the delay between emails in seconds. By default this is set to 0. If you can't find it I'll have a look into it on Monday when I get a chance. It should work as I tested it on my own fork before moving it over here

mkannan22:
Thanks. I have tested it on 2 different machines with the same results (i.e. nothing related to a delay, no field). I can try again on a different OS (current one is Win 10).

jordan-wright:
You're likely building the main gophish repo. This isn't merged in yet. You'll need to build the fork to get this to work until we can address getting this merged.

mkannan22:
That is correct. I will wait until it's merged into the main branch.

S0larflare:
Yeah sorry, I thought you had built my fork of this. If you want to test it out (more than welcome as I've only managed to test it myself and you could find something I've missed) then you're welcome to build the one on my fork, the master has this implemented. I'd say to try it separately rather than replace your current version though

lordsaibat:
I am getting errors when the trigger fires.
I am getting a Error: timeout Exceeded.


Might need to have the code re-authenticated to the sending server.

I have tested 240 and 300 seconds (2/5 min)

S0larflare:
Hmm, yes. I hadn't realised that (as personally I only used this for spreading out sending a little bit, rather than having large periods, so only went up to about a minute of actual delay between batches.

I've had a fiddle with it and I can fix it, essentially by replicating the calls to gomail within the loop, if the delay is sufficient to warrant it. I'm not sure that this is the cleanest way to do it, however.

This PR is pretty stale, so I should probably shut it down. I did extra work on this and added the ability to randomise the sending and state a batch size. It probably needs more work either way before I'd be happy to move on with it.

lordsaibat:
I would make the sending of the mail a function where you pass the data to. Then inside the code call the function with the data.

I would do it but I have no clue on go.

ZeteMKaa:
Is this still being implemented?

S0larflare:
This particular one is old and no longer compatible with Gophish. It was also a bit of a quick and dirty hack. I have something a bit better that I use internally, but still, it is a bit of a hack job. To do this properly would involve completely changing how Gophish handles sending emails, which I know is on the radar, but might take some time before anything is seen. It's something that will likely have to wait until Jordan gets round to it (as he is orders of magnitude better than I am at this) as I don't know the greater plan on how email sending will change.

PR #504, opened 2016-12-23
Deleting campaign & groups fast

svigne1:
Hi,
For campaigns, I am moving the deletion of events and results into a go-routine.
For Groups, I am moving the deletion of groupTargets into a go-routine.

And i am ignoring the errors thrown in the go-routine.

The reasoning is, I think, even if the db operation in the go-routine alone fails for some reason, the details related to that group or campaign being deleted still remains in the database. But there is no way, (either through the api or through the browser), for the end-user to retrieve that data from the database. So, its as good as calling it deleted.

Even in file systems, when we delete a file. Only the inode is deleted and the file contents remain as it is  in the file system. Even though the contents are note deleted and still remain in the file system.There is no way for the end-user to access those contents without the inode entry.

The only problem is, In the case of file systems the new files overwrite the deleted files.
But in database, they stay there forever.

I think, a chance at data inefficiency is acceptable for the boost in performance we get.
Please correct me if i am wrong

### DIFF
--- a/models/campaign.go
+++ b/models/campaign.go
@@ -331,22 +331,22 @@ func PostCampaign(c *Campaign, uid int64) error {
 //DeleteCampaign deletes the specified campaign
 func DeleteCampaign(id int64) error {
 	Logger.Printf("Deleting campaign %d\n", id)
-	// Delete all the campaign results
-	err := db.Where("campaign_id=?", id).Delete(&Result{}).Error
-	if err != nil {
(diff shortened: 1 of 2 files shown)


jordan-wright:
I like the idea of this, but we need to be careful. The SQLite library we're using doesn't appear to be goroutine-safe (see https://github.com/mattn/go-sqlite3/issues/350), so it seems like this could cause issues.

Issue #538, opened 2017-02-13
Localization For Web Application [contributor-friendly]


alenbhclynpblc:
Application can be localized over 2 main area, one of is the Go source code and the javascript codes.

I suggest we can use 'util' package for translation operations and call an translation method over the util package.

The javascript part is be more tricky, i think it will be nicer if somebody use only one localization file not two different localization files so we can generate javascript dictionary over go for text conversions and an simple javascript method do the conversions.

Finally i did coded my suggestion :) with go-i18n implementation, if you want to use i will be happy to contribute.

jordan-wright:
As someone new to this area, what do you mean when you say "localized" in relation to the Go source code or the JS?

I guess the best way to move forward would be to go ahead and post a pull request and we'll take a look 😄

S0larflare:
I have a suspicion that what  alenbhclynpblc means is that if you move Gophish from your Gopath, make changes to it and rebuild it, it builds using the go files that are in you gopath rather than the ones you have moved. Confused me for a little while wondering why I didn't seem to be able to make any changes!

alenbhclynpblc:
@jordan-wright Localization / Internalization etc. 😄 i did want to say all texts (error messages etc) on application source code (which written with go-lang) and hard coded strings on javascript files ('Edit' / 'Delete' etc).

PR #540, opened 2017-02-14
Issue #538 // Localization For Web Application

alenbhclynpblc:
Config file have an 'language' parameter, 'en-us' / 'tr-tr' is the current options and translation files are in the 'translations' directory.

This is the first step of localization, messages need to be changed with keywords (example: "There is no any campaign." => 'noCampaignWarn') it will be more effective, pluralized strings needs to be implemented etc.


jordan-wright:
This diff is pretty huge...

The more I think about this, it might be best to create a file dedicated to "constants" that transparently returned the correctly translated string. This is opposed to calling the translation function individually for each hardcoded string in the file.

We'll also want to split this up. It's pretty infeasible for me to review a diff this large easily. It'd be better if we could split this up logically and work through this a bit at a time.

All things considered - thanks for giving this a first pass!

alenbhclynpblc:
Hey, hi @jordan-wright 

You r right it is a little bit big commit for review.

About the constants... My opinion about constants, It is not a pretty solution but yeah probably faster application runtime.

Right now application uses fixed strings and text concatenation but and of the day application must have formatted strings (all the dynamic parts added by javascript). Right this point application format the string with an function :) (worst case printf)

```
fmt.Printf("Campaign: %s, is succeed!", campName)
```

If we will implement an translation library, any i18n topic will effect only one method so application development time will be reduced about i18n developments. If localization becomes complex development time will increase to much and of course error handling will fail if constant not found etc....

After all localization is not only about language, time is the other concern. Application must be able to set timezone based on user account so app will need an user based constant about time zone etc...

After all if you want to go with constants, i can help about development of this feature, can you share what is in your mind ?

PR #544, opened 2017-02-20
Adding support for custom headers in sending profiles

jordan-wright:
This pull request adds support for using custom SMTP headers in sending profiles.

This is based on the work done by @ss23. I opted to move this functionality to sending profiles instead of templates, because I envision many of these custom headers will likely be used across all templates.

image


jordan-wright:
This also adds the support for headers in the API. Here is the new response format for `/api/smtp/:id`:

```
[
  {
    "id": 1,
    "interface_type": "SMTP",
    "name": "Mailhog",
    "host": "localhost:1025",
    "from_address": "Test Test \u003ctest@test.com\u003e",
    "ignore_cert_errors": true,
    "headers": [
      {
        "key": "Reply-To",
        "value": "example.com"
      },
      {
        "key": "X-URL",
        "value": "{{.URL}}"
      }
    ],
    "modified_date": "2017-02-19T16:52:05.7483549-06:00"
  }
]
```

abhisheeeka-boop:
gm8880007@gmail.com

PR #564, opened 2017-03-16
Added refresh button and reduced lag for large campaigns

S0larflare:
This is a really simple one, I've increased the automatic refresh time to 60 seconds and implemented a refresh button to manually kick off a refresh if desired. I also moved the refresh_message.hide() code to the end of the poll() function, so that it will show 'refreshing' until it is actually finished refreshing


S0larflare:
Apologies, managed to only grab one commit...

Issue #552, opened 2017-03-06
Campaign Overview - Improvement request


mdearlove:
Hi please can you consider adding the date that the status "email sent" was achieved as a column on the Campaign overview view?  

I have some campaigns that are in progress but I have to drill down to see when the email was sent out which could vastly different than the "Created Date".

Also for campaigns that are still queued - please add a tool tip to the word 'Queued' to indicate the actual date the campaign will be launched as it is impossible through the current UI to find this out by drilling down to the campaign.

Thanks for your help :-)

jordan-wright:
Good suggestions!

I'll see what I can do UI-wise to address some of these suggestions. I agree that drilling down into campaigns to find some of this information is not ideal.

mdearlove:
Thanks @Jordan-wright.

S0larflare:
I'm currently having a look at this (as having to dive deep into the api to read all the details isn't ideal). So far I've just added a new "tab" in campaign results which has some details in it:

![image](https://cloud.githubusercontent.com/assets/8939757/24105433/c67d4ae0-0d7c-11e7-9b7a-dae9bf061a9f.png)

So you can hop between the tabs, rather than cluttering up the results page. What extra sort of info would people find useful to go there?

mdearlove:
I appreciate the idea but bear in mind you still need to drill down to the campaign to see this info.
What about at the top level campaign level that lists all campaigns, have another button to the left of the "View Results" button which provides a quick summary pop-up window showing:

Scheduled Start Date/time:
Emails Sent:
Emails Opened:
Attachments Opened (if applicable or possible):
Link Clicked/Success:


Then rename the "View Results" button to "Campaign details"?

Just some ideas :-)

S0larflare:
Is this more what you're after?

![image](https://cloud.githubusercontent.com/assets/8939757/24198020/f8d0c1ae-0efb-11e7-9156-92788bc561b2.png)


![image](https://cloud.githubusercontent.com/assets/8939757/24198038/02ffa640-0efc-11e7-8f5a-fe52a9cba02a.png)

mdearlove:
Hi @S0larflare, Yeh thats looking good thanks.  Just means you get the summary info without needing to drilldown.  great work!

jordan-wright:
@S0larflare I *really* dig that. Good call!

S0larflare:
[...]

PR #600, opened 2017-04-24
Adding a quick stats popup to campaign page

S0larflare:
Finally got around to putting together a pull request for #552.

Currently this is only on the campaign overview screen, I didn't duplicate it for the dashboard view.


jordan-wright:
Excellent! As always, thanks for the great work @S0larflare!

PR #604, opened 2017-04-26
Add robots.txt handler

mattaura:
Disallow all robots from accessing the phishing server, to prevent phishing materials from being indexed during campaigns.


jordan-wright:
Good call - thanks for sending this in and sending a patch at the same time. I think there's a cleaner way to write the content using `fmt.Fprintf` but I'll merge that up. For now, this should work just fine.

Thanks again!

mattaura:
Great, thanks - still a Go newbie :)

jordan-wright:
No worries! Putting up code makes all the difference.

[Here's](https://github.com/gophish/gophish/blob/master/controllers/route.go#L276) the code I wound up using for reference.

Thanks again for sending this in! Good stuff!

Issue #223, opened 2016-03-27
RFC: New way to create campaigns


jordan-wright:
Hey everyone,

I wanted to open up an RFC to see what everyone thought about a new way to create campaigns that I'm working on.

Right now, creating a campaign feels very static and restricted. Select *one* landing page, *one* email template, etc. I'm currently working on a new way that would make campaign creation much more modular with "flows". Here's a screenshot:

![image](https://cloud.githubusercontent.com/assets/1317288/14063730/b4635bcc-f3a3-11e5-8b36-ab5c232e9b4d.png)

This would make creating a campaign much more modular. Using this, you would be able to add as many actions as you want. Want to send multiple emails? You can. Want to redirect the user to multiple pages (#219)? You can.

The other thing I am working on is how to structure this in templates, etc. I'm considering creating a standard template variable to replace ```.URL```. Specifically, I would implement a template variable ```{{.Next}}``` that would point to the next URL in the chain. This way, you can have each landing page or email be plug-n-play to all work together nicely.

Thoughts? Excitement?!

rfdevere:
The best phishing framework just got better. Anything that gives the
product a feature and doesn't ruin stability or ease of use a great.

Guiding people to multiple landing pages would be brilliant, I'm thinking
already being able to load in a 1000 email staff list and send admin staff
to google docs pages, finance staff to banking pages, everyone else to the
company SSL VPN.

Great work Jason we all appreciate them long nights bashing buttons.

On Sunday, 27 March 2016, Jordan Wright <notifications@github.com> wrote:

> Hey everyone,
>
> I wanted to open up an RFC to see what everyone thought about a new way to
> create campaigns that I'm working on.
>
> Right now, creating a campaign feels very static and restricted. Select
> *one* landing page, *one* email template, etc. I'm currently working on a
> new way that would make campaign creation much more modular with "flows".
> Here's a screenshot:
>
> [image: image]
> <https://cloud.githubusercontent.com/assets/1317288/14063730/b4635bcc-f3a3-11e5-8b36-ab5c232e9b4d.png>
>
> This would make creating a campaign much more modular. Using this, you
> would be able to add as many actions as you want. Want to send multiple
> emails? You can. Want to redirect the user to multiple pages (#219
> <https://github.com/gophish/gophish/issues/219>)? You can.
>
> The other thing I am working on is how to structure this in templates,
> etc. I'm considering creating a standard template variable to replace .URL.
> Specifically, I would implement a template variable {{.Next}} that would
[...]

PR #662, opened 2017-06-20
Added Education Template for Creating Educational Redirect Pages Insi...

ConnorPatterson:
…de Static Folder

Redirecting phishing victims to pages hosted on the phishing server is a very useful feature, currently implemented in GoPhish through the /static folder.  One example where this is particularly useful, for instance, is for phishing campaigns in which victims should be redirected to an educational page, alerting them that they had fallen for a phishing attack and explaining how to avoid attacks in the future.  This fork demonstrates an additional feature added to the admin page, which allows HTML files inside the /static directory to be created, modified and deleted.  On the landing_pages URL, these pages are also suggested in a drop down menu on the redirect_URL field.  The changes are very simple, only 2 .go files were modified so the fork can easily be compiled and tested on Desktop, everything is done through AJAX like the Settings page (no new models created or API functions modified).  

Perhaps if this seems desirable, you can consider implementing this feature in the future.  Thank you for your work on GoPhish, the framework is fantastic!

---------------------------------------------------------------------------------------------

gophish.go: /controllers is imported from relevant local directory, for purposes of compiling changes made therein

controllers/route.go: Added handling for /education URL, including handling AJAX requests

templates/education.html: New template for creating, editing and managing HTML pages inside the static folder.  

templates/landing_pages.html: Added some information to the hover help icon by the redirect field, and a dropdown that suggests files from the static folder.

/templates: Most of the other templates had the education page href added to their menus

### DIFF
--- a/controllers/route.go
+++ b/controllers/route.go
@@ -3,9 +3,12 @@ package controllers
 import (
 	"fmt"
 	"html/template"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/gophish/gophish/auth"
 	"github.com/gophish/gophish/config"
@@ -29,6 +32,9 @@ func CreateAdminRouter() http.Handler {
 	router.HandleFunc("/login", Login)
 	router.HandleFunc("/logout", Use(Logout, mid.RequireLogin))
 	router.HandleFunc("/campaigns", Use(Campaigns, mid.RequireLogin))
+
+	router.HandleFunc("/education", Use(Education, mid.RequireLogin))
+
 	router.HandleFunc("/campaigns/{id:[0-9]+}", Use(CampaignID, mid.RequireLogin))
 	router.HandleFunc("/templates", Use(Templates, mid.RequireLogin))
 	router.HandleFunc("/users", Use(Users, mid.RequireLogin))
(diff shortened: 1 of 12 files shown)


jordan-wright:
Hey @ConnorPatterson, thanks for sending this in!

I really like the idea of offering pages for education. I think that it's always been a feature of Gophish that would be useful to quite a bit of people.

That said, I'm probably going to hold off on merging this in the short term, but I *promise* there's a reason 😄 In the next month or two, I'm planning on introducing some pretty big(ish) features into Gophish. One of these will be the new campaign builder as seen in #223. (As a side note, I know the issue has been open forever but I swear it's a work in progress ;))

One thing I want to do with the new campaign builder is let a user define "flows" where they can, for example, say "send the user to landing page x and then to landing page y". Then, instead of making an entirely different model for education pages or serving them from the static assets directory, training pages will simply be a classification of existing landing pages. So users can use the same process to create educational pages as they do with landing pages.

With how much that would change the code in the pull request, it seems like it might not be worth merging this in for the short term gain, since we'd have to change quite a bit of it to adjust to the new campaign builder.

All that said, I really, really appreciate you sending this over. It's always great to see how other people approach these types of problems!

Please don't hesitate to reach out if you have any questions. The last thing I want to do is have people who are kind enough to send over pull requests feel like they aren't getting a full response from me.

ConnorPatterson:
Hey Jordan, thanks for getting back to me so quickly,

I also understand GoPhish is built around some conventions, specifically communicating with the database through API models, etc, that my code doesn't follow (hence using AJAX to modify the least amount of GoPhish source files), and things would need to be reworked pretty drastically with the new campaign builder feature, which looks fantastic, btw :smile: 

Thank you for maintaining such an awesome framework, I'm looking forward to upcoming features!

jordan-wright:
Awesome, thanks for being so understanding. I'm going to go ahead and close this one out, and I'll be sure to keep you updated as the new campaign builder gets finalized.

PR #703, opened 2017-07-21
Submit credentials to original origin after capturing them

clod81:
During client engagements, if the origin being targeted doesn't have CSRF protection, we often want to capture the credentials and submit the actual creds back to the site being phished.
I couldn't currently find any way to do so out of the box with `gophish`, apart from writing some custom JavaScript on the landing page itself.

I came up with this solution that, if the option is checked, it will make use of jQuery, and create some JS code within the `<head>` of the DOM, that intercepts all forms submission, send the data by Ajax back to gophish, and then submits it to the original origin.
As mention, this only works for sites that do not implement CSRF protection on the login page (ex: Outlook Web Access - OWA).
Also, if you clone a site that has already some JS that handles form submission, you'll have to get rid of that, to avoid conflicts on the JS itself when executed.
At the top of this, the form `action` will have to include the full URL, not only the `path`.

Thanks @dunderhay for the nice idea.


clod81:
Note that tests haven't been written yet, I'd like some feedback first on the actual implementation, before carrying on with those.

models/models_test.go
```diff
@@ -299,8 +299,24 @@ func (s *ModelsSuite) TestPostPage(c *check.C) {
 		c.Assert(ok, check.Equals, true)
 		c.Assert(u, check.Equals, "username")
 	})
+	// Submit with SubmitToOriginal set to true
```
jordan-wright: You're just the *best* kind of person for adding tests! 😄

jordan-wright:
Hi @clod81!

Thanks so much for sending this in- I really like the idea. My original thoughts around something similar was to create a "full proxy" where we would make a flow like this:

For every `<a>` link on the page or form `action`, rewrite the response to point to `gophish_ip/proxy?url=[base64_of_actual_url]`
Then, when the link is clicked, we intercept the request and send our own request to the actual URL with the same payload as the original request.
When the response comes back, we rewrite the response again pointing back to Gophish

This seemed to work _ok_ in tests and would let us get around things like CSRF protection, etc. since we're just a mitm at that point. But there were some tricky things I was having to work around like JS that relied on fetching files relative to the current location, etc.

Anyway, this seems like a good stop gap while I finish the full proxy stuff which can make this more general. Plus, I could see some people not being interested in keeping users on the hook, but rather just wanted to do a one time submission.

I'll review the code this week and, if all looks good, I'll merge it in. Thanks again for taking this on!

clod81:
@jordan-wright no problem. I initially thought that the options were taken into consideration when the landing page is rendered, and not when saved. That's why the JS component is written in a way that the user can edit the landing page options and enable and disable the option as many times he prefers (with the only downside of adding the JS block over and over).
I like the `proxy` idea too. Let us know how you get on with that 👍

clod81:
@jordan-wright any updates on this? cheers

jordan-wright:
Hey there!

100% in progress. I just managed to push up some unrelated changes a couple of days ago, so now I can focus on this one. I'll merge it into a separate branch to test everything and, if all looks good, will merge it in.

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

securitygeneration:
Is this going to make it into an upcoming version soon?

PR #704, opened 2017-07-22
Gophish Ansible playbook

derpadoo:


jordan-wright:
Thanks for submitting this pull request! I'm afraid I don't know enough about Ansible to properly review the playbook here, but everything seems reasonable enough. 🤷‍♂️ 

My only immediate concern is around storing a separate copy of the config. If we're pulling the latest gophish binary and the config format has changed (not really likely since I'm picky about those changes, but possible), things might break.

Then again, I suppose we could always just update the ansible config to match.

Let me think on this and maybe try it out myself and we'll see if we can get this merged 😄 

Thanks again for sending this over!

derpadoo:
Yeah it would break if you changed the config format.  I can remove it if you want, but I'd say just keep it.  If people end up using the playbook that issue will arise quickly.  I'm not a Ansible ninja, but that playbook works for me to quickly spin up a Gophish server and I wanted to give back...if you don't want it in the project I'm totally fine with that.  Thanks for adding the static download link to support this playbook and for maintaining the project!

ansible-playbook/roles/gophish/files/config.json
```diff
@@ -0,0 +1,17 @@
+{
+    "admin_server" : {
+        "listen_url" : "127.0.0.1:3333",
+        "use_tls" : false,
```
jordan-wright: Could we set this to `true` to match the existing config file? https://github.com/gophish/gophish/blob/master/config.json

jordan-wright:
Hey there!

Sorry for the delay on this. Now that Blackhat/Defcon are over I finally have time to get back to some of this stuff.

I've added a comment about setting the default admin server tls setting to true. If we could make that match our config.json in the repo then I'll get this merged in.

Thanks again for sending this over!

derpadoo:
Done - submitted new pull request.

jordan-wright:
Looks good! I'll merge this in.

Thanks again for sending this over!

PR #713, opened 2017-08-04
Users & Groups - Generating Random Subgroups

mnickels:
If you load in a large CSV file of users to phish, but don't want to send the same email to every user, or don't want to send all of your phishing emails at the same time of day, you would have to divide the groups into separate CSVs before loading each one individually.

I've added a button to the Users & Groups page called "Create Random Subgroups" that takes a single parent group and divides it into a specified number of subgroups. The distribution of users is uniform (i.e., a group of 50 users divided into 6 groups would yield 4 groups of 8 users and 2 groups of 9 users), and the grouping is random. Number of subgroups must be greater than 1 and less than or equal to the number of users in the selected parent group.

I also made several minor/superficial changes, listed below:

- Changed "overflow: none" to "overflow-x: none; overflow-y: auto" for .modal class in gophish.css (this was preventing scrolling in Landing Page and Email Template modals after importing)
- Changed dashboard graph to label the x-axis by date launched, not date created, and made the displayed date & time format more compact
- Made the pie chart only show data from campaigns that have launched

I made these changes mainly to meet my own needs, but thought they might be useful for others as well. Looking forward to hearing any and all feedback!


jordan-wright:
Hey there!

Thanks for sending these over! I really appreciate the work you put in to make this happen.

Tbh, it would probably be easier to review and work on each of the things you mentioned as separate pull requests.

For example, I like the idea of subgroups, but I'm not sure that making them permanent groups would be the best option. What if instead we had an option in the campaign builder to send to a random set number of users given different groups? That way this happens at campaign time and doesn't leave a group the user has to delete later. Just a thought!

Also, regarding the graphs, those are all great ideas! I'm actually refactoring the graphs (https://twitter.com/jw_sec/status/883115117600264192) to use Highcharts and changing some things about the way they're displayed. This would cause quite a bit of merging, so I'm going to hold off on merging this stuff until I can push up those changes (should be this week).

Thanks again for sending these over! If you could make separate pull requests that'd be super helpful, otherwise we can work through this one as needed throughout the week. 😄

mnickels:
Hey Jordan, thanks for the feedback! I'll get these separated into individual pull requests.

PR #715, opened 2017-08-08
Sidebar href fixes

mnickels:
The hrefs for the Dashboard and User sidebar items redirected in a different manner than those in other pages. I've changed these links to match the behavior of the other templates.


jordan-wright:
Ah, good catch. Thanks for the heads up! I'll merge this in.

PR #716, opened 2017-08-08
Generating Random Subgroups

mnickels:
If you load in a large CSV file of users to phish, but don't want to send the same email to every user, or don't want to send all of your phishing emails at the same time of day, you would have to divide the groups into separate CSVs before loading each one individually.

I've added a button to the Users & Groups page called "Create Random Subgroups" that takes a single parent group and divides it into a specified number of subgroups. The distribution of users is uniform (i.e., a group of 50 users divided into 6 groups would yield 4 groups of 8 users and 2 groups of 9 users), and the grouping is random. Number of subgroups must be greater than 1 and less than or equal to the number of users in the selected parent group.

Jordan - I know you suggested moving these features to campaign-time to prevent the user having to remove the subgroups manually at any point, but I feel as though having the groups persist is more useful than it is harmful; for example, if you want to phish the subgroups at different times / dates, or with a different email template / landing page / sending profile. This is what I am looking for in my use case, but definitely let me know if I'm missing something from your suggestion!


teddielong:
How easy/difficult is this to implemenet?  I'm a fairly novice user but have managed to get GoPhish up and running at least.

mnickels:
Do you mean how difficult was it to write the code for this PR? If so, not very difficult - the Gophish API is pretty simple and intuitive, and you can use the js already there to see how it's used. This PR only required changes to the html and js files.

teddielong:
Thanks for the quick response!!

Apologies, sorry - I meant how easy is it to implement this within my current version of GoPhish?  I'm a very novice user, but would love to be able to send via sub-groups as currently all emails being sent at the same time isn't ideal but nor is manually adding lots of little groups - so this sounds ideal.

mnickels:
Got it! Currently Jordan has been too busy to fully review and accept this PR, so you can't get it here. I have a branch here:
https://github.com/mnickels/gophish/tree/temp-deploy
that integrates this PR as well as PR #728 that you could just clone onto your Gophish server and build from there, it builds the same exact way as Gophish (see the readme's "Building From Source" section). Fair warning, that branch is only up to date with Gophish from the end of August, but I can update it in 3-4 hours if you want to check back then.

mnickels:
Alright, I've updated the temp-deploy branch and verified it builds. To build, use the command
`go get https://github.com/mnickels/gophish.git`
`cd` into [your go path]/src/github.com/mnickels/gophish
and you should be able to `git checkout temp-deploy`
then run `go build` and a gophish executable should be created in that directory. Of course, you will also need the js / html templates from my branch, so you will want to clone the whole repo and replace your current gophish directory with it.

turnerdb:
Is this version still available for testing?  I would like to test this in our dev lab.

Thanks.

mnickels:
@turnerdb Yes, although I haven't touched it in a while, so it's pretty far behind this repo. I'm planning on revisiting it soon, I think the reason why this and my other pull requests were never merged is because I didn't properly handle asset minification (whoops! I was a pretty new web dev at that point in time though).

I'll try to give it another go this weekend, but ITM the fork should still work.

lukepritchard-homstar:
Hi can we look at getting this merged?

PR #728, opened 2017-08-21
Aggregate Results for All Targets

mnickels:
There is currently no way to view all targets of multiple phishing campaigns side-by-side. I've added a new page to the web UI that displays all targets of each campaign, their status, and a link to the respective campaign that they were included in.

The targets are automatically sorted by status. I added a `<p>` tag with `style="display:none;"` before the status message in order to artificially sort the statuses in the following order:

- Submitted Data
- Clicked Link
- Email Opened
- Email Sent
- others

Unfortunately, the invisible `<p>` tag gets included in the copy and export actions (so statuses end up looking like "0Submitted Data" when exported). Possible solution could be filtering these out before exporting, then re-inserting, or maybe just figuring out another method by which to create this sort order.

I would love to hear any suggestions for improvement! Since I added a new page to the UI, the route.go file had to be modified, so this PR requires a `go build` to route properly to the new page.


H4xl0r:
@mnickels could you make a seperate page with anonymised infos about a specific campaign?
like an extra page for a (every) specific campaign , showing just anonymised infos like : mails send , mails opend , timeline , most used browsers and os and devices ?

H4xl0r:
still working with some changes !

PR #851, opened 2017-11-21
Removed unwanted event "CampaignCreated" fro each email user

jas-deep:
One unnecessary event "CampaignCreated" per email user is being logged.


jordan-wright:
Thanks for the heads up on this! I'm in the process of refactoring this exact code, so I'll just remove the affected lines in the refactor to avoid merge conflicts when I push up the code.

Thanks again!

PR #890, opened 2017-12-15
fix: SQL syntax error

Thiago-Medeiros:
update autoincrement -> auto_increment
add ";" after drop table


aaniceto (approved):

jordan-wright:
Good catch - thanks!

Issue #904, opened 2018-01-03
Problem of display in "New Group"


Cycl0pe:
What version of Gophish are you using?:  V 0.4

Brief description of the issue: When adding a new user to fish in "New Group", there is a problem of display when the input is too long.

What are you expecting to see happen? : Perhaps,  a carriage return so the display of the new user to phish does not impinge out of the modal.

What are you seeing happen? The display of the new user impinge out of the modal

![test](https://user-images.githubusercontent.com/24911604/34515020-779df12c-f070-11e7-885d-ec274dd65685.png)

jordan-wright:
What browser are you using?

Either way, thanks for reporting this. I'll see if I can fix this when I get a chance.

Cycl0pe:
Hey,

Firefox 57.0.1!

Thanks

shuheiktgw:
Let me work on this issue! 👍

shuheiktgw:
Worked on this issue @ https://github.com/gophish/gophish/pull/915

jordan-wright:
This was fixed in #915.

Cycl0pe:
Hello Jordan,

I still have the same problem? Am I the only one with this problem?

PR #915, opened 2018-01-07
hidden overflow on targetsTable

shuheiktgw:
## Related Issue
https://github.com/gophish/gophish/issues/904

## Overview
Texts overflow on targetsTable, so I hidden overflow by adding some styles.

**before**

**after**

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -437,8 +437,14 @@ p {
 
 /* Table Styling */
 
+.modal-content table {
+    table-layout: fixed;
+}
+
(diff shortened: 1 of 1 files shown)


shuheiktgw:
I'm not pretty sure if I need to include `static/css/dist/gophish.css` ....

jordan-wright:
Nah, don't worry about committing the build file. I prefer to build them myself.

This looks good - I'll just test and make sure the other modal tables still look ok.

Thanks for sending this over!

PR #916, opened 2018-01-07
Add gophish_admin.crt and gophish_admin.key to .gitignore

shuheiktgw:
Those files should not be committed at any chance so I think that would be safer if those files are added to .gitignore


jordan-wright:
Looks good - thanks!

Issue #693, opened 2017-07-17
Can't delete registered users?


Void1ight:
I made a few test accounts that I used to configure everything, and not I would like to delete them. Is there a way to do this? I understand if there isn't a way to delete the default admin account, but is there a way to delete user created accounts? There doesn't seem to be, unless I'm missing something....

jordan-wright:
I'm afraid there isn't a current way to delete registered users from the GUI. However, you should be able to remove the rows in the `users` table using any SQL client.

jordan-wright:
For now, I'll keep this open as a feature request, though it's pretty low priority compared to other things in the pipeline.

Void1ight:
No problem. Thanks!

shuheiktgw:
Let me work on this issue! 👍

joao-intelliway:
This feature is very important to be GDPR compliant

PR #919, opened 2018-01-09
Add deleteUser method to user.go

shuheiktgw:
## Related Issue
https://github.com/gophish/gophish/issues/693

## Overview
This is one of the several PRs to enable to delete a user! 👍 
In this PR, I implemented `Delete~byUserId`and `DeleteUser` methods to each models in order to delete all the related records when deleting a user.

## Next PR?
I will work on a method in a controller and front after this PR is merged!

## FYI
I personally think I need to use a [transaction](http://jinzhu.me/gorm/advanced.html#transactions) when deleting a user, but I will work on this issue in a different PR since I need to fix several other places where executing a batch deletion without a transaction.

### DIFF
--- a/models/campaign.go
+++ b/models/campaign.go
@@ -471,8 +471,8 @@ func DeleteCampaign(id int64) error {
 	return err
 }
 
-//DeleteCampaignByUserId deletes campaigns with a given user id
-func DeleteCampaignByUserId(uid int64) error {
+//DeleteCampaignsByUserId deletes campaigns with a given user id
+func DeleteCampaignsByUserId(uid int64) error {
 	// Get campaigns
 	cs, err := GetCampaigns(uid)
 	if err != nil {
--- a/models/campaign.go
+++ b/models/campaign.go
@@ -471,6 +471,44 @@ func DeleteCampaign(id int64) error {
 	return err
 }
 
+//DeleteCampaignByUserId deletes campaigns with a given user id
+func DeleteCampaignByUserId(uid int64) error {
+	// Get campaigns
+	cs, err := GetCampaigns(uid)
+	if err != nil {
+		Logger.Println(err)
+		return err
+	}
+
+	var cids []int64
+	for _, c := range cs {
+		cids = append(cids, c.Id)
+	}
+
+	// Delete all the results
+	err = db.Where("campaign_id in (?)", cids).Delete(&Result{}).Error
+	if err != nil {
+		Logger.Println(err)
+		return err
+	}
+
+	// Delete all the events
+	err = db.Where("campaign_id in (?)", cids).Delete(&Event{}).Error
+	if err != nil {
+		Logger.Println(err)
+		return err
+	}
+
+	// Delete the campaigns
+	err = db.Where("user_id=?", uid).Delete(&Campaign{}).Error
+	if err != nil {
+		Logger.Println(err)
+		return err
+	}
+
+	return err
+}
+
 // CompleteCampaign effectively "ends" a campaign.
 // Any future emails clicked will return a simple "404" page.
 func CompleteCampaign(id int64, uid int64) error {
--- a/models/smtp.go
+++ b/models/smtp.go
@@ -238,7 +238,7 @@ func DeleteSMTP(id int64, uid int64) error {
 	return err
 }
 
-func DeleteSMTPByUserId(uid int64) error {
+func DeleteSMTPsByUserId(uid int64) error {
 	var ss []SMTP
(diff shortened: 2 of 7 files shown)


models/campaign.go
```diff
@@ -471,6 +471,44 @@ func DeleteCampaign(id int64) error {
...
+		return err
+	}
+
+	var cids []int64
+	for _, c := range cs {
+		cids = append(cids, c.Id)
+	}
+
+	// Delete all the results
+	err = db.Where("campaign_id in (?)", cids).Delete(&Result{}).Error
```
jordan-wright: Hmmm... I wish you didn't have to delete results this way. I'm wondering if it'd be better for me to create a SQL migration that adds a cascading delete condition to the results -> campaign relationship. What do you think?

models/campaign.go
```diff
@@ -471,6 +471,44 @@ func DeleteCampaign(id int64) error {
...
+	if err != nil {
+		Logger.Println(err)
+		return err
+	}
+
+	// Delete all the events
+	err = db.Where("campaign_id in (?)", cids).Delete(&Event{}).Error
```
jordan-wright: Same thing here. Maybe events should just have a cascading delete condition.

models/campaign.go
```diff
@@ -471,6 +471,44 @@ func DeleteCampaign(id int64) error {
 	return err
 }
 
+//DeleteCampaignByUserId deletes campaigns with a given user id
+func DeleteCampaignByUserId(uid int64) error {
```
jordan-wright: Minor nitpick - can we make these function names plural? We're deleting multiple campaigns, groups, etc. So if we could name it DeleteCampaignsByUserId, DeleteGroupsByUserId, etc.

models/template.go
```diff
@@ -213,3 +213,34 @@ func DeleteTemplate(id int64, uid int64) error {
...
+		return err
+	}
+
+	var tids []int64
+	for _, t := range ts {
+		tids = append(tids, t.Id)
+	}
+
+	// Delete attachments
+	err = db.Where("template_id in (?)", tids).Delete(&Attachment{}).Error
```
jordan-wright: I should also probably add a cascade delete condition for attachments.

models/smtp.go
```diff
@@ -237,3 +237,33 @@ func DeleteSMTP(id int64, uid int64) error {
...
+		return err
+	}
+
+	var sids []int64
+	for _, s := range ss {
+		sids = append(sids, s.Id)
+	}
+
+	// Delete all custom headers
+	err = db.Where("smtp_id in (?)", sids).Delete(&Header{}).Error
```
jordan-wright: I should probably also add a cascade delete condition to delete headers when the SMTP profile is deleted.

jordan-wright (changes requested):
Hey there!

Thanks for sending this in! I've added some notes in here, but overall I'm wondering if this would be far easier if I added some SQL migrations first that enforced a cascading deletion of child objects.

Also, I'd like to ensure that only admins can delete other users. Right now, this will be difficult since we don't have proper roles in Gophish. So until we make this happen, we might just have to do something manual like ensuring the username is "admin" or something. Otherwise, if any user can delete any other user, that'd lead to problems.

jordan-wright:
Thanks again for sending this over. We'll work our way towards merging this in. I'll see about adding those SQL migrations in the new few days or so. Then we can add a conditional to ensure only admin(s) can delete users. Finally, we'll merge the other PR's almost in reverse order working our way back to this one.

Awesome job getting to this so quickly and for providing good test coverage. Great stuff!

shuheiktgw:
Thank you for reviewing this PR! 😄 
I totally agree that it is better to add a cascading deletion!

So I should probably leave this PR until you figure out the other changes?
Please let me know if you need any other help to implement a user deletion feature! 👍

jordan-wright:
Yeah, let's go ahead and leave this PR as-is until I can add the SQL migrations. I'll need to consider them a bit more, since SQLite unhelpfully doesn't allow `ALTER TABLE` commands to include foreign key edits. So I'll basically need to do the rename -> create -> move -> delete flow, which I'm not excited about 😐 

But, that's still the best way to move forward, since we'll be letting the db do the heavy lifting for us.

I'll keep you posted. Just a heads up, this probably won't make it into the upcoming release, since I'm planning a release this week. But, I'm also going to start working to have a more frequent release cycle so this won't hopefully won't sit in source too long before it's included in a release 😄 

Thanks again for all the great work here!

PR #920, opened 2018-01-10
Refactor GetUserByUsername method not to suppress an error

shuheiktgw:
## Overview
`GetUserByUsername` method in `user.go` suppresses an error and that makes it really hard to use the method. I believe the method itself should focus on fetching a user by name, and a client who uses the method should decided how to handle the error the method returns.

So I refactored the method not to suppress an error, and simply return it.

Thank you! 👍 

# FYI
Once this PR is merged, some tests in another PR I made (https://github.com/gophish/gophish/pull/919) should be broken, so I'll fix them after this PR is merged.

### DIFF
--- a/models/models_test.go
+++ b/models/models_test.go
@@ -100,6 +100,18 @@ func (s *ModelsSuite) TestGetUser(c *check.C) {
 	c.Assert(u.Username, check.Equals, "admin")
 }
 
+func (s *ModelsSuite) TestGetUserByUsernameWithExistedUser(c *check.C) {
+	u, err := GetUserByUsername("admin")
+	c.Assert(err, check.Equals, nil)
+	c.Assert(u.Username, check.Equals, "admin")
+}
+
+func (s *ModelsSuite) TestGetUserByUsernameWithNotExistedUser(c *check.C) {
+	u, err := GetUserByUsername("test user does not exist")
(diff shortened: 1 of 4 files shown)


models/models_test.go
```diff
@@ -100,6 +100,18 @@ func (s *ModelsSuite) TestGetUser(c *check.C) {
 	c.Assert(u.Username, check.Equals, "admin")
 }
 
+func (s *ModelsSuite) TestGetUserByUsernameWithExistedUser(c *check.C) {
```
jordan-wright: Super minor nitpick, can we change these to "WithExisting" instead of "WithExisted"?

jordan-wright (changes requested):
I agree with this. You're right - it's almost certainly better practice to just pass the error up and let the upper layer decide how to handle it.

One minor nitpick with the test names and I'll be happy to merge this in.

Thanks for sending this over!

models/models_test.go
shuheiktgw: Thank you for your review! 
I just fixed the names, so please let me know if you have any comments on this PR! 👍

jordan-wright (approved):

jordan-wright:
No idea what happened with the odd indentation in `models_test.go` when resolving the merge. Oh well, it'll get fixed in the next fmt when I change that file again. The code still looks good, so I'll merge it in.

PR #921, opened 2018-01-10
Remove redundant return nil statement in user.go

shuheiktgw:
## Overview
In `user.go` some methods has redundant return nil statement, so I just simply removed them and add tests! Thank you! 👍

### DIFF
--- a/models/models_test.go
+++ b/models/models_test.go
@@ -106,6 +106,24 @@ func (s *ModelsSuite) TestGetUserDoesNotExist(c *check.C) {
 	c.Assert(u.Username, check.Equals, "")
 }
 
+func (s *ModelsSuite) TestGetUserByAPIKeyWithExitingAPIKey(c *check.C) {
+	u, err := GetUser(1)
+	c.Assert(err, check.Equals, nil)
+
+	u, err = GetUserByAPIKey(u.ApiKey)
+	c.Assert(err, check.Equals, nil)
+	c.Assert(u.Username, check.Equals, "admin")
+}
+
+func (s *ModelsSuite) TestGetUserByAPIKeyWithNotExitingAPIKey(c *check.C) {
+	u, err := GetUser(1)
+	c.Assert(err, check.Equals, nil)
+
+	u, err = GetUserByAPIKey(u.ApiKey + "test")
+	c.Assert(err, check.Equals, gorm.ErrRecordNotFound)
+	c.Assert(u.Username, check.Equals, "")
+}
+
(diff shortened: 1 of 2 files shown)


models/models_test.go
```diff
@@ -94,12 +94,36 @@ func (s *ModelsSuite) createCampaign(ch *check.C) Campaign {
...
 	c.Assert(u.Username, check.Equals, "admin")
 }
 
+func (s *ModelsSuite) TestGetUserDoesNotExist(c *check.C) {
+	u, err := GetUser(100)
+	c.Assert(err, check.Equals, gorm.ErrRecordNotFound)
+	c.Assert(u.Username, check.Equals, "")
+}
+
+func (s *ModelsSuite) TestGetUserByAPIKeyWithExitingAPIKey(c *check.C) {
```
jordan-wright: Hi there!

Did you mean TestGetUserByAPIKeyWithExi**s**tingAPIKey? 😄

jordan-wright: (note the "s" in Existing)

models/user.go
```diff
@@ -15,21 +15,15 @@ type User struct {
 func GetUser(id int64) (User, error) {
 	u := User{}
 	err := db.Where("id=?", id).First(&u).Error
-	if err != nil {
-		return u, err
-	}
-	return u, nil
+	return u, err
```
jordan-wright: Ah, yes, this particular code "smell" is probably all over the Gophish codebase. It really was just a preference at the time, since I just didn't feel right returning whatever error I was getting from gorm. Having the return nil made it feel a bit safer to me, even though functionally it didn't make a difference. And I agree that this is cleaner. 👍

jordan-wright (changes requested):
This is looking good! Just one super minor thing in the test names. Otherwise, I dig it! Thanks for sending this in!

models/user.go
shuheiktgw: To me it seems a bit redundant, but if you feel it is safer, I'm happy to remove some commits and just leave the test cases! So please let me know what do you think! 😄

models/models_test.go
shuheiktgw: Oops, sorry and thanks for pointing it out! 🙇

models/user.go
jordan-wright: Nah, no worries. I agree completely that this is a cleaner look.

jordan-wright (approved):

jordan-wright:
Thanks again for sending this in! Great work!

PR #926, opened 2018-01-12
Fix to raise error when trying to register a duplicate username

shuheiktgw:
Sorry, this is caused by my change by https://github.com/gophish/gophish/pull/920 .... 🙇 
Now Gophish does not raise any errors when trying to register a duplicate username, so I fix it to raise a proper error.


jordan-wright:
Good catch - thanks for sending in this fix!

PR #930, opened 2018-01-18
Separate user model test cases into a independent test file

shuheiktgw:
Hi, I personally believe that it would be easier to understand and beautiful if each model has its own test file. So in order to show an example I separated User model's test into a separate file!

If you like it, I'll work on the other ones and add more tests for their public methods!
Thank you! 👍


jordan-wright:
Yeah, this is Go best practice. I didn't do it originally but you'll see that I started the work with `maillog_test.go`.

Good stuff! Thanks for sending this over.

PR #938, opened 2018-01-24
Forward Email responses

GregCFTS:
To see if a user has replied to an email, is it possible to have emails forwarded to another email address

Spam Email to User-> User Receives -> User replies to Email -> Users' reply is forwarded to another mailbox so we can read it

Cheers


jordan-wright:
Hi @GregCFTS,

It appears you opened a pull request instead of an issue, so I'm going to go ahead and close it.

To answer your question, I think what you're looking for is called a "Reply-To" header and can be set using the "custom headers" functionality when creating a sending profile. Hope this helps!

Issue #955, opened 2018-02-03
If subject = blank don't send "SUBJECT:" to SMTP server


p292:
Not an issue but a small feature request.
If Subject is left blank in a campaign, could gophish omit sending "SUBJECT:" to the SMTP server?

What I think happens currently:
Subject set to blank in template, SMTP receives:
MAIL FROM: phisher@phish.com
RCPT TO: victim@victimcorp.com
DATA
TO: Firstname Lastname
FROM: Mr Phisher
SUBJECT:

Hello You Phished etc

Ideally if subject left blank it'd be nice if gophish could send server:
MAIL FROM: phisher@phish.com
RCPT TO: victim@victimcorp.com
DATA
TO: Firstname Lastname
FROM: Mr Phisher

Hello You Phished etc

Reason: Some companies add (OUTSIDE MAIL) to subject, if subject is not sent then this doesn't get appended.

Thanks

jordan-wright:
Hi there! Thanks for reaching out. This seems reasonable and easy to implement. Looking through the RFC, I show the subject being an optional header, so it seems to make sense that if it isn't provided, we'll just exclude it entirely.

truekonrads:
PR submitted

PR #961, opened 2018-02-07
If Subject is empty, don't set the header

truekonrads:
Implements https://github.com/gophish/gophish/issues/955 - if subject is empty, don't send it at all

### DIFF
--- a/models/email_request.go
+++ b/models/email_request.go
@@ -96,7 +96,12 @@ func (s *SendTestEmailRequest) Generate(msg *gomail.Message) error {
 	if err != nil {
 		Logger.Println(err)
 	}
-	msg.SetHeader("Subject", subject)
+	// don't set the Subject header if it is blank
+	if len(subject) == 0 {
+		Logger.Println("Subject is empty and the Subject header was not set")
+	}else {
+		msg.SetHeader("Subject", subject)
+	}
 
 	msg.SetHeader("To", s.FormatAddress())
 	if s.Template.Text != "" {
--- a/models/email_request.go
+++ b/models/email_request.go
@@ -97,9 +97,7 @@ func (s *SendTestEmailRequest) Generate(msg *gomail.Message) error {
 		Logger.Println(err)
 	}
 	// don't set the Subject header if it is blank
-	if len(subject) == 0 {
-		Logger.Println("Subject is empty and the Subject header was not set")
-	}else {
+	if len(subject) != 0 {
 		msg.SetHeader("Subject", subject)
 	}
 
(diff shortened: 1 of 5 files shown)


models/email_request.go
```diff
@@ -96,7 +96,12 @@ func (s *SendTestEmailRequest) Generate(msg *gomail.Message) error {
 	if err != nil {
 		Logger.Println(err)
 	}
-	msg.SetHeader("Subject", subject)
+	// don't set the Subject header if it is blank
+	if len(subject) == 0 {
+		Logger.Println("Subject is empty and the Subject header was not set")
```
jordan-wright: I'd say we don't have to worry about logging this for now. In the future, I might have some validation like you'll find in email clients where maybe before the user saves the email template it'll validate if they want to have an empty subject.

jordan-wright (changes requested):
Hi there!

Thank you *so much* for taking this on. This is definitely the approach I'd take on this, too, so this is great stuff.

I'd suggest just some minor changes. If we remove the logging line, we can make the conditional in `email_request.go` even cleaner, which would be a plus.

Also, you might consider adding some simple tests to validate that this is working correctly. You should be able to find solid examples in the `_test.go` files that can be copied and modified to make this super easy - let me know if you run into any issues and I'll be happy to step you through it.

Finally, I'd encourage you to make sure you `gofmt` your code. If you're using an IDE for Go, a good practice is to install [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports) and have it run on-save so your code is automatically formatted every time it's saved which is 🔥 .

This is looking great, and thanks again for taking the time to take the initial pass here.

models/email_request.go
truekonrads: Hey,

Thanks for your feedback. Will make the changes. My original plan was to have a checkbox in Email template which would say “don’t send subject” that would grey out the field. However, that would mean differentiating between the case where a blank subject should be sent and where none should be sent at all. That would mean schema change unless the subject would have some magic marker like —GOPHISH-DONTSENDSUBJECT— which felt hacky and I didn’t feel like doing a schema change.

truekonrads: Hi, quick question: how do I run the test suite, for some reason go test wasn't helpful:

Running tool: C:\Go\bin\go.exe test -timeout 30s github.com\gophish\gophish\models -run ^$

ok  	github.com/gophish/gophish/models	0.110s **[no tests to run]**
Success: Tests passed.

jordan-wright: I wouldn't worry about the checkbox - it seems much more usable to just alert when the subject is blank and even going so far as to alert is debatable - we might as well as just trust the user to make the choices they want to make. Something we'll add on later if we get complaints 😄 

To run the test suite, try going into the models directory and running `go test -v ./...`. It usually works pretty well.

Let me know if you run into any issues! Thanks again for all your work on this.

truekonrads:
I added the test suite, hope this works!

jordan-wright (approved):

jordan-wright:
Looks good to me! Thanks again for taking this on- great work!

PR #1003, opened 2018-03-10
Remove redundant type declarations in composite literals

shuheiktgw:
I just removed a few redundant type declarations!
Thank you! 👍

### DIFF
--- a/controllers/api_test.go
+++ b/controllers/api_test.go
@@ -63,8 +63,8 @@ func (s *ControllersSuite) SetupTest() {
 	// Add a group
 	group := models.Group{Name: "Test Group"}
 	group.Targets = []models.Target{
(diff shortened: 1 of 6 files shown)


jordan-wright:
Hi @shuheiktgw, thanks for sending this in!

I have no doubt this pattern is seen all over Gophish. To be honest, I didn't know it was even possible to avoid putting the types in a struct like that. TIL!

However, I don't think I'll merge this one in, only because being explicit (and maybe even redundant) simply reads better to me. It's a personal preference but I feel that by being explicit I don't have to track back up the struct to remember what type of things I'm putting in this list.

That said, I appreciate you sending this in and encourage you to keep an eye out for any future fixes we can make! Thanks again!

PR #1014, opened 2018-03-19
Adding "Report Email" Support

jordan-wright:
This PR builds on the work from @S0larflare to add the ability to report emails in Gophish.

**Note: No mechanism is currently provided in Gophish to actually report emails. This functionality is only possible with an external email client add-on.**

image


S0larflare:
I really like that. The check looks far better than the true/false and I had duplicated the icon that long ago I had completely forgotten about it, much better having it match the dashboard. I'll freely admit that I didn't know how to replicate an effective test, I wasn't sure what was being done in yours (I'm not particularly good with Go, still learning...). Thanks for getting this merged, just updated my test instance to this and it works great. I'll get to work sorting an outlook add-in that can be used by everyone for reporting when I get a bit of time!

PR #1069, opened 2018-05-17
Increase the size for Landing Pages and Templates in MySQL.

tamtu:
Currently if you are using MySQL the columns for Landing Page and Template contents are limited in size by their type - TEXT. This means that if you try uploading a page larger than 65535 characters it will get silently cut off.

It is easy to fix this in the console by modifying the type:

`alter table pages modify html longtext;`

However it would of course be best to have this done automatically. These are the two places that I've found this to be configurable.

### DIFF
--- a/db/db_mysql/migrations/20160118194630_init.sql
+++ b/db/db_mysql/migrations/20160118194630_init.sql
@@ -2,11 +2,11 @@
 -- +goose Up
 -- SQL in section 'Up' is executed when this migration is applied
 CREATE TABLE IF NOT EXISTS users (id integer primary key auto_increment,username varchar(255) NOT NULL UNIQUE,hash varchar(255),api_key varchar(255) NOT NULL UNIQUE );
-CREATE TABLE IF NOT EXISTS templates (id integer primary key auto_increment,user_id bigint,name varchar(255),subject varchar(255),text text,html text,modified_date datetime );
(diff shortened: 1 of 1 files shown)


tamtu:
SQLite does not have multiple TEXT types, but MySQL has, with differing lengths. There would be more logic required to set the proper value.

jordan-wright:
Hi there!

Thanks for sending this over. I agree that changing the field to something like `LONGTEXT` would be a good change.

To get this added, you'll need to create a new database schema migration file. You'll see examples of these in the [migrations folder](https://github.com/gophish/gophish/tree/master/db/db_mysql/migrations). Each of these files is timestamped so our migration tool (`goose`) knows which migrations to apply and in what order.

To create a new migration file, you'll want to install [`goose`](https://bitbucket.org/liamstask/goose). Then, in the `db/` folder, you can run the following:

```
goose -env production -path db_mysql create 0.6.0_new_migration sql
```

You'll see output like this:

```
goose: created go/src/github.com/gophish/gophish/db/db_mysql/migrations/20180520183244_0.6.0_new_migration.sql
```

Then, you can edit that file to make the table alterations.

Hope this helps!

jordan-wright:
I've gone ahead and added a migration in da6091e0212782f8784fb54addca343484fda301 to help resolve this issue. Thanks again for reporting this!

Issue #611, opened 2017-05-03
Throttle email delivery / email send interval


HarryR:
When sending email out to a large block of users it would be good to be able to schedule the e-mails to be sent out gradually, options for this could include:

  * N emails, delivered at regular intervals between Begin and End dates
  * Static day of N seconds, or random delay between X and Y

This should help email deliverability over time and make it more difficult to reject a batch of emails, it would also allow multiple campaigns to run at the same time with different delays.

S0larflare:
I'm working on this one at the moment (mk3. to be precise). The way I am making it will allow to specify an end date, the size of the batch and the interval between batches (this is static rather than randomised). I've also got a randomisation function so that it isn't going to everyone in order down a list.

Time to work on it has been a bit tight and due to some parts of GoPhish changing I've had to redo a few things differently, and there is one small issue, but I am hoping to get it done 'soonish'.

HarryR:
Ah ok, that sounds promising.

This would be an awesome feature because we can have many campaigns spread out over a month or two.

bringcoffeebringbeer:
I would love to see this.  I use Groups to track progress by Department - except that the whole department is getting the email at the same time!  

Love the system - thanks for what you do!

liamosaur:
+1 for this feature.

My issue isn't being able to spread out emails over long time periods, it's being able to throttle the send rate so I don't end up with rate-limiting email rejections from the 3rd party SMTP server we use.

jonfurmanski:
+1 for this feature

Also, not to keep expanding further, but if we can throttle either via rate limit or time based, it would be nice to specify what hours of the days the emails can be sent. I tend to like to send the emails during business hours.

teddielong:
+1 - we have issues when sending emails as all are received at the same time (unless we manually group which is a nightmare).
Any update on timescales?  Thanks again for this, it's a great tool.

cptwin:
Looking into this as well, did anyone else manage to create a solution?

jordan-wright:
Refactoring the mailing system is a large task that I'm doing carefully. I don't have a timeline for it yet, other than saying it's the highest priority big item on my list.

cptwin:
How would I go about requesting access to create a pull request? I've created a simple solution that would work in the interim.

thryb:
Not related directly to gophish but we ran into that issue during a campaign. If you guys use postfix you can add the following to your main.cf which will delay the time between emails:

Time is in seconds.
[...]

PR #1090, opened 2018-05-30
Added support for timed campaigns

c-f:
Hi Jordan, 
at first: **awesome work** and a very good reference project when starting a new go app.
In order to give something back to the community i made a bunch of changes - starting with this PR.

Sumary:
> This PR creates support for setting a delay between sending emails per campaign. The time (seconds) can be defined when creating a new campaign.

References:
- Reference Issues to this PR: #110 , #611 
- Open similar PR: #482 ( from oct. 2016)

Notes:
> The timeouts for different SMTP Servers varies tremendously. Therfore for sending each mail will create a new connection with the server. 
----------------------------

As this is my first PR i'm curious about feedback.
Best regards :)


jordan-wright:
Hi @c-f! Thanks for getting in touch, and thanks for sending this over!

At a high level, the code itself looks pretty good. There might be some tweaks here and there but for a first pass I'm super impressed! Well done.

What I'd like to chat about is the design for sending delayed emails in general. Right now, it seems that we're asking for the number of seconds between each email. This makes it a bit trickier to implement as you've seen because we have to account for handling batches of delayed emails which isn't the way the mailer was built.

Instead, what do you think about this approach - we already have a `launch_date`, so we could create a new date field, say, `send_by_date` that is the date by which all emails in the campaign should be sent.

This has a few benefits:

* We can delays for each email when the campaign is created and add that to the `maillog` information
* Then, the mailer which runs every minute will pick up only the emails that need to be sent during that minute window and send them off - no need to adjust the mailer at all
* Everything else (backoffs, errors, and such) will "just work" since we're abstracting away from the mailer and just fiddling with the `send_date` field of the maillog
* We can trivially add a "random" checkbox or something to make this logic have random delays instead of uniform delays with very little changes.

The only thing we'd lose with this other approach is the ability to have per-second delays since the mailer would still be running every minute. But, from my experience, it seems that if I want delays I'm looking more at minutes/hours/days, not seconds since second-long delays might just happen naturally due to the way email works 😛.

I could be missing something so let me know if I'm wrong here. But, if you like this approach then I'd be more than happy to list out how to approach it. It might not be able to re-use a lot of the code we have here since it's a pretty different approach but I'll of course be happy to help any way I can.

c-f:
Thanks for the feedback so far :) 
I aggree, that your idea fits better in general. 

If you'll provide some more information - i would be happy to help. 

**Notes**:
> Regarding the PR - it simply bypasses common spam filters, which will count the amount of mails in a given period and mark them if necessary. This simple approach by delaying each mail (e.g. 30secs), had worked just fine for multiple scaled campaigns.

Best regards :)

krasian:
Awesome implementation idea gents!  I've been waiting for something like this.  It may not be exactly what I want, but it will at least address the need.  I can work with that!  Can't wait to try this out!!!  Keep up the great work!

jordan-wright:
Hey @c-f!

Awesome! Sounds like we're on the same page then. I'd say for now, let's focus on the simple case of just having a "send by date", where each email is spaced apart equally. Once this plumbing is in place, you'll see that adding the random delays to this will be a super simple piece of cake.

The entire goal of this is to update our campaign objects to have another field, `send_by_date`, that tells Gophish to send all the campaign emails by this date.

Here's how I'd implement this if it helps get you started:

#### Database Migrations

I'd start by making the database migrations. I noticed in this pull request the migrations are for v0.7.1, but we're developing for the v0.7.0 release so as a small note we'd want to update that.

The migrations themselves would be a simple `ALTER TABLE` to add a column `send_by_date` to the `campaigns` table.

#### UI

The UI should hopefully be pretty simple. I'd just add a date field probably next to the launch date. I'd have the labels for the fields changed to compliment one another. Something like "Launch Campaign At" and "Send Emails By".

This is similar to what you have in the pull request since we'd just assign that form field to the `send_by_date` field when the request is sent off.

#### Backend

So now that we have the data, we want to make the magic happen. All of this will happen in `models/campaign.go`. The first step will be to update the `Campaign` and `CampaignSummary` structs to have the SendByDate field.

Next, we'll update `PostCampaign` to handle the logic of setting the right date when [creating the `Maillog` objects.](https://github.com/gophish/gophish/blob/420410b52c6d7ada6ac675a42d36ef0c84431f6f/models/campaign.go#L481).

The ideal way to do this will be to determine the **number of seconds after our `launch_date` that we want to send our email**. We can use this `launch date + offset` as the `SendDate` in our maillogs that we generate.

Here's what this would look like:

* First figure out if we have a `send_by_date` set. If we don't, that means we want emails sent immediately with no delay. That means our offset would be 0. We'd just set the launch date as the `SendDate` in our maillogs which is what happens now 😄 
* If we _do_ have a `send_by_date`, then we need to figure out our spacing in second. This would probably be something like `(send_by_date - launch_date).Seconds / number_of_emails`. Then, we could build emails at a super high level like this:

```
secondsBetweenEmail = [calculate the offset]
sendOffset := 0
for _, t := range g.Targets {
    sendOffset += secondsBetweenEmail
    maillog.SendDate := campaign.LaunchDate + sendOffset
```

This means that the first email would be sent immediately, the second email would be send at `launchdate + offset` and the third email would be sent at `launchdate + offset + offset`, etc.

But you'll see we have a problem- right now we call  `GenerateMaillog` which just takes a campaign and result. It doesn't let us set a senddate.

This is where, IMO, I made a mistake 😄

You'll see that the [only place](https://github.com/gophish/gophish/search?q=GenerateMailLog&unscoped_q=GenerateMailLog) we actually use `GenerateMaillog` is when building a campaign, so I would personally remove it and just move that logic into the campaign for loop instead. 🤷‍♂️ 

When all this is done, we'll have maillogs with the right send date that spaces them out as needed. Our mailer will do all the heavy lifting of actually sending the emails out at the right time. We might consider adding logic in the mailer to make sure that if an error occurs and we have to backoff, that we're not backing off past the campaign `send_by_date`. If that would happen, I'd prefer to just error out the email and call it a day, since I see that `send_by_date` being a "don't send emails after this" date. Though, this is open to discussion!

#### Looking Ahead

Finally, we'll consider the random interval case. You'll notice that this is super easy! All we'd have to do to make this happen would be to add logic to see if that option was checked and, if it was, just calculate our number of seconds offset for each email to be `random(0, (send_by_date - launch_date))`. Piece of 🍰 !

I hope this helps. Fortunately, this looks like it should be pretty straightforward and keeps us from having to make changes to the mailer. It's just a matter of setting the right date and our mailer will take care of the rest!

I'm happy to help throughout this process any way I can. I think the work you've done so far will translate really well to this new approach.

Please don't hesitate to let me know if you have any/all questions! Thanks again so much for taking this on and working on this. It's a feature that's been asked for many times, so you'll be the hero to a lot of awesome community members 😄

jordan-wright:
Hey @c-f!

Just wanted to check-in- are you still wanting to take this on? Totally understood if you don't have time to knock it out, just wanted to check since I should have some time soon(ish) and I could probably get it taken care of.

However, I love seeing community contributions so if you're still interested in making this happen, let me know and I'll hold off.

Thanks again for all your hard work!

c-f:
Hey @jordan-wright ,
sorry for the delay - have been busy weeks. I'm currently working on this and i'll push it as soon as i have time to finish it. Should be done before july. 

Best regards :)

jordan-wright:
That sounds great! Please don’t hesitate to push commits as you go along. I’m always happy to help test and provide feedback early and often.

Thanks so much!

cjundy26:
Hi All,

Just checking in to see where this is at!

Thanks again for a great product Jordan!

jordan-wright:
Hey @c-f!

I have some time this weekend, so I'm going to go ahead and merge this into a branch and make the adjustments mentioned above.

If you'd like, I'd be happy to tag you on the commits so you can see how I implemented the changes.

Thanks again **so much** for all your work here. This really help lay the groundwork for getting timed campaigns implemented, which is super exciting!

PR #1154, opened 2018-08-01
Added AWS terraform template

kontez:
Hey all,

I created a terraform template for gophish.
This lets users instantly boot up an aws machine with gophish already installed and configured.
All they have to do is run `terraform apply` and it will then output the ip of the aws box.

cheers

### DIFF
--- a/gophish-main.tf
+++ b/gophish-main.tf
@@ -1,0 +1,112 @@
+provider "aws" {
+  region     = "eu-central-1"
+}
+
+// details of the aws instance
+resource "aws_instance" "example" {
+  ami = "ami-40d5672f"
+  instance_type = "t2.micro"
+  vpc_security_group_ids = ["${aws_security_group.instance.id}"]
+  key_name = "${aws_key_pair.auth.id}"
+
+  tags {
+    Name = "phishing-machine"
+  }
+
+  user_data = <<HEREDOC
+    #!/bin/bash
+    yum update -y
+    yum install wget -y
+    yum install unzip -y
+    su ec2-user
+    cd /home/ec2-user/
+    wget https://github.com/gophish/gophish/releases/download/v0.6.0/gophish-v0.6.0-linux-64bit.zip
+    unzip gophish-v0.6.0-linux-64bit.zip
+    cd gophish-v0.6.0-linux-64bit
+    sudo openssl req -newkey rsa:2048 -nodes -keyout gophish.key -x509 -days 365 -out gophish.crt -subj "/C=DE/ST=Example/L=Example/O=example/OU=Cyber"
+    echo '{
+      "admin_server" : {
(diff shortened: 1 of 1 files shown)


jordan-wright:
Awesome, thanks @kontez! I apologize for the late response - not sure how this slipped past me.

I'll give this a shot here in a few days and, if it works as intended, I'll get it merged. 🎊

kontez:
No worries mate!
Let me know if I can be of any help :)

gophish-main.tf
```diff
@@ -0,0 +1,112 @@
+provider "aws" {
+  region     = "eu-central-1"
```
jordan-wright: We might consider avoiding specifying a default region here. I could see people surprised that the instance is spun up in `eu-central-1`. Instead, it [looks like](https://www.terraform.io/docs/providers/aws/) the `AWS_DEFAULT_REGION` environment variable is respected by Terraform, so perhaps we could just document that that variable is required?

What are your thoughts on this?

gophish-main.tf
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  user_data = <<HEREDOC
+    #!/bin/bash
+    yum update -y
+    yum install wget -y
+    yum install unzip -y
+    su ec2-user
+    cd /home/ec2-user/
+    wget https://github.com/gophish/gophish/releases/download/v0.6.0/gophish-v0.6.0-linux-64bit.zip
```
jordan-wright: I've set up some Cloudflare magic to make it easier to get the latest version if you're interested. The format is `getgophish.com/releases/latest/<os>/<arch>`. So, we could change this to something like:

`wget https://getgophish.com/releases/latest/linux/64 -O gophish-linux-64bit.zip` and then update the call to `unzip` further down to take into account this new ZIP filename.

jordan-wright (changes requested):
Hey @kontez,

Thanks again for submitting this! I'm admittedly not a Terraform user, so everything _seems_ good to me, I've just added a couple of small suggestions for your review.

Let me know if you have any questions!

gophish-main.tf
kontez: Yes, you are right. Didn't think of this when I wrote the template. I'll change it like described [here](https://www.terraform.io/docs/providers/aws/#environment-variables)

gophish-main.tf
kontez: Updated this, thank you :)

kontez:
I updated the template with the two suggestions you made!
Let me know if I can help you further :)

cheers

securitygeneration:
It would be cool if the script could have the option of setting up a local SMTP relay too (i.e. you end up with a fully self-contained instance of GoPhish)

davisjackie321:
Yo

alex4108:
Did this PR die?

PR #1184, opened 2018-09-02
Add Support for Timed Campaigns

### LINKED ISSUE #110: Add delays between email deliveries

### LINKED ISSUE #611: Throttle email delivery / email send interval

jordan-wright:
This PR builds on the initial work from @c-f in #1090 to add support for timed campaigns.

The UI changes are minimal:

screen shot 2018-09-02 at 11 01 57 am

Like the Launch Date, this field is a datepicker with the following description:

screen shot 2018-09-02 at 11 09 34 am

The rules are simple:

* If no "send emails by" date is specified, the emails are sent immediately according to the launch date (this was previously the "Schedule" field.)
* If a "send emails by" date is specified, the emails are spaced evenly between the launch date and the send emails by date.

> Note: We still backoff emails in the case of temporary errors. We don't consider the "send emails by" date when doing these backoffs, since I figured that would cause confusion and unexpected email failures. So this date is only for the _initial_ attempt to send the email. Backoff retries may be after this date.

This allows emails to be spaced out throughout a period of time. 🔥 

Closes #110 
Closes #611

The next step will be to implement randomized sending between these intervals. This should be an easy addition, since I'll just add a checkbox for the option only when the "send emails by" date is active (otherwise, we wouldn't know how long to make our random waits 😉)


static/js/src/app/campaign_results.js
```diff
@@ -357,14 +357,15 @@ var renderDevice = function (event_details) {
...
+    console.log(data)
     record = {
         "first_name": data[2],
         "last_name": data[3],
         "email": data[4],
         "position": data[5],
         "status": data[6],
-        "send_date": data[7],
-        "reported": data[8]
+        "reported": data[7],
```
jordan-wright: Whoops! This was a bug that's apparently been there since we introduced reporting. I went ahead and fixed it here.

jordan-wright:
@c-f, here are the changes that I made to implement https://github.com/gophish/gophish/pull/1090#issuecomment-394060116. It looks like a ton of changes, but really the bulk of everything is in [campaigns.go](https://github.com/gophish/gophish/pull/1184/files?w=1#diff-eeba1ca34493009d8e903058d694764e) if you're interested in taking a look! Everything else is really minor changes / testing stuff.

PR #1252, opened 2018-10-28
Add Go 1.10 and 1.11 to Travis

gliptak:


.travis.yml
```diff
@@ -3,6 +3,8 @@ sudo: false
 
 go:
   - 1.9
+  - 1.10
```
jordan-wright:
```suggestion
  - "1.10"
```

jordan-wright:
Thanks for sending this in! It looks like Travis CI is parsing the 1.10 as 1.1, which is related to [this bug](https://github.com/travis-ci/travis-ci/issues/9247). I've added a suggestion that should fix the issue if you want to double-check it 😄 

Thanks again for this!

gliptak:
@jordan-wright thank you for the pointer. change applied

jordan-wright:
LGTM! Thanks for the quick fix.

PR #1267, opened 2018-11-08
Update page.go to Re-Capture Passwords

GalDagan0:
fixed a bug when marking the capture password, saving and unmarking the capture password, the attribute does not comeback and the password will never be captured again for this template.

### DIFF
--- a/models/page.go
+++ b/models/page.go
@@ -46,7 +46,16 @@ func (p *Page) parseHTML() error {
 						input.RemoveAttr("name")
 					}
 				})
+			}else {
+				inputs := f.Find("input")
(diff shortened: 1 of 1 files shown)


jordan-wright:
Thanks for this @GalDagan0! Would you mind running `goimports` on the file to format it? It seems as though there may be some minor spacing issues.

Once that's done, I'll give it a shot and, if things look good, get it merged. Thanks again so much for sending this in!

jordan-wright:
Thanks for the quick update! This looks good to me. I'm going to merge it into a branch so I can add a test before putting it into master.

PR #1264, opened 2018-11-08
Change campaign results tooltip to use percentage instead of count.

mcvic1rj:
Fixes #1261 

Had to push the count of results to the the email_data object in order to continue to have the number in the donut display correctly.

PR #1272, opened 2018-11-11
Make Campaign Results Pie Chart Consistent with Dashboard

jordan-wright:
This PR updates the campaign results pie charts to match the dashboard, in that the tooltip will show the percentage, rather than the full counts for each event.

Thanks to @mcvic1rj for the fix!


mcvic1rj:
This fixes #1264 should we close the issue or wait until a release is built?

jordan-wright:
I went ahead and closed out the issue. Thanks for the reminder, and thanks again for helping out with the fix!

PR #1314, opened 2018-12-04
Add encryption for sensitive submitted data

NHAS:
Hi, I've added the ability to encrypt sensitive data submitted to gophish from those who have been phished.
This allows for the phisher to add public keys to gophish and then select them in the campaign creation menu.


auth/auth.go
```diff
@@ -97,6 +98,10 @@ func Register(r *http.Request) (bool, error) {
 	u.Hash = string(h)
 	u.ApiKey = GenerateSecureKey()
 	err = models.PutUser(&u)
+	if err != nil {
```
jordan-wright: Let's leave out changes to this file, since the only addition is an error check. We can keep things scoped to just the changes being made.

db/db_mysql/migrations/20160118194630_init.sql
```diff
@@ -2,6 +2,7 @@
 -- +goose Up
 -- SQL in section 'Up' is executed when this migration is applied
 CREATE TABLE IF NOT EXISTS users (id integer primary key auto_increment,username varchar(255) NOT NULL UNIQUE,hash varchar(255),api_key varchar(255) NOT NULL UNIQUE );
+CREATE TABLE IF NOT EXISTS public_keys (id integer primary key autoincrement, user_id integer NOT NULL, friendly_name varchar(255), pub_key blob NOT NULL);
```
jordan-wright: Let's create a new database migration for this instead of editing old ones. This makes it easier to keep track of changes that have been applied over time.

db/db_mysql/migrations/20160131153104_0.1.2_add_event_details.sql
```diff
@@ -2,6 +2,7 @@
 -- +goose Up
 -- SQL in section 'Up' is executed when this migration is applied
 ALTER TABLE events ADD COLUMN details BLOB;
+ALTER TABLE events ADD COLUMN key BLOB;
```
jordan-wright: Likewise here, let's move this into a separate migration for these changes.

models/group.go
```diff
@@ -212,6 +213,12 @@ func PutGroup(g *Group) error {
 	if err := g.Validate(); err != nil {
 		return err
 	}
+
+	record, err := GetGroupByName(g.Name, g.UserId)
+	if err == nil && record.Id != g.Id {
```
jordan-wright: Let's remove the checks to this new error you're introducing. I see the reasoning behind it, but it's better to keep these changes focused solely on the public key support you're adding.

models/page.go
```diff
@@ -138,6 +138,16 @@ func PostPage(p *Page) error {
 // Per the PUT Method RFC, it presumes all data for a page is provided.
 func PutPage(p *Page) error {
 	err := p.Validate()
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+
+	record, err := GetGroupByName(p.Name, p.UserId)
```
jordan-wright: As mentioned earlier, let's remove all the added checks in here, since it increases the scope of these changes. Also, I'm not quite sure what you're doing here, since you're getting a group when the user is editing a landing page?

models/result.go
```diff
@@ -49,7 +55,49 @@ func (r *Result) createEvent(status string, details interface{}) (*Event, error)
 		if err != nil {
 			return nil, err
 		}
-		e.Details = string(dj)
+
+		if EVENT_DATA_SUBMIT == status && c.PublicKeyId != 0 { // Zero is unset
```
jordan-wright: Minor nit - let's switch this around
```suggestion
		if status == EVENT_DATA_SUBMIT && c.PublicKeyId != 0 { // Zero is unset
```

models/result.go
```diff
@@ -49,7 +55,49 @@ func (r *Result) createEvent(status string, details interface{}) (*Event, error)
...
+			//Taken from crypto/cipher CFB example
```
jordan-wright: I'll have to review this a bit more closely. Tbh, I'm not a crypto expert so I'll take a look at the examples provided in the Golang source that you mentioned.

jordan-wright: Also, let's move this to a function in `publickey.go`. Perhaps creating a function like this (untested):
```suggestion
			func (pk *PublicKey) Encrypt(data []byte) ([]byte, error) {} 
```

models/smtp.go
```diff
@@ -203,6 +204,12 @@ func PutSMTP(s *SMTP) error {
 		log.Error(err)
 		return err
 	}
+
+	record, err := GetSMTPByName(s.Name, s.UserId)
```
jordan-wright: As mentioned before, let's remove these checks that aren't necessary for these changes.

models/template.go
```diff
@@ -137,6 +137,12 @@ func PutTemplate(t *Template) error {
 	if err := t.Validate(); err != nil {
 		return err
 	}
+
+	record, err := GetGroupByName(t.Name, t.UserId)
```
jordan-wright: As mentioned before, let's remove these checks that aren't necessary for these changes.

models/user.go
```diff
@@ -35,5 +35,9 @@ func GetUserByUsername(username string) (User, error) {
 // PutUser updates the given user
 func PutUser(u *User) error {
 	err := db.Save(u).Error
+	if err != nil {
```
jordan-wright: As mentioned before, let's remove these checks that aren't necessary for these changes.

jordan-wright (changes requested):

jordan-wright:
Hi @NHAS,

Thanks for sending this in! This is quite a large pull request, so it's going to take some time to work through. I've added some notes to get started.

But, at a higher level, what benefits do you see here besides being able to encrypt submitted data? Are there any other uses for these public keys that you can envision (it's ok if there's not).

Also, what are your thoughts around making decryption as easy as possible for users? Unless I'm missing something, it appears that there would be quite a bit of friction right now for users trying to decrypt submitted data. They would have to export the data and manually figure out how to use their private key to decrypt the data, and that kind of friction goes against the UX we've strived for so far.

I'm open to ideas, but we'll definitely want to figure out the decryption story before getting this exposed to everyone.

That said, this is a really interesting concept and I appreciate all the hard work you put into the pull request! You definitely took some time to make sure the bases were covered, which is always appreciated.

models/group.go
NHAS: I believe this issue isnt from my input, what I found during development of this is that put requests can modify the names of landing pages/groups/etc and set them to be the same name as other entries. 
However you are correct in that this should be a seperate merge on its own, rather than being lumped in with the public key stuff.

NHAS:
Hi, thanks for your quick response!

Currently I can't think of any extension for the public keys, other than being able to have different levels of paranoia for the encryption. 
I.e 
None = No public key, nothing encrypted.
Low = Just sensitive submitted data encrypted.
High = All data submitted is encrypted.

I've made a small command line tool that takes the exported raw results csv file and decrypts it given the public key. I can upload that to git if you want to have a look at it, although its not very polished. 
My personal bias is to not have the decryption done on the server side at all. 
However if it needs to be done for UX reasons I'd suggest having a box popup, when exporting anything with encrypted results, where the user can copy and paste their private RSA key. This would then be used for that decryption, and promptly flushed.

As Im not particularly familiar with forking in git, would you prefer me to add new commits to my fork and remerge?
Or is it better to go back and rewrite my commits?

models/result.go
NHAS: That's a good suggestion. As for the crypto side of things, if you'd like me to do a bit of a walk through of my thought process for it I'd be more than happy. 
I would suggest reading through the source first, so you spot anything I might bias you against with an explanation.

jordan-wright:
Hey @NHAS,

I see you've been adding commits to this pull request. Thank you for your continued work on this!

Let me know when you're ready for another code review, and I'll take a look at things. I didn't want to review things if you were still actively working on them.

NHAS:
I've just added the ability to decrypt events before downloading them, I'd say this meets your requested changes.
As such, I'd say now would be a good time for a code review.

TheGroundZero:
May I recommend adding an option to hash sensitive information using SHA1?
I'm thinking about running captured passwords thorugh [Have I Been Pwned](https://www.troyhunt.com/ive-just-launched-pwned-passwords-version-2/#cloudflareprivacyandkanonymity), which can be queried using SHA1-hashed passwords?

NHAS:
@TheGroundZero 
I agree that it would add some confidentiality to the passwords and is probably a good idea. 
The only problem with this is the fact that they would still need to be encrypted on the server, due to the fact unsalted SHA1 is unsuitable for storing passwords. 

To quote troy hunt:
"It doesn't matter that **SHA1 is a fast algorithm unsuitable for storing your customers' passwords** with because that's not what we're doing here, it's simply about ensuring the source passwords are not immediately visible."

TheGroundZero:
> @TheGroundZero
> I agree that it would add some confidentiality to the passwords and is probably a good idea.
> The only problem with this is the fact that they would still need to be encrypted on the server, due to the fact unsalted SHA1 is unsuitable for storing passwords.

My view is that you'd only capture sensitive data if this is approved by your client, your employer, ... and that you wouldn't store this data for a long time period (plus the system running GoPhish would be properly secured).

The use of SHA1 would add a (small) barrier to being able to view the data in plain-text, which would mean someone would need to actively perform an action to retrieve the original data, breaking any contract there would be with a client.

I would of course encourage the secure storage of any sensitive data on a system, including the GoPhish database.  
You could even combine an encrypted storage solution with (optionally) storing the sensitive data in a hashed format.

NHAS:
@TheGroundZero 
I definitely think this should be another pull request. But perhaps after this one gets merged, as I already have encryption working for plaintext credentials.


Then it would be, say a checkbox in the encryption settings.

NHAS:
@jordan-wright 
I've just updated the encryption to use the new api scheme. If you wanted to have a look at the code and give some feedback, that'd be great!

rmarks-jacobs:
Had this exact same idea tonight and asked one of my team members to look into whether we could use AWS CloudFront's support for field level encryption to satisfy this requirement as a stop-gap solution, but looks like you are close to having this native support added.  Thanks for the hard work!

NHAS:
@rmarks-jacobs 
Thanks!
If anyone on your team has experience with encryption and whatnot, it'd be good to get some feedback on what I've done.

Sevalel:
Hey i am planning to run a preventive phishing campaign in the hospital in which im working. But i have no right to know their password according to the headoffice (I guess they're right). Is there any way to encrypt these even while checking the "capture password" box?
I mean have instead of lets say for exemple "WHATEVER" as a password, somthing like WH****ER, in the "submitted datas" section?

TheGroundZero:
> 
> 
> Hey i am planning to run a preventive phishing campaign in the hospital in which im working. But i have no right to know their password according to the headoffice (I guess they're right). Is there any way to encrypt these even while checking the "capture password" box?
> I mean have instead of lets say for exemple "WHATEVER" as a password, somthing like WH****ER, in the "submitted data" section?

If you have no business with the password, I wouldn't know why you'd capture them in the first place.
Displaying passwords plain-text, even partially may/will have an impact on the confidentiality of the password.
Would your idea be to show the partial password to the user and ask whether this was in fact his/her password? If so, that would lead to a whole different discussion of revealing the identity of those who've fallen for the phish.

However, this whole discussion is taking this Pull Request off-topic.
The original PR by NHAS was to include encrypted storage of the captured data on the GoPhish server for security purposes.
My addition to that was to hash the passwords, so you can test them against HIBP or rainbow tables, without reading any clear-text password, adding confidentiality. Which would probably also be better served with a separate PR.

Sevalel:
Sorry about that, just delete this discussion right after if it's off-topic.

The aim of this campaign is to educate internet users in the hospital about phishing and the risks of such an activity. We want to show them that we can easily get their logs because they’re not suspicious enough and then explain them how to be more careful while using internet. 

But at the same time we do not want to know their password for confidentiality reasons, just the first one or two letters for example instead of the full clear-text password in "submitted data", to show them that a real hacker can get more. And they will be invited to change their password if they want to. 

Just wondering if such a thing was doable using the v0.7.1.

NHAS:
Is there anything more I can do to help with this merge request? @jordan-wright

J3rry367:
Are there any updates on this? Having the ability to have the credentials encrypted would be a huge help. I just came across this project and am thinking of implementing it, but it will be a big hurdle if the credentials cannot be encrypted. The way we currently do phishing is the submitted details GPG encrypted and emailed to the person/people running the campaign in order to allow them to weed out anyone who filled in blank creds and/or attempted to hack the phishing page during the campaign. As I understand if you do not capture the data submitted the target may have just clicked the submit button without filling out the form. I'd like to filter those out without compromising on security any more than necessary.

NHAS:
As for implementation this PR has the functionality of being able to provide a public key that the server will then encrypt all results with. Blank or not. 

I think the functionality for filtering out blank results doesn't fit the scope of this PR to add encryption. After downloading the results of a phishing campaign you could write a tool to filter them before sending I'd imagine.

J3rry367:
Yeah, I get that. I was just saying there needs to be a way to tell if the submissions are blank/faked which you can only do if you capture the info, but you don't particularly want to capture the info if it's not encrypted. Mostly just illustrating the need for this to be implemented.

l2k-ais:
Are there any more blockers for this @jordan-wright ? I'd love to see this merged.

NHAS:
Unfortunately I imagine my fork is now quite out of date. If there is still interest from @jordan-wright in this Im happy to update it in the coming months. 
But otherwise I'll probably close this as it hasnt gone anywhere.

NHAS:
Well, Im going to close this pull request as nothing has happen and nothing will probably happen around this.

Issue #448, opened 2016-11-23
Archiving past campaigns


S0larflare:
Currently, after an appropriate time has passed I am closing campaigns and then finally deleting them in order to not clutter the interface (and to make it a bit easier to find the new/current campaigns). However this means that the historic data of a completed campaign is then lost to the aether. Would it be possible to archive old campaigns and place them in their own tab/section. This way they would be viewable in the future if they are needed and don't get in the way of live campaigns.

jordan-wright:
I see an issue to simply order campaigns by the created date. Would that be sufficient?

S0larflare:
Personally, I would say no. Ordering would be useful regardless, and would help when going through both active and archived campaigns. But the advantage of being able to archive campaigns when they aren't needed any more would be shown when you have a large amount of campaigns. As I am currently making one campaign out of multiple smaller campaigns (the only real way to do some kind of randomised/staggered sending at the moment) I am ending up with a large amount of campaigns at the end of it. If completing these campaigns put them into a different place, so that they aren't front and centre but under their own tag/tab on the campaigns area, it would make it a lot easier to track live campaigns whilst still having the historic data for older campaigns.

tmgordon:
You can create a new user account when you want a fresh campaign list until this is implemented. I would like to see a way to transfer campaigns from one user to another though.

S0larflare:
I built this in to my prod build so long ago I forget that it’s not a feature. I created a separate tab on the campaign screen for active and archived campaigns, makes it a lot cleaner. I’ll update and make a PR for it when reporting gets merged

PR #1334, opened 2019-01-02
adding archived view for campaigns

S0larflare:
Pretty simple change, this adds a different view on the campaigns page which allows you to more easily identify campaigns which are currently active. Any which have been marked as completed will move to the 'archive' tab. Functionally, nothing different is happening (Gophish is still loading all the campaigns) this just loads them in to two separate tables each in their own tab for readability.

image

image

(names removed from my actual campaigns)

I've uploaded a new campaigns.min.js, but I probably generated that in a different way to you, so you might want to do that in your usual way.

EDIT: This covers #448. 26 months in the making...


jordan-wright:
Hey @S0larflare!

Sorry for the delay. I appreciate you sending this in. I'll give it a look through this weekend (if I don't get to it even earlier) and I'm sure it'll be good to get merged. 😄

PR #1386, opened 2019-03-05
Go min version

Ne0nd0g:
The `github.com/jinzhu/gorm` package uses `sync.Map` which requires go version 1.9


jordan-wright:
Eventually I just need to vendor these dependencies. Gorm has updated over time which slowly crept up the required version of Go needed for Gophish 🙁 

Anyway, this looks good. Thanks for the fix!

PR #1402, opened 2019-03-21
Wait for db

gfrancqu:
Hi,

just added a few line to wait for database connexion.

WHY ?

i'm building a phishing stack using docker-compose, i'm using mysql as database and the struggle is real when it come to wait for the database  initialization. Those few lines will simply wait for 50 seconds and keep trying to connect until the DB is up. 

Why don't I add this to my docker container ? I'm building a distroless container and don't want to include useless binaries and scripts (can add the Dockerfile to the source if you want)

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"io"
 	"time"
-	
+
 	"bitbucket.org/liamstask/goose/lib/goose"
 
 	_ "github.com/go-sql-driver/mysql" // Blank import needed to import mysql
@@ -17,6 +17,8 @@ import (
 
 var db *gorm.DB
 var conf *config.Config
+
+const DefaultMaxDatabaseConnectionAttempts int = 3
 
 const (
 	CampaignInProgress string = "In progress"
@@ -95,21 +97,25 @@ func Setup(c *config.Config) error {
 		return err
 	}
 	// Open our database connection
-	max_attempts := 10
+	// Open our database connection
+	var max_attempts = DefaultMaxDatabaseConnectionAttempts
+	if conf.MaxDBConnectionAttempts != 0 {
+		max_attempts = conf.MaxDBConnectionAttempts
+	}
 	i := 0
-	for{
+	for {
+		log.Warn("waiting for database to be up...")
 		db, err = gorm.Open(conf.DBName, conf.DBPath)
-		if err == nil{
+		if err == nil {
 			break
 		}
-		if err != nil && i > max_attempts {
+		if err != nil && i >= max_attempts {
 			log.Error(err)
 			return err
 		}
 		i += 1
 
 		time.Sleep(5 * time.Second)
-		log.Error("waiting for database to be up...")
 	}
 	db.LogMode(false)
 	db.SetLogger(log.Logger)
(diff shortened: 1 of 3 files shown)


models/models.go
```diff
@@ -94,7 +95,22 @@ func Setup(c *config.Config) error {
 		return err
 	}
 	// Open our database connection
-	db, err = gorm.Open(conf.DBName, conf.DBPath)
+	max_attempts := 10
```
jordan-wright: I'd recommend moving this to a constant at the top of the file. Something like `MaxDatabaseConnectionAttempts`. This will make it more clear where this is coming from and easier to modify later on if needed.

models/models.go
```diff
@@ -94,7 +95,22 @@ func Setup(c *config.Config) error {
...
+	i := 0
+	for{
```
jordan-wright: Please run `gofmt` on your code to ensure that formatting is kept consistent across the project. Something that's helped me a ton is to set my editor to automatically run `goimports` (basically `gofmt` + import management) every time the file is saved.

jordan-wright (changes requested):
Hi @gfrancqu,

Thanks for sending this in! I'd be ok adding this, since it shouldn't affect the default case where the database is already up and running.

I've added a couple of small nits to try and keep things consistent with the rest of Gophish. If you wouldn't mind knocking those out, I'd be happy to merge this in.

Thanks again for taking the time to send in your improvements!

gfrancqu:
Hi, I choose to make it configurable in the config.json. this would allow people to set the timetout according to their need

models/models.go
gfrancqu: made it configurable from the config.json, default value will be 0 which mean: only 1 try

models/models.go
gfrancqu: ran `gofmt` on the last push

jordan-wright:
Hey there!

I appreciate the quick response. I'd ask that we just move it to a `const` at the top of the file as opposed to a configuration value. I'm pretty selective about what gets added to the config, so that it doesn't become overwhelming.

This is a good example of a value that shouldn't need to change for any reasonable installation. If someone need more than 10 attempts to connect to the database, then they would likely be in a position to build Gophish from the source, changing that value themselves.

I also noticed that you ran `gofmt` on the code - thanks for doing that! It's looking good.

models/models.go
```diff
@@ -94,7 +97,21 @@ func Setup(c *config.Config) error {
 		return err
 	}
 	// Open our database connection
-	db, err = gorm.Open(conf.DBName, conf.DBPath)
+	i := 0
+	for {
+		log.Warn("waiting for database to be up...")
```
jordan-wright: One more super small nitpick - can we move this to line 111 (above the `i +=1`)? The benefit is that if the database connection succeeds the first time, then this message won't show up in the logs.

jordan-wright (changes requested):
Hi @gfrancqu,

Looking good! I added one final small request that will make the output cleaner for users when the database connection succeeds the first time. If you wouldn't mind moving that log line, then I think we'll be in a good position to merge the changes.

Great work! I'd be interested in chatting about getting an official Dockerfile created as a separate pull request (though, we can start with an issue to chat about details first), if you're interested.

models/models.go
gfrancqu: done!

jordan-wright (approved):

jordan-wright:
Looks good to me! Thanks again for sending this in, and being patient with all the nitpicks 😉

PR #1437, opened 2019-04-21
Add Dockerfile

wjwoodson:
Adds a Dockerfile and startup script.

`Dockerfile`:
* builds gophish from source on `golang:1.11`
* exports gophish binary and assets to `debian:latest` runtime image

`docker/run.sh`:
* sets values in config.json with any set env var: `ADMIN_LISTEN_URL`, `ADMIN_USE_TLS`, `ADMIN_CERT_PATH`, `ADMIN_KEY_PATH`, `PHISH_LISTEN_URL`, `PHISH_USE_TLS`, `PHISH_CERT_PATH`, `PHISH_KEY_PATH`, `CONTACT_ADDRESS`.
* starts gophish

Building from source _should_ enable CI to autobuild master, tags, etc. and push to public image repository -- I've been testing this here: https://hub.docker.com/r/wjwoodson/gophish

The final image can be run like so:
```
docker run -id \
    --name gophish \
    -e ADMIN_USE_TLS=false \
    -p 3333:3333 \
    -p 80:80 \
    gophish/gophish:latest
```


jordan-wright:
Thanks for this @wjwoodson!

Creating and hosting an official Docker image is something I've been meaning to do for a while now (I have the [@gophish](https://cloud.docker.com/u/gophish/repository/list) DockerHub organization made for this). In the past, there have been considerations as to whether or not we want to include a mail server in the image, but thinking more about it, I like this approach of just keeping Gophish separate.

I'll test this out and if it looks good will get it merged in. I wish that we didn't need `jq` to manipulate the config file. Instead, we could _probably_ update Gophish somehow to use environment variable overrides but that may be for future us to figure out 😛

Reedtechno:
Is there a reason for using Debian as the base image instead of something like alpine?  I was originally using one of the Debian based GoPhish Docker images but ended up building one based on alpine instead to reduce vulnerabilities and image size.  By using the Alpine base, I was able to reduce the image size by 40% and reduce the the 19 known vulnerabilities to 0.  

I am new to the container game so really just asking if I am doing something wrong :)  I have been trying to find some time to put together a Docker-Compose file with a reverse proxy to make certs and multiple domains easier.

jordan-wright:
I’d be ok with giving alpine a shot. I’ve only ever had some trouble in the past running precompiled binaries in alpine if they were compiled with glibc.

Another option which might still give the flexibility of Debian but reduce file size would be to change `latest` to `stable-slim`, as that base looks to only be 22MB. We can also update the COPY lines to only copy the needed files (the binary, templates, static folder, version, and migrations... I think) instead of the whole directory.

But those are optimizations that we can do once things work :) still great input @Reedtechno!

jordan-wright:
After poking around on this for a while, I think for now let's just move `debian:latest` to `debian:stable-slim`. That brings the container size from 250-ish MB close to 200-ish. Still not _great_, but better.

I tried various things such as moving the config generation to a separate build step, trying to move away from `debian` (which is likely still possible, but will be difficult trying to move to something like `scratch` or `alpine`).

For now, I'm ok just getting _something_ out the door and we can optimize further later. 😄

wjwoodson:
This is probably mostly my inexperience with packaging golang projects as well as with docker generally, but I wasn't able to get build working with an `alpine`-based image (or build with debian/gcc & run in `alpine`).

I've updated the PR to use `debian:stable-slim` for the final image.

jordan-wright:
LGTM - thanks @wjwoodson!

arturslogins:
Hello, why you don' define all config parameters in env ?  Like DB_PATH and so on

Issue #1428, opened 2019-04-14
JS devDependencies outdated


cschwartz:
**Thanks for reaching out! We're happy to help resolve issues as quickly as possible.**

**Please use this template when creating a new issue. If you do not follow this template format, your issue may be closed without comment.**

What version of Gophish are you using?: Master branch

Brief description of the issue: 
First of all, thanks for all the continued great work with GoPhish.

I was just setting up my environment to look if I could contribute, when I realized that the devDependencies used to build GoPhishs frontend are outdated and include deprecated packages (see below for the npm output).

Unless there is a reason why you want to keep the current version of packages around, I could provide a pull request to update all used packages to their current version (note, that the current version of Gulp changed the syntax to define tasks, so this would probably include some changes in the `gulpfile.js`).

What are you expecting to see happen? :
No deprecated packages are being used.

What are you seeing happen?
Deprecated packages are being used.

Please provide any terminal output that may be relevant below:

```
npm WARN deprecated gulp-util@3.0.8: gulp-util is deprecated - replace it, following the guidelines at https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5
npm WARN deprecated wreck@6.3.0: This version has been deprecated in accordance with the hapi support policy (hapi.im/support). Please upgrade to the latest version to get the best features, bug fixes, and security patches. If you are unable to upgrade at this time, paid support is available for older versions (hapi.im/commercial).
npm WARN deprecated boom@2.10.1: This version has been deprecated in accordance with the hapi support policy (hapi.im/support). Please upgrade to the latest version to get the best features, bug fixes, and security patches. If you are unable to upgrade at this time, paid support is available for older versions (hapi.im/commercial).
npm WARN deprecated hoek@2.16.3: This version has been deprecated in accordance with the hapi support policy (hapi.im/support). Please upgrade to the latest version to get the best features, bug fixes, and security patches. If you are unable to upgrade at this time, paid support is available for older versions (hapi.im/commercial).
npm WARN deprecated graceful-fs@3.0.11: please upgrade to graceful-fs 4 for compatibility with current and future versions of Node.js
npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated natives@1.1.6: This module relies on Node.js's internals and will break at some point. Do not use it, and update to graceful-fs@4.x.
[...]

PR #1438, opened 2019-04-21
updated devDependencies, migrated gulpfile.js to gulp 4.0 syntax

cschwartz:
Hi,

as promised in issue #1428, this is the pull request to update the js devDependencies.
I've updated the packages to their most current version as per npmjs.org.
I've also removed `gulp-util`, which is deprecated (see https://www.npmjs.com/package/gulp-util) as no other dependencies exist anymore.

The `gulpfile.js` now uses the current gulp 4.0 syntax (tasks are now exposed as exports rather than by calling `gulp.task`).
The `build` task now calls the `vendorjs`, `scripts` and `styles` sub-tasks in parallel, as I did not see any interdependencies in the tasks.

As requested, this pull request does not apply the new dependencies to the static files. This has to be done in an extra commit.

If you require any additional modifications, just let me know.

Best,
  Christian


jordan-wright:
Excellent, thanks @cschwartz! I'll merge this into a branch, build the dependencies, and push them back up to `master`.

Issue #1417, opened 2019-04-02
really needs Message-ID header


jhaar:
What version of Gophish are you using?: 

current binary for Linux (0.7.1).

Brief description of the issue: 

Being caught as spam: gophish doesn't add a Message-ID header - which is a pseudo-requirement according to RFC5322. Most antispam solutions will mark down email without it - and Google even mentions it WRT delivering email to Gmail

https://support.google.com/a/answer/81126

What are you expecting to see happen? :

Have a Message-ID header

What are you seeing happen?

Have a Message-ID header

I actually tried faking one using the "Additional headers" feature, but I of course each email will needs it's own unique Message-ID header, and my attempts at "<{{.Rid}}{{.FirstName}}{{.LastName}}@example.com>" created a Message-ID of "<" (ie one char) - so that went wrong

It would be better to do a good generalized header that looks something like "Message-ID: <523B6001.3010003@example.com>"

jordan-wright:
Hi @jhaar,

This should be no problem to add. I guess the gomail library doesn't add a message ID (I thought it did).

Thanks for the heads up!

jhaar:
Actually, I found a workaround. I added a header, ie

Message-ID with value "<randomstring-{{.RId}}@our.domain>" 

(don't forget the angle brackets).

That is enough to emulate a legitimate and unique-per-msg Message-ID, with the additional advantage I can make "our.domain" match or not match what I'm using in the From: field (which can add favour to testing). 

Hint for beginners like me: it's "{{.RId}}" - not ".{{Rid}}" (uppercase I) - I had tried this unsuccessfully earlier. :-/

justaskingquestions:
@jhaar  Thanks for the hint with the message id! My e-mails to G-Mail started bouncing compared to last year. This fixed it.

aczietlow:
I took a crack at adding the message-id header by default.

jordan-wright:
Closed via #1441

justaskingquestions:
I know this is fixed now, but AFAIK not in any release build yet. 
I have found an additional solution for those who are not running the dev build and are using postifx as a mail server Adding the following line to your main.cf and restarting postfix will add a vaild Message-ID header matching the hostname of your mailserver.
```
#Interim fix to add message-id header. This should be obsolete with gophish 0.8
always_add_missing_headers = yes
```

PR #1441, opened 2019-04-22
Update Sending Profile Message-ID headers (#1417)

aczietlow:
* Adds a default message-ID header to any sending profile on create or update when one does not already exist.

### DIFF
--- a/models/maillog.go
+++ b/models/maillog.go
@@ -1,12 +1,15 @@
 package models
 
 import (
+	"crypto/rand"
 	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
 	"math"
+	"math/big"
 	"net/mail"
+	"os"
 	"strings"
 	"time"
 
@@ -162,6 +165,14 @@ func (m *MailLog) Generate(msg *gomail.Message) error {
 	if conf.ContactAddress != "" {
 		msg.SetHeader("X-Gophish-Contact", conf.ContactAddress)
 	}
+
+	// Add Message-Id header as described in RFC 2822.
+	messageID, err := m.generateMessageID()
+	if err != nil {
+		return err
+	}
+	msg.SetHeader("Message-Id", messageID)
+
 	// Parse the customHeader templates
 	for _, header := range c.SMTP.Headers {
 		key, err := ExecuteTemplate(header.Key, ptx)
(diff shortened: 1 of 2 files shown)


jordan-wright:
Hi @aczietlow,

Thanks for taking a first stab at this! I appreciate you getting the ball rolling.

I can 100% see where you're going with this PR, but I think we can make this a bit simpler. Right now, in `models/maillog.go`, we generate the message in the `Generate` function. The way this function works for headers is that it sets defaults (e.g. [here's](https://github.com/gophish/gophish/blob/master/models/maillog.go#L160) how we set the transparency headers) and then it [enumerates through the saved headers](https://github.com/gophish/gophish/blob/master/models/maillog.go#L165) and sets them. So, we could set a default `Message-Id` header, and then if the user sets a custom one, it'll be overwritten.

Something like this:

```go
messageId, err := m.generateMessageId()
if err != nil {
    return err
}
msg.SetHeader("Message-Id", messageId)
// Parse the customHeader templates
for _, header := range c.SMTP.Headers {
....
```

For implementation, I would just move the `generateMessageId` to the Maillog struct, so it'd be something like:

```go
func (m *MailLog) generateMessageId() (string, error) {
    // Implementation here
}
```

In terms of implementation for this approach, I don't think what's here is bad, maybe just a bit complicated for what it's doing. [Another approach](https://github.com/jordan-wright/email/blob/3ea4d25e7cf84fe656d10353672a0eba4b919b6d/email.go#L585) is in a separate email library I maintain, which implements the description given in RFC 2822:

> Since the msg-id has a similar syntax to angle-addr (identical except that comments and folding white space are not allowed), a good method is to put the domain name (or a domain literal IP address) of the host on which the message identifier was created on the right hand side of the "@", and put a combination of the current absolute date and time along with some other currently unique (perhaps sequential) identifier available on the system (for example, a process id number) on the left hand side.  
> 
> Using a date on the left hand side and a domain name or domain literal on the right hand side makes it possible to guarantee uniqueness since no two hosts use the same domain name or IP address at the same time.  Though other algorithms will work, it is RECOMMENDED that the right hand side contain some domain identifier (either of the host itself or otherwise) such that the generator of the message identifier can guarantee the uniqueness of the left hand side within the scope of that domain.

I'm personally fond of that approach, but would love to hear your feedback.

All this said, more than anything I really appreciate you taking the time to make this happen! Looking forward to hearing your thoughts!

aczietlow:
You're absolutely right, adding this to the mailLog struct is a much better approach. I moved my work from the Smtp model to the mailLog.

After looking over it, I also greatly prefer your previously implemented approach to generating the Message-ID number. I took another pass based on your feedback from above.

I appreciate you taking the time to give me feedback.

jordan-wright:
Thanks for the quick follow-up! This looks good to me.

Thanks again for taking the time to knock this out. Great work!

PR #1448, opened 2019-05-09
Prevent adding an empty recipient

dmaciejak:
If the CSV is malformed, parsing it will still add an empty entry to the user table. That patch is to prevent it.


jordan-wright:
Hi @dmaciejak,

This looks good to me! Thanks for sending this in.

Issue #1456, opened 2019-05-16
Support for encrypted connections to a mysql db


cschwartz:
**Thanks for reaching out! We're happy to help resolve issues as quickly as possible.**

**Please use this template when creating a new issue. If you do not follow this template format, your issue may be closed without comment.**

Before filing a new issue, please use the search bar at the top of the browser to search for similar issues. Also, please make sure you have read our [documentation](https://getgophish.com/documentation/), which covers how to use Gophish.

What version of Gophish are you using?: 
master

Brief description of the issue: 
Currently no configurations exist to specify the ssl ca (i.e. PEM file) for connections to a mysql db. A configuration via the data source name (`db_path`) string is not possible.

Introducing a new config string which registers the referenced PEM file via `mysql.RegisterTLSConfig` is not terribly complex, and if you're interested I'd love to provide a pull request.

What are you expecting to see happen? :
The config file contains an option which would allow me to specify the path to a PEM file, which in turn would allow me to use the `tls=<custom>` option in the data source name string.

What are you seeing happen?
No such configuration option exist, preventing the use of encrypted mysql connections where the use of a ssl ca is required (i.e. AWS RDS).

If this question is related to email templates or landing pages not working as expected, please provide your template or landing page below:

```html
Insert code here
```

Please provide any terminal output that may be relevant below:

```
Insert output here
```

Please provide as many steps as you can to reproduce the problem:
 * Step 1 - 
 * Step 2 -

jordan-wright:
Hi @cschwartz,

Thanks for reaching out! This sounds great. If you want to take a first pass at it and send up a pull request, I'll help you get things merged in.

Thanks for offering to get this implemented!

jordan-wright:
Closed via 26d99b5a653da4fc5c59d04b262d59aab35c32c9. Thanks again for sending this in!

PR #1460, opened 2019-05-20
added support for encrypted connections to mysql

cschwartz:
This pull request provides a first pass on a new config option `db_sslca_path` in `config.json` in order to support encrypted connections to a mysql db (see #1456).

If the `db_sslca_path` is present in the configuration and `db_path` includes the query parameters `tls=ssl_ca`, the connection to the mysql db is encrypted using the given public key (which would be provided by the administrator of the db).

If the query string `tls=ssl_ca` is added, even though no `db_sslca_path` is set, the database connection fails in models/models.go#129  (or 102, pre pull request) and the error is presented to the user after `MaxDatabaseConnectionAttempts` attempts.
Specifying a non-existing, unreadable or invalid PEM file would also result in errors in models/models.go#110 and 114.
If the `db_sslca_path` is present but the `db_path` is not adapted, the connection is attempted without using encryption (and may fail, depending on db configuration).

Currently, no attempt is made to check if the `tls` query parameter is set correctly if a `db_sslca_path` is present. This could be done (either by parsing the DSN manually - which I would like to prevent - or by relying on `go-sql-driver/mysql#ParseDSN`).
If we check for the tls parameter, two consequences would be possible:

- Display an error and abort, as the connection would be unencrypted if we continued with a missing `tls` parameter
- Display a warning and continue, as the user might be debugging the config and knowingly removed the `tls` parameter and kept the reference to the CA for later use.

Alternatively, we could of course change nothing on this specific behaviour :)

If you would like this, or any other changes, please let me know.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -32,6 +32,7 @@ type Config struct {
 	PhishConf      PhishServer   `json:"phish_server"`
 	DBName         string        `json:"db_name"`
 	DBPath         string        `json:"db_path"`
+	DBSSLCaPath    string        `json:"db_sslca_path"`
 	MigrationsPath string        `json:"migrations_prefix"`
 	TestFlag       bool          `json:"test_flag"`
 	ContactAddress string        `json:"contact_address"`
--- a/models/models.go
+++ b/models/models.go
@@ -5,10 +5,13 @@ import (
 	"fmt"
 	"io"
 	"time"
+	"crypto/tls"
+	"crypto/x509"
+	"io/ioutil"
 
 	"bitbucket.org/liamstask/goose/lib/goose"
 
-	_ "github.com/go-sql-driver/mysql" // Blank import needed to import mysql
+	mysql "github.com/go-sql-driver/mysql"
 	"github.com/gophish/gophish/config"
 	log "github.com/gophish/gophish/logger"
 	"github.com/jinzhu/gorm"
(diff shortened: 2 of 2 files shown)


jordan-wright:
Hey @cschwartz,

Thanks for sending this in! I'll give this a shot here pretty soon and, if things work as they appear they should, I'll get this merged in.

jordan-wright:
Hey @cschwartz,

Can you walk me through a little bit more about the use case here? I'm not terribly familiar with MySQL's TLS support, but every example seems to suggest that clients would want to specify the client keypair as well as the CA root certs presumably so that the client can ensure the server's cert was signed by a valid CA and vice-versa.

In this PR, I only see support for specifying the CA certificate, not the client keypair. Do you think that there would be value in specifying the client details per [this example](https://godoc.org/github.com/go-sql-driver/mysql#RegisterTLSConfig)?

That said, the implementation here looks great - I'm just trying to understand the flow a little better on my end to make sure we're not missing anything 😄

cschwartz:
Sure, I'll try 😄 
I'm currently thinking about what would be necessary to deploy gophish in AWS, especially using RDS (their Database as a Service solution).
If I were to use RDS, I would want to ensure encryption in transit. To this end, AWS provides a root certificate (see https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html) signed by the AWS CA.
In order to trust the RDS root certificate, one would either have to add it to the system wide certificate store, or directly provide the PEM to the application (in this case gophish, though the mysql cli provides a similar mechanism using the `--ssl-ca` parameter).

From my understanding, the SSL client certificates provide additional means to authenticate the client to the server.
There probably is use in supporting client certificates (though my particular use case didn't require it), the reason I did not include client certificate support in this pull request is, to be honest, quite simple: I didn't have client certificates handy to test them 😃 

How do you suggest we proceed?
- We could either wait until the need in the community for client certificates arises, then an implementation should be pretty much straight forward, or
- I could find out how to generate me some client certificates and we would increase the scope of this pull request.

Best,
  Christian

jordan-wright:
Thanks for the explanation @cschwartz!

I think what I was missing was that there are solutions that request validation with their provided CA cert without also requiring client certificates. To that end, I think waiting for client certs to become a need is reasonable instead of working to develop everything upfront. Using the path you laid out here, I think it'd be an easy addition if that time ever comes. 😄

PR #1517, opened 2019-07-22
Added response headers for CORS and server identification

glennzw:
Added _"Access-Control-Allow-Origin": "*"_ to the **ReportHandler** to allow reporting extensions to bypass CORS restrictions.

Added _"X-Server": "GoPhish"_ to the **PhishHandler** endpoint to assist in identifying a server as a GoPhish server.

### DIFF
--- a/controllers/phish.go
+++ b/controllers/phish.go
@@ -196,7 +196,7 @@ func (ps *PhishingServer) ReportHandler(w http.ResponseWriter, r *http.Request) 
 // (such as clicked link, etc.)
 func (ps *PhishingServer) PhishHandler(w http.ResponseWriter, r *http.Request) {
 	r, err := setupContext(r)
-	w.Header().Set("X-Server", "GoPhish") // Useful for checking if this is a GoPhish server (e.g. for campaign reporting plugins)
+	w.Header().Set("X-Server", config.ServerName) // Useful for checking if this is a GoPhish server (e.g. for campaign reporting plugins)
 	if err != nil {
 		// Log the error if it wasn't something we can safely ignore
 		if err != ErrInvalidRequest && err != ErrCampaignComplete {
--- a/controllers/phish.go
+++ b/controllers/phish.go
@@ -196,8 +196,8 @@ func (ps *PhishingServer) ReportHandler(w http.ResponseWriter, r *http.Request) 
 // (such as clicked link, etc.)
 func (ps *PhishingServer) PhishHandler(w http.ResponseWriter, r *http.Request) {
 	r, err := setupContext(r)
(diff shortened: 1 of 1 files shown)


controllers/phish.go
```diff
@@ -195,6 +196,7 @@ func (ps *PhishingServer) ReportHandler(w http.ResponseWriter, r *http.Request)
 // (such as clicked link, etc.)
 func (ps *PhishingServer) PhishHandler(w http.ResponseWriter, r *http.Request) {
 	r, err := setupContext(r)
+	w.Header().Set("X-Server", "GoPhish") // Useful for checking if this is a GoPhish server (e.g. for campaign reporting plugins)
```
jordan-wright: Let's move this below the error check. The reason is that I only want to reveal that this is a server running Gophish if a valid `rid` parameter is provided.

Additionally, I have a constant called `ServerName` in `config.go` that's set to `gophish`. We can probably reuse this to keep things consistent, since this is what's set in the `X-Mailer` header when sending emails.

```suggestion
	w.Header().Set("X-Server", config.ServerName) // Useful for checking if this is a GoPhish server (e.g. for campaign reporting plugins)
```

jordan-wright (changes requested):
Hey @glennzw,

Thanks so much for taking the time to send this in! I've dropped a comment on the `X-Server` addition that will hopefully be a quick change. Once that's knocked out, I'll get this merged in.

Thanks again!

glennzw:
Makes sense, thanks!

controllers/phish.go
```diff
@@ -195,6 +196,7 @@ func (ps *PhishingServer) ReportHandler(w http.ResponseWriter, r *http.Request)
 // (such as clicked link, etc.)
 func (ps *PhishingServer) PhishHandler(w http.ResponseWriter, r *http.Request) {
 	r, err := setupContext(r)
+	w.Header().Set("X-Server", config.ServerName) // Useful for checking if this is a GoPhish server (e.g. for campaign reporting plugins)
```
jordan-wright: Hey @glennzw,

Thanks for the quick follow-up! The last thing to knock out would be to just move this line below the `if err != nil` error check below. Then we should be good to merge! 😄

jordan-wright (changes requested):
Hey @glennzw,

Thanks for the quick follow-up! The last thing to knock out would be to just move this line below the if err != nil error check below. Then we should be good to merge! 😄

jordan-wright:
Hey @glennzw,

Thanks for the continued work on this! I'm going to go ahead and merge this into a branch and knock out a final review. Then I'll get things merged into master 😄 

Thanks again for taking the time to send this in! Great work!

PR #1519, opened 2019-07-22
Added feature: Descriptions

quelsan:
This commit contains changes to allow campaigns, mail templates and
landing pages to have a description.

My primary usecase for this feature is to extract the data and include it during
report generation.

Changes affect DB migration scripts, models and static front-end
files.


quelsan:
Note that the client-side code needs to be re-minified, as new properties have been added

Issue #884, opened 2017-12-12
Adding Pictures within Email Template


kjohnson5:
**Thanks for reaching out! We're happy to help resolve issues as quickly as possible.**

**Please use this template when creating a new issue. If you do not follow this template format, your issue may be closed without comment.**

Before filing a new issue, please use the search bar at the top of the browser to search for similar issues. Also, please make sure you have read our [documentation](https://getgophish.com/documentation/), which covers how to use Gophish.

What version of Gophish are you using?:  v.4.0

Brief description of the issue: Some images do not appear when the HTML source button is selected. Therefore, I would like to embed these images myself within the email. However, I've added the image to the server and point to the location within html using the <img src> tag but that does not work. I've tried to add the image to the bottom of the email template but then I do not know how to insert it within the email I'm attempting to generate. I believe the issue lies with the html code [cid:image001.png@01D36858.2E71D860] but I've tried changing this to <img src> pointing to the image location on the server and that doesn't work as well as adding the image to the template but I'm not sure how to point to that added image or how to insert it where I would like to be. How can I add additional images to the template without using another image hosting service?

What are you expecting to see happen? : The actual image when the source button is press.

What are you seeing happen? Nothing is appearing when I attempt to insert an image using the html code. 

If this question is related to email templates or landing pages not working as expected, please provide your template or landing page below:

```html
Delivered-To: therecipient@gmail.com
Received: by 10.223.148.165 with SMTP id 34csp3591764wrr;
        Tue, 28 Nov 2017 12:07:23 -0800 (PST)
X-Google-Smtp-Source: AGs4zMZqZ1IBMhgGXN/79dVm+t/3+0ZRhkxTi4E4n2gaqCnFVyYlGOLRS77wjmRD6Db2394eq1J+
X-Received: by 10.98.242.67 with SMTP id y3mr363521pfl.82.1511899643628;
        Tue, 28 Nov 2017 12:07:23 -0800 (PST)
ARC-Seal: i=1; a=rsa-sha256; t=1511899643; cv=none;
        d=google.com; s=arc-20160816;
        b=ug/pG+hHEBopGBHNTECWNwtYz3dhYkY0qe32V4fu6D7GBuf3Uq+MM7zJuWI6tQKWNy
         VCGc/iwDsgYl+5btrBJN0/gKmCIYSKXqePmeY1qat/bPkyhG8Fn0LaWp754hJqXzdZiU
[...]

PR #1525, opened 2019-07-26
Embed or attach files based on their file extension

dzsibi:
There have been a few discussions about embedding images in e-mails (#251, #884). As it is right now, while Content-IDs were added, the Content-Disposition header was still set to attachment for all files, and as such, e-mail clients showed them as attached files. My proposed solution makes the judgement on whether a file should be embedded or attached based on its extension - jpg, jpeg, png and gif files are embedded (and have a Content-ID set), while other file types are attached. 

It might be better on the long run to offer a checkbox to set whether a file should be attached or embedded, but I believe that 99% of the time images are only added to a message to be shown inline. With these changes, adding images to a message will work as expected, while other files can still be added as attachments.

### DIFF
--- a/models/email_request.go
+++ b/models/email_request.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"io"
 	"net/mail"
+	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/gophish/gomail"
@@ -173,14 +175,18 @@ func (s *EmailRequest) Generate(msg *gomail.Message) error {
 	}
 	// Attach the files
 	for _, a := range s.Template.Attachments {
-		msg.Attach(func(a Attachment) (string, gomail.FileSetting, gomail.FileSetting) {
-			h := map[string][]string{"Content-ID": {fmt.Sprintf("<%s>", a.Name)}}
-			return a.Name, gomail.SetCopyFunc(func(w io.Writer) error {
-				decoder := base64.NewDecoder(base64.StdEncoding, strings.NewReader(a.Content))
+		copyFunc := gomail.SetCopyFunc(func(c Attachment) func(w io.Writer) error {
+			return func(w io.Writer) error {
+				decoder := base64.NewDecoder(base64.StdEncoding, strings.NewReader(c.Content))
 				_, err = io.Copy(w, decoder)
 				return err
-			}), gomail.SetHeader(h)
+			}
 		}(a))
+		if sort.SearchStrings(EmbeddedFileExtensions, filepath.Ext(a.Name)) < len(EmbeddedFileExtensions) {
+			msg.Embed(a.Name, copyFunc)
+		} else {
+			msg.Attach(a.Name, copyFunc)
+		}
 	}
 
 	return nil
--- a/models/email_request.go
+++ b/models/email_request.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"io"
 	"net/mail"
-	"path/filepath"
-	"sort"
 	"strings"
 
 	"github.com/gophish/gomail"
@@ -182,7 +180,7 @@ func (s *EmailRequest) Generate(msg *gomail.Message) error {
 				return err
 			}
 		}(a))
-		if sort.SearchStrings(EmbeddedFileExtensions, filepath.Ext(a.Name)) < len(EmbeddedFileExtensions) {
+		if shouldEmbedAttachment(a.Name) {
 			msg.Embed(a.Name, copyFunc)
 		} else {
 			msg.Attach(a.Name, copyFunc)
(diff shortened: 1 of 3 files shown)


Stuw:
Tested - patch works well.

I've applied it on top of master and tested email template with docx and jpg.
Image was embedded as expected, document was attached.

Thanks for the patch!

jordan-wright:
Hey there!

Thanks for sending this in- sorry for the delay in getting back to you. Can you expand a bit on this workflow? How are you adding images in the editor?

I’m excited for this since it definitely is a pain point for folks. Thanks for taking this on!

dzsibi:
Hey!

So the way this works is that once you add an image with the correct extension, you can then reference it in image tags or CSS using its name. You have to use the "cid:" URL format, like this:

``

Now this can only be done using the source code editor (I don't really use the WYSIWYG editor), but I guess in theory one could create data URLs from the uploaded images and use them to stand in place for the CID URLs when displaying the images in the editor.

dzsibi:
Hey, 

So I saw the discussion in #1865, and I would like to clarify a few points:

- Referencing images by CID will work without this patch
- The images will still show up as attachments without this patch

The reason for this is that when Gophish [uses](https://github.com/gophish/gophish/blob/ec8b17238ee725e84f0d2cff4c20c2b8bbb362f5/models/email_request.go#L172) msg.Attach to add the image files, the Content-Disposition header will be set to attachment instead of inline, this can be traced to gomail's implementation, see [here](https://github.com/go-gomail/gomail/blob/81ebce5c23dfd25c6c67194b37d3dd3f338c98b1/writeto.go#L131).

glennzw:
@dzsibi Yeah I use your patch in production, it works great, thank you.

nerotix:
I really this change, how would I go to try and get this running in my local installation? I tried building from source and make your changes inside the files and build it but that destroys the whole email templates section within the webgui.

fumamatar:
Are there any reasons why this is not merged? Inline attachments are vital to our use case and I am sure to many others too.

glennzw:
Hi @dzsibi. Thanks so much for your work on this, and your patience with us getting it merged in. I'm running some tests, and have noticed that all attachments get marked with `Content-Disposition: inline` (not just the image extensions in your `EmbeddedFileExtensions` slice. e.g here's a .docx attachment:

Screenshot 2021-12-22 at 18 22 08


I tested master vs your PR with two images and a docx, these are the results:

Running on master I get:
```
Content-Disposition: attachment; filename="cat.docx"
Content-Disposition: attachment; filename="cat.jpg"
Content-Disposition: attachment; filename="cat.png"
```

Running on your PR I get:
```
Content-Disposition: inline; filename="cat.docx"
Content-Disposition: inline; filename="cat.jpg"
Content-Disposition: inline; filename="cat.png"

```

glennzw:
The problem seems to be with this logic:

```go
		if sort.SearchStrings(EmbeddedFileExtensions, filepath.Ext(a.Name)) < len(EmbeddedFileExtensions) {
			msg.Embed(a.Name, copyFunc)
		} else {
			msg.Attach(a.Name, copyFunc)
		}
```
The _if statement_ always returns True, because if an extension is not in the list, an index position of zero is returned, which is less than `len(EmbeddedFileExtensions)`. The return value is the index to insert x if x is not present (so if not present it will return 0, because that's where you should insert it).

Also, `SearchStrings` [expects](https://pkg.go.dev/sort) a sorted slice, and your slice is not sorted so searching for, example, ".gif" returns an index of 0 (should return 4):

```go
var EmbeddedFileExtensions = []string{".jpg", ".jpeg", ".png", ".gif"}
```

So I don't think `sort.SearchStrings` is what we want to use. How about this:

```go
func contains(s []string, searchterm string) bool {
	i := sort.SearchStrings(s, searchterm)
	return i < len(s) && s[i] == searchterm
}
```

and then:

```go
                if contains(EmbeddedFileExtensions, filepath.Ext(a.Name)) {
			msg.Embed(a.Name, copyFunc)
		} else {
			msg.Attach(a.Name, copyFunc)
		}
```

This seems to work, you can test it here: https://go.dev/play/p/H7oSyscky1T

dzsibi:
Hey [glennzw](https://github.com/glennzw),

I rebased against the latest master and fixed the issue you mentioned. It was truly a rookie mistake, my apologies. Also added some tests just to make sure it works as expected. I also cleaned up the merge conflict with the templating support for attachments that was added after the original PR. 

(There are some formatting errors on the current master that are unrelated to my code. I fixed it on my own branch, but didn't want to polute the PR with these unrelated changes.)

glennzw:
Hi @dzsibi, thanks so much for following up with this, I think it's a great feature to add! No worries on the mistake, I'm quite the expert at making them. I'll take this for a spin now, really appreciate you adding the test conditions and rebasing.

Good spot on the formatting issues, I'd better get on top of that.

PR #1547, opened 2019-08-20
Fix multiple XSS issues in User Management page

dmaciejak:
If the user name is embedding some JS code, it will be executed on the client side. Note: gophish/static/js/dist/app/users.min.js will need to be regenerated too.

### DIFF
--- a/static/js/src/app/users.js
+++ b/static/js/src/app/users.js
@@ -19,7 +19,7 @@ const save = (id) => {
         user.id = id
         api.userId.put(user)
             .success(function (data) {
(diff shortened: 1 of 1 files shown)


jordan-wright:
LGTM - thanks! I don't know why I presumed that template literals would escape the variable contents. Great catch!

Just to make sure I'm not missing anything here, this seems like a self-stored XSS, in that only admins can create users, and only admins can view this user management page, right?

dmaciejak:
From my test I would say 'yes' only the user management page is affected. When authenticated as a malicious user name, the Account Settings page is not triggering the XSS.

PR #1556, opened 2019-08-24
Remove not working link for non admin role users

dmaciejak:
Users&Group link in the top menu to access the user management page should not be displayed for non admin role users.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -47,8 +47,10 @@
                     </li>
                     <li><a href="/campaigns">Campaigns</a>
                     </li>
+		    {{if .ModifySystem}}
                     <li><a href="/users">Users &amp; Groups</a>
                     </li>
+		    {{end}}
                     <li><a href="/templates">Email Templates</a>
                     </li>
                     <li><a href="/landing_pages">Landing Pages</a>
--- a/templates/base.html
+++ b/templates/base.html
@@ -43,22 +43,6 @@
             </div>
             <div class="navbar-collapse collapse">
                 <ul class="nav navbar-nav navbar-right">
-                    <li><a href="/">Dashboard</a>
-                    </li>
-                    <li><a href="/campaigns">Campaigns</a>
-                    </li>
-		    {{if .ModifySystem}}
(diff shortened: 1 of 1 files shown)


jordan-wright:
Ah, looks like that slipped past me. That should point to /groups now.

But honestly, I'm leaning towards removing all the navigation entries in the top nav. They're redundant from what's in the sidebar, so it'd probably be best to just keep everything in the sidebar in case that nav gets more complex over time (e.g. with nesting).

What do you think?

S0larflare:
I’d possibly make the top links just symbols?

Sent from my iPhone

On 24 Aug 2019, at 19:53, Jordan Wright <notifications@github.com<mailto:notifications@github.com>> wrote:


Ah, looks like that slipped past me. That should point to /groups now.

But honestly, I'm leaning towards removing all the navigation entries in the top nav. They're redundant from what's in the sidebar, so it'd probably be best to just keep everything in the sidebar in case that nav gets more complex over time (e.g. with nesting).

What do you think?

jordan-wright:
I’m a bit worried that might be even more confusing. However, you and I are on the same page that icons might be preferred for some users.

To that end, I’ve considered (among other dashboard improvements) having a button at the top left of the top navbar which would collapse the sidebar into just icons. This seems to be a common pattern in dashboards these days and I think it’s a nice touch. :)

dmaciejak:
Completely agree, the top nav entries are redundant and useless, it's just overloading the UI for me. A collapsing sidebar (with icons) is the way to go these days !

jordan-wright:
@dmaciejak I've thought about it a bit more, and I think we should just go ahead and remove the top nav.

Would you be willing to update this PR to do that? Otherwise, I can knock it out when I get a spare moment.

dmaciejak:
Will do !

jordan-wright:
LGTM- Thanks @dmaciejak!

rc4:
I agree the topnav is redundant now that you mention it, but it was what I always used before. If it didn't exist I suppose I would just use the side nav. Not a big deal either way :)

Issue #1262, opened 2018-11-07
Send Emails By does not exactly provide an even distribution


rubberducky69:
**Thanks for reaching out! We're happy to help resolve issues as quickly as possible.**

**Please use this template when creating a new issue. If you do not follow this template format, your issue may be closed without comment.**

Before filing a new issue, please use the search bar at the top of the browser to search for similar issues. Also, please make sure you have read our [documentation](https://getgophish.com/documentation/), which covers how to use Gophish.

What version of Gophish are you using?: 
**0.7.1**

Brief description of the issue: 
**When sending campaigns that are throttled (with a specified end-date), the emails are not actually distributed evenly between the two specified dates.**

**An example:**
_**I want to send 250 mails with a rate of 10 mails per minute (6 second interval). I specify a starting date: 12/7/2018 16:00 PM and an end date: 12/7/2018 16:25 PM. This results in 10 mails being scheduled to send every minute, and once the campaign is started, the 10 mails are sent with 1 second intervals.**_

What are you expecting to see happen? :
**When two dates are specified it takes the difference between the two dates, and divide it with 250, thus resulting in an interval of 6 seconds and an even distribution of emails.**

**I would love to see a change to this, or even an extra option with a MPM-rate (Mails Per Minute) that can be specified on campaign launch.**

jordan-wright:
Hi @rubberducky69,

Thanks for reaching out! You're correct, but I'm afraid this isn't something that we'll be able to fix.

The worker responsible for sending the emails polls the database for new emails once per minute. This is to balance frequent polls without taking the chance of overwhelming the database in the case of larger campaigns.

This means that when creating our send dates for emails, we round it to the nearest minute. Then, when new emails are picked up to be sent that particular minute, we send them immediately so we can both send emails quickly as well as reuse the existing SMTP connection.

At the very least, I'll update the documentation to be more clear that emails aren't evenly spaced down to the second, but rather to the minute. Sorry for any inconvenience this caused!

Issue #1176, opened 2018-08-23
Attachment Tracking & Open Rate


mirkatson:
Before filing a new issue, please use the search bar at the top of the browser to search for similar issues. Also, please make sure you have read our [documentation](https://getgophish.com/documentation/), which covers how to use Gophish.

What version of Gophish are you using?: 
0.6

Brief description of the issue: 
Tracking which user has downloaded the attachment (e.g word doc/pdf)

What are you expecting to see happen? :
Expecting to see which user has downloaded the attachment and opened it.

Or is it possible to include a individual tracking code (user's ID) in the document, so when the user opens it, it will generate a callback to the GoPhish server. From there, we are able to extract the callback URL in the log file and correlate it with the user's ID.
 
Thanks!

jordan-wright:
Hi @mirkatson,

Thanks for reaching out! As you've seen, we have issues #513 and #227 for tracking attachment support. We don't have an ETA at the moment, since we're actively working on other issues.

Since this appears to be a duplicate of those issues, I'm going to go ahead and close this one out so we can track everything through the existing issues.

Thanks again for getting in touch!

blshkv:
FYI, there is also https://github.com/onvio/gophish/commits/master. You might want to merge both if onvio fork has done something better

PR #1557, opened 2019-08-27
Added support to specify email sending frequency

glennzw:
Using the 'Send Emails By' option in the campaign allows for the sending of emails at set intervals, but requires some mental arithmetic. For example; 

A campaign has 1,000 emails, and I want them sent at 5 second intervals so as to not overload my mail server (or help desk). I want it to start now (say 11am). Therefore, 11h00 + (1000 * 5) = 5,000 seconds, so I'd need to set a 'Send Emails By' of 12h23.

This pull request helps by adding a new input field called 'Send Rate'. It's a drop down option underneath the 'Send Emails By' option consisting of a few common send rates:

- As fast as possible
- Every 5 seconds
- Every 30 seconds
- Every minute
- Every five minutes
- Add Custom

The 'As fast as possible' option  just clears the 'Send Emails By' calendar option, setting the form to its default state.

The final option in the dropdown is 'Add Custom', allowing you to set a custom send rate. When selecting a Send Rate the 'Send Emails By' calendar will be calculated, giving the estimated finish time.

So for our example above, selecting a start time of 11am for 1,000 targets, and selecting the Send Rate as 'Every 5 seconds' the calendar will be automatically updated to 12h23. 

Conversely, if the user selects a 'Send Emails By' option then the 'Send Rate' below it will be automatically calculated, so the user has a feel for how quickly emails will be dispatched.

Finally, as the user adds and removes target groups the calendar or send rate are updated to reflect the total number of targets. The last option edited between the calendar and the rate is the one that will stay fixed, with the other one changing to match the number of targets.

These are all front end changes and the send rate is approximate, as noted in [this discussion](https://github.com/gophish/gophish/issues/1262) (there's a burst of emails as the campaign starts). This allows us to have no backend changes for this pull request as we're only doing time calculations in Javascript and updating the 'Send Emails By'.

Here are some screenshots:

- [New Campaign Layout](https://user-images.githubusercontent.com/3966613/63766859-fec4d080-c8c3-11e9-8a77-fb320eea29b7.png)
- [Selecting a Rate](https://user-images.githubusercontent.com/3966613/63766877-084e3880-c8c4-11e9-8f23-cdb5886e87e5.png)
- [Rate Set](https://user-images.githubusercontent.com/3966613/63766875-07b5a200-c8c4-11e9-87a9-909e09381537.png)
- [Select Custom Rate](https://user-images.githubusercontent.com/3966613/63766874-07b5a200-c8c4-11e9-8410-ec5055d76dfd.png)
- [Custom Rate Set](https://user-images.githubusercontent.com/3966613/63766872-071d0b80-c8c4-11e9-8b1e-b1d83afc250a.png)

Other additions:

- Added a mouse over tooltip to the user/groups at the bottom to indicate how many targets are in said group. Example [here](https://user-images.githubusercontent.com/3966613/63765557-341bef00-c8c1-11e9-901e-b31d2514b2b4.png).

- Fixed a problem whereby selecting the 'Send Emails By' calendar in Chrome would result in a text autocomplete dropping down, making it impossible to select a date. Example of the problem 
[here](https://user-images.githubusercontent.com/3966613/63765316-a93af480-c8c0-11e9-9b60-75829ad0f599.png); the calendar is hidden behind the autocomplete.

This pull request is a PoC to gauge enthusiasm for this change, there are probably some bugs/inefficiencies but wanted to get a feel from maintainers before putting more work into it. I've just noticed there are also some left over debugging statements too.


jordan-wright:
Thanks for sending this in, @glennzw!

I look forward to testing this out. It might be a bit before I can give it a proper review since I have a ton on my plate atm, but just know that this is an exciting feature! I appreciate you handling all this in the front end, since that makes things significantly easier to review, as opposed to API changes 😄

sikelos:
> 
> 
> Thanks for sending this in, @glennzw!
> 
> I look forward to testing this out. It might be a bit before I can give it a proper review since I have a ton on my plate atm, but just know that this is an exciting feature! I appreciate you handling all this in the front end, since that makes things significantly easier to review, as opposed to API changes 😄

Hello Jordan,

Really impressed from this application so far. I am testing the last version and I missing the "Reporting emails" and the "tracking of attachments opening" (initiated in #1176 ).
My questions is: where can I get the software package with this included enhancements. Or is there another way to have these packages installed for testing?

Best regards

jordan-wright:
Hey @glennzw!

With an increase in folks asking about sending rates, I wanted to revisit this and get it merged. Would you mind removing the .min.js file?

Additionally, I'm wondering how best to communicate this to the user, but Gophish checks for new emails to send every minute. This means that if the user says they want emails to be sent every 5 seconds, then what really happens is that 12 emails will be queued and sent as fast as possible every minute.

Perhaps this might be better structured as "How many emails per minute" the user would like to send?

glennzw:
Hi @jordan-wright ! Ah great, be good to get this functionality rolled in. Hmm I did a `git rm static/js/dist/app/campaigns.min.js` and committed and pushed, not sure if that resolves the issue? I promise I'll stop committing .min files ones day 😄

So, I think conceptually it's easier for users to understand _'every n seconds'_ rather than _'n per minute'_, even if it's not 100% accurate as to what goes on under the hood (as they'd probably still end up doing the mental maths to get back to an approximate  per second rate)? Happy to change it if you prefer though! :)

static/js/src/app/campaigns.js
```diff
@@ -166,8 +166,10 @@ function setupOptions() {
             } else {
                 var group_s2 = $.map(groups, function (obj) {
                     obj.text = obj.name
+                    obj.title = obj.targets.length + " targets"
                     return obj
                 });
+                console.log(group_s2)
```
jordan-wright, line 172: Could we remove this log line?

glennzw: Done 👍🏻

jordan-wright:
Hey @glennzw!

Just a heads up, it looks like the .min.js file is still there. In this case, it looks like the `git rm` staged a change which deletes the entire file, which isn't what we want. You might need to `git checkout -- <file>` or something. Honestly, I haven't had much time to dig into that. If I come across any tips on how to get this resolved, I'll let you know.

jordan-wright:
Something like this will probably help https://stackoverflow.com/questions/39459467/remove-a-modified-file-from-pull-request

glennzw:
Gosh sorry about that, I've re-added the min.js file from the branch where I originally checked this pull request out from. Hopefully that works.

jsemoun:
Hey, thanks @glennzw for this awesome improvement. I wanted to know if this functionality of sending emails with randomized delays has been finally and added by @jordan-wright ? 
I'm currently using the 0.9.0. Would be so happy to try this new functionality!

glennzw:
Hi @jsemoun, thanks for the interest! Hoping to get this rolled into master soon.

k0z4c:
Any news about this pull request ? It would be so useful

FLX-0x00:
I think there is no progress anymore for external PRs. We have switched our server to a forked version

https://github.com/evait-security/gophish with some selected PRs merged. We have merged the code into our version from this PR. Workes as expected

PR #1579, opened 2019-09-09
Improved Dockerfile and run script

quelsan:
This pull request contains various changes to the Dockerfile and run script. Feel free to cherry pick changes as you deem appropriate.

## TL;DR:
- Allow local modifications of Gophish source code without rewriting import statements
- Include minimization process for client side assets (JavaScript)
- Run container as unprivileged user
- Add option to specify database path

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
 # Minify client side assets (JavaScript)
 FROM node:latest AS build-js
 
+RUN npm install gulp gulp-cli -g
+
 WORKDIR /build
 COPY . .
-
-RUN npm install gulp gulp-cli -g
 RUN npm install --only=dev
 RUN gulp
 
@@ -14,28 +14,29 @@ FROM golang:1.11 AS build-golang
 
 WORKDIR /go/src/github.com/gophish/gophish
 COPY . .
-
 RUN go get -v && go build -v
 
 
 # Runtime container
 FROM debian:stable-slim
 
-WORKDIR /opt/gophish
-RUN useradd -d /opt/gophish -s /bin/bash app
+RUN useradd -m -d /opt/gophish -s /bin/bash app
 
 RUN apt-get update && \
 	apt-get install --no-install-recommends -y jq && \
 	apt-get clean && \
 	rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+WORKDIR /opt/gophish
 COPY --from=build-golang /go/src/github.com/gophish/gophish/ ./
 COPY --from=build-js /build/static/js/dist/ ./static/js/dist/
 COPY --from=build-js /build/static/css/dist/ ./static/css/dist/
-RUN touch config.json.tmp && chown app. config.json config.json.tmp
-RUN sed -i 's/127.0.0.1/0.0.0.0/g' config.json
 
 USER app
+COPY --from=build-golang /go/src/github.com/gophish/gophish/config.json ./
+RUN sed -i 's/127.0.0.1/0.0.0.0/g' config.json
+RUN touch config.json.tmp
+
 EXPOSE 3333 8080 8443
 
 CMD ["./docker/run.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,9 +31,10 @@ WORKDIR /opt/gophish
 COPY --from=build-golang /go/src/github.com/gophish/gophish/ ./
 COPY --from=build-js /build/static/js/dist/ ./static/js/dist/
 COPY --from=build-js /build/static/css/dist/ ./static/css/dist/
+COPY --from=build-golang /go/src/github.com/gophish/gophish/config.json ./
+RUN chown app. config.json
 
 USER app
-COPY --from=build-golang /go/src/github.com/gophish/gophish/config.json ./
 RUN sed -i 's/127.0.0.1/0.0.0.0/g' config.json
 RUN touch config.json.tmp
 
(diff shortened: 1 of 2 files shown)


jordan-wright:
Hey there!

Thanks for sending this in. I'm excited to get these changes merged. For a while now, I've wanted to improve the Dockerfile so I appreciate you taking this first pass.

> Allow local modifications of Gophish source code without rewriting import statements

Could you expand a bit on this? I'm not seeing where in the changes this has been improved.

docker/run.sh
```diff
@@ -5,61 +5,69 @@ if [ -n "${ADMIN_LISTEN_URL+set}" ] ; then
...
         '.admin_server.listen_url = $ADMIN_LISTEN_URL' config.json > config.json.tmp && \
-        mv config.json.tmp config.json
+        cat config.json.tmp > config.json
 fi
 if [ -n "${ADMIN_USE_TLS+set}" ] ; then
     jq -r \
         --argjson ADMIN_USE_TLS "${ADMIN_USE_TLS}" \
         '.admin_server.use_tls = $ADMIN_USE_TLS' config.json > config.json.tmp && \
-        mv config.json.tmp config.json
+        cat config.json.tmp > config.json
```
jordan-wright: Just curious, why did you take this approach?

quelsan:
> Hey there!
> 
> Thanks for sending this in. I'm excited to get these changes merged. For a while now, I've wanted to improve the Dockerfile so I appreciate you taking this first pass.
> 
> > Allow local modifications of Gophish source code without rewriting import statements
> 
> Could you expand a bit on this? I'm not seeing where in the changes this has been improved.

"WORKDIR /go/src/github.com/gophish/gophish" is the solution I came up with.
Before go tries to fetch dependencies, it checks if they are already located in the directory.
If the source is not located in this directory, go will download the imported modules from the upstream Github repository.

docker/run.sh
quelsan: It is a work around for situations where the configuration is writable by the user but not necessarily owned by it. In this specific case however, if things are run using the shipped Dockerfile, it should not matter.

Dockerfile
```diff
@@ -1,30 +1,41 @@
...
+FROM node:latest AS build-js
 
-# build Gophish binary
-WORKDIR /build/gophish
+WORKDIR /build
 COPY . .
-RUN go get -d -v ./...
-RUN go build
 
+RUN npm install gulp gulp-cli -g
```
jordan-wright: As a small note, you could move this above the `COPY` line to speed up builds.

Dockerfile
```diff
@@ -1,30 +1,41 @@
...
-# copy Gophish assets from the build image
-WORKDIR /gophish
-COPY --from=build /build/gophish/ /gophish/
-RUN chmod +x gophish
 
-# expose the admin port to the host
+# Runtime container
+FROM debian:stable-slim
+
+WORKDIR /opt/gophish
```
jordan-wright: When running this, I show that gophish is unable to open a SQLite database in /opt/gophish since the folder is owned by root.

It seems a fix for this would be to move this under the `useradd` command, as well as adding the `-m` flag to the `useradd` command in order to have the home directory created.

Dockerfile
```diff
@@ -1,30 +1,41 @@
...
+RUN useradd -d /opt/gophish -s /bin/bash app
+
+RUN apt-get update && \
+	apt-get install --no-install-recommends -y jq && \
+	apt-get clean && \
+	rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
```
jordan-wright: I would move the `USER app` line here, that way files that get copied into the directory are owned by the `app` user.

Dockerfile
```diff
@@ -1,30 +1,41 @@
...
+COPY --from=build-golang /go/src/github.com/gophish/gophish/ ./
+COPY --from=build-js /build/static/js/dist/ ./static/js/dist/
+COPY --from=build-js /build/static/css/dist/ ./static/css/dist/
+RUN touch config.json.tmp && chown app. config.json config.json.tmp
```
jordan-wright: Once the `USER app` line is moved up, you can remove the `&& chown...` line since it won't be needed anymore.

jordan-wright (changes requested):
Hey @quelsan,

I was having some issues getting this to work, so I've dropped some comments on how things can probably be tweaked and improved. Give them a shot and let me know what you think!

quelsan:
Thanks for the great feedback - I have used a volume mounted DB so I've not experienced the permission issue. Will look into the optimizations and update the branch.

quelsan:
I've now updated the branch, per your request.

The only difference is that I just moved the copy of the configuration file, as I see no need for the application to be able to modify itself.

jordan-wright (approved):
These look good to me! Thanks again for taking a pass at this @quelsan!

Reedtechno:
Looks like this change may have restricted the user down too much:
`time="2019-11-14T02:25:19Z" level=fatal msg="listen tcp 0.0.0.0:80: bind: permission denied"`

icouper-ng:
I have run into the issue @Reedtechno identified.

I haven't been able to find instructions on how to elevate the user privileges within the docker container to resolve this. Any suggestions?

jordan-wright:
I’ll have a fix for this out soon

icouper-ng:
Great to hear. Thanks Jordan!

ACPK:
@Reedtechno - I also ran into the same issue after installling the docker issue.

@ jordan-wright - Any updates?

PR #1612, opened 2019-10-07
Added IMAP support for checking reported emails

glennzw:
This is quite a big pull request, with front end and back end changes. It was a tonne of work to be honest 🙂 This pull request allows you to set an IMAP server in Gophish that will be continuously monitored for Gophish campaigns that have been forwarded to it by users. If a campaign email is found, it's reported.

The logic is as follows:

1. Add IMAP details under `Account Settings > Report Settings` (see screenshot below)
2. If a user receives a suspicious email they can forward it to the address associated with the IMAP account
3. The IMAP Monitor checks for new email every n seconds (60 by default) and scans email for the `rid=?` tag, and reports it if found

Here's a story illustrating the process:

IMAP

Here's a video of it in action:

[https://youtu.be/D3rAYvAthdI](https://youtu.be/D3rAYvAthdI
)

There are some advanced configuration options;

- Set the monitor to delete reported campaign emails after successfully reporting them (so as an admin any emails in your specified inbox will only be non-Gophish campaigns and should be investigated).
 - Restrict checking of emails to a specified domain (e.g. set your company domain, widgets.com, so the monitor only checks emails from your staff).

There's decent error checking / testing too:

goodlogin

badlogin

<hr>

**Some notes on implementation and configuration:**

In `config.json` we can set the frequency of IMAP checking:
```
        "imap_freq": 60
```

The IMAP structure is defined in `models\imap.go`:
```
// IMAP contains the attributes needed to handle logging into an IMAP server to check
// for reported emails
type IMAP struct {
        UserId            int64     `json:"-" gorm:"column:user_id"`
        Enabled           bool      `json:"enabled"`
        Host              string    `json:"host"`
        Port              uint16    `json:"port,string,omitempty"`
        Username          string    `json:"username"`
        Password          string    `json:"password"`
        TLS               bool      `json:"tls"`
        Folder            string    `json:"folder"`
        RestrictDomain    string    `json:"restrict_domain"`
        DeleteCampaign    bool      `json:"delete_campaign"`
        LastLogin         time.Time `json:"last_login,omitempty"`
        LastLoginFriendly string    `json:"last_login_friendly,omitonempty"`
        ModifiedDate      time.Time `json:"modified_date"`
}
```

The actual IMAP account monitoring is done by `controllers/imapmon.go`. A separate worker is launched for each IMAP account (ie each Gophish user could have a different IMAP server set). This ensures against one account/server denying resources to others. IMAP connectivity is done with [eazye](https://github.com/glennzw/eazye). It uses outdated libraries, but was really easy to get up and running. I should ideally switch to[ go-imap](https://github.com/emersion/go-imap) at some point, but it'll be a bit more work (IMAP is a tricksy hobbit). Also need to add backing off logic, and maybe proper worker code.

Sticking with Gophish convention the management of IMAP settings is handled via an API, as you can see in `controllers/api/imap.go`. The endpoints are:
```
/api/imap --> Set IMAP settings
/api/imap/test --> Test IMAP supplied settings
```

The endpoints are accessible via JavaScript calls (implemented in `static/js/src/app/gophish.js`) . e.g: calling `api.IMAP.get()`:

apicall

The IMAP monitor is launched in `gophish.go` along with the adminServer and phishServer:

```
	imapMonitor := controllers.NewImapMonitor(conf)

	go adminServer.Start()
	go phishServer.Start()
	go imapMonitor.Start()
```

There's probably a bit of work to be done on how I handled this.

Finally the `settings.js` front end code probably needs some tidying in terms of code reuse and naming conventions.

Overall I've done a lot of testing and it seems to work very well. Also, watch this space for my Chrome and Outlook email reporting plugins.


quelsan:
Really nice and useful feature - haven't looked at the code yet, but I have been thinking about a feature such as this.

S0larflare:
So first off, that looks really neat, I really like the entire workflow there to be honest. I'm not sure if you need a realtime count for last checked in the email settings, but thats just me :smile:. Other than that, this is pretty fantastic and works better than how I have my reporting set up.

On another note, I've been butting my head against IMAP in Go for ages now and have never managed to get it to work properly. Any chance you'd be willing to field a few questions I have about parsing .msg attachments?

controllers/imapmon.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+}
+
+// NewImapMonitor returns a new instanfe of the ImapMonitor
+//func NewImapMonitor(configP config.PhishServer, configA *config.Config) *ImapMonitor {
+func NewImapMonitor(config *config.Config) *ImapMonitor {
+	// Make sure database connection exists. Not sure why I have to do this here, but
+	//  otherwise db is <nil> when calling models.GetEnabledIMAPs()
+	err := models.Setup(config)
```
jordan-wright: Hmm.... yeah, you shouldn't need to do this. Let me think on this a bit once I get the initial review done.

controllers/imapmon.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return im
+}
+
+// Start launches the IMAP campaign monitor
+func (im *ImapMonitor) Start() error {
+	log.Info("Starting IMAP monitor")
+	ctx, cancel := context.WithCancel(context.Background()) //ctx is the derivedContext
+	im.cancel = cancel
+	go PeriodicIMAPMonitor(ctx, im)
```
jordan-wright: What do you think about moving `PeriodicIMAPMonitor` to something like `im.start(ctx)` or some other name? Right now, it's not clear what use case we'd have that would require calling the exported `PeriodicIMAPMonitor` directly from outside the package.

controllers/imapmon.go
```diff
@@ -0,0 +1,184 @@
...
+func checkForNewEmails(im models.IMAP, wg *sync.WaitGroup, reportURL string) {
+	defer wg.Done()
+	mailSettings := eazye.MailboxInfo{
+		Host:   im.Host,
+		TLS:    im.TLS,
+		User:   im.Username,
+		Pwd:    im.Password,
+		Folder: im.Folder}
+
+	//Make sure server ends with a trailing slash
```
jordan-wright: Small nits - Go typically asks that comments have a space after the slashes. I'm not sure if `gofmt` (or better `goimports`) does this for you but you'll want to make sure that you're using those, mainly because they're super useful 😄

controllers/imapmon.go
```diff
@@ -0,0 +1,184 @@
...
+				}
+			}
+
+			body := string(append(m.Text, m.HTML...)) //Empty body is being returned for emails forwarded from ProtonMail. Need to investigate more. Perhaps it's being attached, and eazye doesn't support attachments.
+			rid := goPhishRegex.FindString(body)
+
+			if rid != "" {
+				rid := strings.TrimSpace(rid)
+				reportURL := fmt.Sprintf("%sreport%s", reportURL, rid)
+				response, err := netClient.Get(reportURL)
```
jordan-wright: This is an interesting approach. On one hand, it seems odd that we're making (almost) an external HTTP request to Gophish for the reporting endpoint from within Gophish. Instead, it seems preferred that we would load the `models.Result` object and handle the email report like we do in the [report handler](https://github.com/gophish/gophish/blob/master/controllers/phish.go#L188).

But, on the other hand, the benefit this approach provides is that for logging it can appear as though someone hit the reporting endpoint. 🤷‍♂ 

Just curious to hear your thoughts here while I think a little more about the approach I'd prefer.

db/db_mysql/migrations/20160118194630_init.sql
```diff
@@ -12,6 +12,7 @@ CREATE TABLE IF NOT EXISTS `group_targets` (group_id bigint,target_id bigint );
 CREATE TABLE IF NOT EXISTS `events` (id integer primary key auto_increment,campaign_id bigint,email varchar(255),time datetime,message varchar(255) );
 CREATE TABLE IF NOT EXISTS `campaigns` (id integer primary key auto_increment,user_id bigint,name varchar(255) NOT NULL ,created_date datetime,completed_date datetime,template_id bigint,page_id bigint,status varchar(255),url varchar(255) );
 CREATE TABLE IF NOT EXISTS `attachments` (id integer primary key auto_increment,template_id bigint,content text,type varchar(255),name varchar(255) );
+CREATE TABLE IF NOT EXISTS `imap` (user_id bigint,host varchar(255),port int,username varchar(255),password varchar(255),modified_date datetime,tls boolean,enabled boolean,folder varchar(255,restrict_domain varchar(255),delete_campaign boolean,last_login datetime,last_login_friendly varchar(255)));
```
jordan-wright: We'll want to move all of these into a new database migration. I'll reach out through the Google group with a note on how to make those using `goose`.

models/imap.go
```diff
@@ -0,0 +1,177 @@
...
+	UserId            int64     `json:"-" gorm:"column:user_id"`
+	Enabled           bool      `json:"enabled"`
+	Host              string    `json:"host"`
+	Port              uint16    `json:"port,string,omitempty"`
+	Username          string    `json:"username"`
+	Password          string    `json:"password"`
+	TLS               bool      `json:"tls"`
+	Folder            string    `json:"folder"`
+	RestrictDomain    string    `json:"restrict_domain"`
+	DeleteCampaign    bool      `json:"delete_campaign"`
```
jordan-wright: I'd consider renaming this. `DeleteCampaign` makes me think that the entire campaign will be deleted somehow. 😄

models/imap.go
```diff
@@ -0,0 +1,177 @@
...
+	LastLogin         time.Time `json:"last_login,omitempty"`
+	LastLoginFriendly string    `json:"last_login_friendly,omitonempty"`
```
jordan-wright: I'd recommend we remove this, personally. Instead, we can let the frontend (specifically moment.js) do all the work. This saves us from introducing another dependency and makes the backend logic cleaner.

models/imap.go
```diff
@@ -0,0 +1,177 @@
...
+	ip := net.ParseIP(s.Host)
+	_, err := net.LookupHost(s.Host)
+	if ip == nil && err != nil {
+		return ErrInvalidIMAPHost
+	}
+
+	// Make sure 1 >= port <= 65535
+	if s.Port < 1 || s.Port > 65535 {
+		return ErrInvalidIMAPPort
+	}
```
jordan-wright: This is where I'd set the default polling interval to 60 seconds.

models/imap.go
```diff
@@ -0,0 +1,177 @@
...
+		return ErrIMAPPortNotSpecified
+	case s.Username == "":
+		return ErrIMAPUsernameNotSpecified
+	case s.Password == "":
+		return ErrIMAPPasswordNotSpecified
+	}
+
+	// Set the default value for Folder
+	if s.Folder == "" {
+		s.Folder = "INBOX"
```
jordan-wright: I'd recommend moving this to a constant. Something like:

```go
const DefaultIMAPFolder = "INBOX"
```

models/imap.go
```diff
@@ -0,0 +1,177 @@
...
+// DeleteIMAP deletes the existing IMAP in the database.
+func DeleteIMAP(uid int64) error {
+	err := db.Where("user_id=?", uid).Delete(&IMAP{}).Error
+	if err != nil {
+		log.Error(err)
+	}
+	return err
+}
+
+// TestIMAP tests supplied IMAP settings by connecting to the server
```
jordan-wright: I'd move this functionality to the `gophish/imap` package mentioned earlier. I've tried (probably inconsistently, in some places) to keep the `models` package to just the data storage layer.

static/js/dist/app/settings.min.js
```diff
@@ -1 +1,247 @@
-$(document).ready(function(){$("#apiResetForm").submit(function(e){return api.reset().success(function(e){user.api_key=e.data,successFlash(e.message),$("#api_key").val(user.api_key)}).error(function(e){errorFlash(e.message)}),!1}),$("#settingsForm").submit(function(e){return $.post("/settings",$(this).serialize()).done(function(e){successFlash(e.message)}).fail(function(e){errorFlash(e.responseJSON.message)}),!1});var e=localStorage.getItem("gophish.use_map");$("#use_map").prop("checked",JSON.parse(e)),$("#use_map").on("change",function(){localStorage.setItem("gophish.use_map",JSON.stringify(this.checked))})});
\ No newline at end of file
+$(document).ready(function () {
```
jordan-wright: As a heads up, it looks like this is editing the `.min.js` file instead of the `.js` file.

static/js/dist/app/settings.min.js
```diff
@@ -1 +1,247 @@
...
+            return false
+        }
+        if (isNaN(imapSettings.port) || imapSettings.port <1 || imapSettings.port > 65535  ){ 
+            errorFlash("Invalid IMAP Port")
+            document.body.scrollTop = 0;
+            document.documentElement.scrollTop = 0;
+            return false
+        }
+
+        //api.IMAP.post(imapSettings).done(function (data) { // When using this API approach we get an error in the logs "http: TLS handshake error from 127.0.0.1:53858: remote error: tls: unknown certificate"
```
jordan-wright: I'd be curious to look into this further, since that's definitely not expected.

jordan-wright (changes requested):
Hey @glennzw!

**Wow!** You've done a stellar job with this PR! I really appreciate you taking the time to send this in. Very exciting stuff.

I like the approach and the way you've architected things. I think across the board you're spot on.

I've left some comments from an initial pass. It's possible there's more to review later, but I wanted to start giving feedback ASAP. Let me know your thoughts and we can work through this to get it in great shape and get it merged.

Thanks again so, so much for taking the time to send this over! Looking forward to adding this to Gophish!

jordan-wright:
Just an update, here's how to create a new SQL migration. First, you need to install `goose` if you don't have it already:

```
go get bitbucket.org/liamstask/goose/cmd/goose
```

> Note: We currently use `liamstask/goose`, but eventually I'd like to see about moving to `github.com/pressly/goose` since it's far more maintained.

Then, in the `db/` folder, you can create a SQL migration for both SQLite and MySQL. The format is `version_description`, like this:

```
gophish/gophish/db$ goose -path db_sqlite3 -env production create 0.9.0_webhooks sql
goose: created /src/go/src/github.com/gophish/gophish/db/db_sqlite3/migrations/20191008201325_0.9.0_webhooks.sql
```

Then you can add the relevant SQL statements to the new migration file. These will execute when Gophish is run, since we compare the version in the database with the latest migration file. This means that you'll want to create a backup of your database when developing migrations so that you can easily roll back to a previous state.

glennzw:
Hi @jordan-wright, thanks so much for the quick response and positive feedback! I'll address each of your comments individually but they all seem to make sense. I'll work on getting everything in line to make a merge! I'm sure more issues will crop up as you look closer at the code too. I'll ping you if I get stuck with goose too, thanks!

@S0larflare; I feel ya, IMAP is a pain. I can't give more advice than the example code on the go-imap repo for parsing attachments. I used eazye in the end as it's a super simple wrapper (but doesn't support attachments; I actually spent about a day trying to add that support!).

controllers/api/imap.go
```diff
@@ -0,0 +1,72 @@
...
+	"net/http"
+	"time"
+
+	"github.com/dustin/go-humanize"
+	ctx "github.com/gophish/gophish/context"
+	log "github.com/gophish/gophish/logger"
+	"github.com/gophish/gophish/models"
+)
+
+// ImapServerTest handles requests for the /api/imapserver/test endpoint
```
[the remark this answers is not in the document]
glennzw: Agreed, validate is a better verb. I've updated instances of "Test IMAP" to "Validate IMAP" and pushed a new commit.

controllers/imapmon.go
```diff
@@ -0,0 +1,184 @@
...
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+	"sync"
+	"time"
+
+	log "github.com/gophish/gophish/logger"
+
+	"github.com/glennzw/eazye"
```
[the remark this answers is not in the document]
glennzw: Yeah totally agree, additional 3rd party packages make us all nervous! 

So to be clear, [github.com/glennzw/eazye](github.com/glennzw/eazye) is my fork of [https://github.com/jprobinson/eazye](https://github.com/jprobinson/eazye) with added functionality (so it's not really my package). I agree with your idea to create a `github.com/gophish/gophish/imap` package to avoid nasties creeping in.

I'd like to move away from `eazye` at some point (as it uses outdated/unmaintained packages) and use `imap-go` but this was the quickest way to get up and running.

controllers/imapmon.go
```diff
@@ -0,0 +1,184 @@
...
+var goPhishRegex = regexp.MustCompile("(\\?rid=[A-Za-z0-9]{7})")
+
+// ImapMonitor is a worker that monitors IMAP serverd for reported campaign emails
+type ImapMonitor struct {
+	cancel    func()
+	reportURL string
+	checkFreq int
+}
+
+// PeriodicIMAPMonitor will periodically check the database for
```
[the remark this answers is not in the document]
glennzw: Eagle eyes! 🦅👀

Updated.

controllers/imapmon.go
```diff
@@ -0,0 +1,184 @@
...
+				go checkForNewEmails(i, &wg, im.reportURL)
+			}
+			wg.Wait()
+			time.Sleep(time.Duration(im.checkFreq) * time.Second)
+		}
+	}
+
+}
+
+// NewImapMonitor returns a new instanfe of the ImapMonitor
```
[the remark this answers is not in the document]
glennzw: Fixed.

controllers/imapmon.go
glennzw: Done. There are a few more I found in other Go files too in case you want to change them:

```
grep --include \*.go "[^:]//\S" -ir *

auth/auth.go:	//If we've made it here, we should have a valid user stored in u
auth/auth.go:	//Let's check the password
controllers/route.go:		//Attempt to login
controllers/route.go:		//If we've logged in, save the session and redirect to the dashboard
controllers/api/page.go:	//POST: Create a new page and return it as JSON
controllers/api/util.go:		//default message body
controllers/api/template.go:	//POST: Create a new template and return it as JSON
controllers/api/smtp.go:	//POST: Create a new SMTP and return it as JSON
controllers/api/group.go:	//POST: Create a new group and return it as JSON
controllers/api/campaign.go:	//POST: Create a new campaign and return it as JSON
middleware/session.go:	[]byte(securecookie.GenerateRandomKey(64)), //Signing key
models/campaign_test.go:	//problem
models/models.go:			Hash:     "$2a$10$IYkPp0.QsM81lYYPrQx6W.U6oQGw7wMpozrKhKAHUBVL4mkm/EvAS", //gophish
```

controllers/imapmon.go
glennzw: I went through the same thought process. It seems neater (and faster) to update the database directly, but as you guessed I did it this way for debugging to the console.

I'm happy to update to rather call `HandleEmailReport()` if you'd prefer.

models/imap.go
glennzw: Oops, good point! How about `DeleteCampaignEmail` or `DeleteReportedCampaign` ?

models/imap.go
glennzw: Done

static/js/dist/app/settings.min.js
glennzw: Thanks, shot myself in the foot there and had to come back here and grab the minified file.

models/imap.go
glennzw: I initially tried to handle this purely in the frontend, but couldn't get moment.js to play nicely. It was a while ago now and I can't remember the specific issue but will dig into it again.

S0larflare:
Why is LastLoginFriendly “omitonempty” rather than “omitempty”?

Sent from my iPhone

On 11 Oct 2019, at 21:05, Glenn Wilkinson <notifications@github.com> wrote:

﻿

@glennzw commented on this pull request.

________________________________

In models/imap.go<https://github.com/gophish/gophish/pull/1612#discussion_r334155121>:

> +
+// IMAP contains the attributes needed to handle logging into an IMAP server to check
+// for reported emails
+type IMAP struct {
+       UserId            int64     `json:"-" gorm:"column:user_id"`
+       Enabled           bool      `json:"enabled"`
+       Host              string    `json:"host"`
+       Port              uint16    `json:"port,string,omitempty"`
+       Username          string    `json:"username"`
+       Password          string    `json:"password"`
+       TLS               bool      `json:"tls"`
+       Folder            string    `json:"folder"`
+       RestrictDomain    string    `json:"restrict_domain"`
+       DeleteCampaign    bool      `json:"delete_campaign"`
+       LastLogin         time.Time `json:"last_login,omitempty"`
+       LastLoginFriendly string    `json:"last_login_friendly,omitonempty"`


I initially tried to handle this purely in the frontend, but couldn't get moment.js to play nicely. It was a while ago now and I can't remember the specific issue but will dig into it again.

jordan-wright:
I think the `omitonempty` is probably just a typo. Probably a quick fix @glennzw.

db/db_mysql/migrations/20160118194630_init.sql
glennzw: Created new migration files with goose 👍

glennzw:
> I think the `omitonempty` is probably just a typo. Probably a quick fix @glennzw.

Yup, fixed!

controllers/imapmon.go
glennzw: Done. Removed `PeriodicIMAPMonitor` and added `im.start(ctx)`

config/config.go
```diff
@@ -37,6 +37,7 @@ type Config struct {
 	TestFlag       bool          `json:"test_flag"`
 	ContactAddress string        `json:"contact_address"`
 	Logging        LoggingConfig `json:"logging"`
+	IMAPFreq       int           `json:"imap_freq"`
```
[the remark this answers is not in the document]
glennzw: Done, see https://github.com/glennzw/gophish/commit/8cd7c076e57624fc9c0fb713ef745eb983f9dfd2

models/imap.go
glennzw: Done. See https://github.com/glennzw/gophish/commit/8cd7c076e57624fc9c0fb713ef745eb983f9dfd2#diff-8340115d226d70820df1676949ea4d43R94

```
// Make sure the polling frequency is between every 30 seconds and every year
// If not set it to the default
if s.IMAPFreq < 30 || s.IMAPFreq > 31540000 {
	s.IMAPFreq = DefaultIMAPFreq
}
```

glennzw:
Thanks for your patience on getting these fixes in, I'm sure you're all very familiar with squeezing in open source projects in evenings and weekends 🙂

I've modified the IMAP polling logic (https://github.com/glennzw/gophish/commit/8cd7c076e57624fc9c0fb713ef745eb983f9dfd2) so that each user has their own IMAP monitoring Go routine, for which they can set their own polling frequency via the web UI advanced settings. The backend logic is:

1. Start an IMAP monitoring Go routine for each user
2. For each of the above, check if IMAP settings are present and enabled. If so, check then sleep for the poll frequency (default 60s)
3. If a user is deleted, stop their Go routine.
4. If a new user is created, a new Go routine is launched to monitor their IMAP settings

Took a while to settle on this logic, but I think it makes the most sense. Open to hear your feedback.

jordan-wright:
Hey @glennzw,

This is looking good! It seems like there's still some outstanding review items (moving things to `github.com/gophish/gophish/imap`, potentially removing the `LastLoginFriendly`, etc.) but let me know when you're ready for me to do another full review and I'd be happy to.

Keep up the great work! This is really quickly getting in great shape.

models/imap.go
glennzw: Renamed:

```
DeleteCampaign    bool      `json:"delete_campaign"`
```

to:

```
DeleteReportedCampaignEmail bool      `json:"delete_reported_campaign_email"`
```

models/imap.go
glennzw: I've moved this validation testing (attempting to login to the IMAP server) to the new `gophihs/imap` package. There's still some basic testing in this `models/imap.go` file, such as checking if the supplied variables are blank or if the port is in allowed range. I presume these are OK to be left here? (I copied this from other files in the `controllers` package.

controllers/imapmon.go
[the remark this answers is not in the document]
glennzw: I've created the `gophish\imap` package and moved functionality there. (I just copied in the single `eazye.go` file and added the `ValidateIMAP()` from the controller).

Also renamed `ImapMonitor` to just `Monitor`.

Next up: remove unused functionality in the new `imap.go` file and try use your `email.NewReader` to parse emails.

Commit: https://github.com/glennzw/gophish/commit/7ddba13507fab0fc2b5359afb929b9ec3793a201

models/imap.go
glennzw: The `LastLoginFriendly` database field isn't actually used to store data, it's defined in the IMAP struct so I can return the JSON value below:

```
ss[0].LastLoginFriendly = humanize.Time(ss[0].LastLogin)
delta := time.Now().Sub(ss[0].LastLogin).Hours() // Default value if never logged in is "0001-01-01T00:00:00Z"
if delta > 87600 {
	ss[0].LastLoginFriendly = "Never" // Well, either Never or > 10 years ago.
}
```
(See the code [here](https://github.com/glennzw/gophish/blob/imap/controllers/api/imap.go#L46))

I'm not sure if there's a way to return additional fields without having them defined in the imap struct?

The reason I ended up doing it this way was: (i) to avoid time differences between client local time and server time (including thinking about timezones),  (ii) it was easier to handle on the server side, (iii) `moment.js` doesn't have granularity below 44 seconds:

Screen Shot 2019-10-28 at 13 33 57
 
So two questions maybe someone can help with:

- Can I include a JSON response key/pair that's not defined in the IMAP struct
- Do you know how to wrangle `moment.js` to have greater granularity below 44 seconds?
- Rather, if you think the `moment.js` granularity is fine we can just use that

Alternatively @jordan-wright if you don't think this is useful I can remove it. I found it useful to have in the UI to know if Gophish was successfully logging in when configuring my IMAP server.

static/js/dist/app/settings.min.js
glennzw: Well, I tested this again and it works now. Not sure why it wasn't before. Updated.

controllers/imapmon.go
glennzw: We don't need the `GetEnabledIMAPs()` function with the new polling logic, so removed this and all seems fine.

glennzw:
Hi @jordan-wright, I've resolved all the reviews I was able to. Looking for comments on:

1. Created `gophish\imap` package https://github.com/gophish/gophish/pull/1612#discussion_r332308441
2. Reporting campaign via db or http query https://github.com/gophish/gophish/pull/1612#discussion_r332311631
3. Friendly last login https://github.com/gophish/gophish/pull/1612#discussion_r332312139

Thanks!

templates/settings.html
```diff
@@ -82,6 +84,142 @@ <h1 class="page-header">Settings</h1>
                 <label for="use_map">Show campaign results map</label>
             </div>
         </div>
+        <!-- Reporting Settings Begin -->
+        <div role="tabpanel" class="tab-pane" id="reportingSettings">
+            <form id="imapForm" >
+            <br />
+            <div class="row">
+                <div>
```
jordan-wright: For these divs, it's recommended to add the preferred [column class](https://getbootstrap.com/docs/3.3/css/#grid-options). You'll see this used by the other tabs, and for this case it's essentially it's there to indent the elements to the right a little bit, since the `row` gives it a negative margin.

static/js/src/app/settings.js
```diff
@@ -21,9 +22,229 @@ $(document).ready(function () {
...
+
+      }); //end testclick
+
+    $("#reporttab").click(function() {
+        loadIMAPSettings()
+    })
+
+    $("#advanced").click(function() {
+        $("#advancedarea").toggle("slow");
+        $("html, body").animate({ scrollTop: $(document).height() }, 1000);
```
jordan-wright: Could we avoid animating this? It's kind of jarring with how slow it is. My personal preference would be to change this to `toggle()` and remove the scroll animation following it. It's ok if someone clicks the link and the settings drop down below- the intuitive things, IMO, is that someone would then scroll to see the added settings that they can modify.

imap/imap.go
```diff
@@ -0,0 +1,630 @@
...
+	mailSettings := MailboxInfo{
+		Host:   s.Host,
+		TLS:    s.TLS,
+		User:   s.Username,
+		Pwd:    s.Password,
+		Folder: s.Folder}
+
+	client, err := newIMAPClient(mailSettings)
+	defer func() {
+		client.Close(true)
```
jordan-wright: By calling this every time the function runs, you run the risk of causing crashes if `err != nil`. In my case, during testing I gave invalid IMAP settings which threw an error, setting `client` to `nil`. This caused a crash since we're trying to call `nil.Close()`.

Instead, in this case we should check the error first, returning it if it's not nil and then either call the defer function or, probably easier here, just run those cleanup functions directly since there isn't any code to run after error checking.

imap/imap.go
```diff
@@ -0,0 +1,630 @@
...
+}
+
+var GenerateBufferSize = 100
+
+func generateMail(info MailboxInfo, search string, since *time.Time, markAsRead, delete bool) (chan Response, error) {
+	responses := make(chan Response, GenerateBufferSize)
+	client, err := newIMAPClient(info)
+	if err != nil {
+		close(responses)
+		return responses, fmt.Errorf("uid search failed: %s", err)
```
jordan-wright: Minor nit - Is this the right error here? The "uid search failed" threw me off, since this should just be connecting to the IMAP server.

imap/monitor.go
```diff
@@ -0,0 +1,212 @@
...
+	im.cancel()
+	return nil
+}
+
+// checkForNewEmails logs into an IMAP account and checks unread emails
+//  for the rid campaign identifier.
+func checkForNewEmails(im models.IMAP, reportURL string) {
+
+	mailSettings := MailboxInfo{
+		Host:   im.Host,
```
jordan-wright: There's a bug right now where the `Port` in the `models.IMAP` object doesn't appear to be used in the `MailboxInfo` object. This appears to be the case in both the validation and general IMAP monitoring.

This causes `newIMAPClient` to call `DialTLS` or `Dial` with just the hostname, causing `github.com/mxk/go-imap/imap` to choose the default port (993 for TLS, 143 without).

This appears to only affect the IMAP polling. You do add the port in the IMAP validation.

templates/settings.html
glennzw: Done.

static/js/src/app/settings.js
glennzw: Sure no problem! Done.

glennzw:
Thanks for taking the time to give feedback @jordan-wright, I addressed the issues in https://github.com/gophish/gophish/pull/1612/commits/81f750ec18e02cfe8c6be83f650c28b8febf8ac8

jordan-wright:
Awesome, thanks @glennzw!

Starting last night, I'm finally getting a chance to really dig into the code and take it for a test drive, hence all the smaller comments that you'll start to see coming through. Over the next few days, I'm sure we can get it to a great place and (hopefully!) see about having it merged as early as the end of this week 🎊 

Keep up the great work!

models/imap.go
```diff
@@ -0,0 +1,152 @@
...
+func DeleteIMAP(uid int64) error {
+	err := db.Where("user_id=?", uid).Delete(&IMAP{}).Error
+	if err != nil {
+		log.Error(err)
+	}
+	return err
+}
+
+func SuccessfulLogin(s *IMAP) error {
+	err := db.Model(&s).Update("last_login", time.Now()).Error
```
jordan-wright:
```suggestion
	err := db.Model(&s).Update("last_login", time.Now().UTC()).Error
```

We try to use UTC everywhere else, so it'd be good to keep things consistent here.

jordan-wright:
Ok, I think I figured out how we can remove the `LastLoginFriendly`. Here's what I'd recommend:

* Remove that field entirely
* Avoid polling every second like we're doing from the front-end. This is a personal opinion only from the standpoint that it puts (not much, but some) load on the Gophish server just to support having near-realtime data for a text box that's hidden by default. Alternatively, if you think polling is required, then perhaps we can change the frequency from 1 second to the retrieved IMAP polling frequency (default 60 seconds)
* When the data is loaded, you can get a relative offset using `moment.utc(imap[0].last_login).fromNow()` I'll add a suggestion for this in the PR.

Using moment.js gives you output very similar to what you already had. You can see an example of this [here](https://momentjs.com/docs/#/displaying/fromnow/).

To me, keeping this on the frontend will make the API more consistent and easier to reason about. Plus it removes a dependency which I'm all about 😛

static/js/src/app/settings.js
```diff
@@ -21,9 +22,228 @@ $(document).ready(function () {
...
+        $("#advancedarea").toggle();
+    })
+
+    function lastLoginTicker(){
+
+        if ($("#lastlogin").is(":visible") && $('#use_imap').prop('checked') == true) {
+            api.IMAP.get()
+            .success(function (imap) {
+                if (imap.length > 0){
+                    $('#lastlogin').val(imap[0].last_login_friendly)
```
jordan-wright:
```suggestion
                    $('#lastlogin').val(moment.utc(imap[0].last_login).fromNow())
```

models/imap.go
glennzw: Done ✅

imap/imap.go
```diff
@@ -0,0 +1,628 @@
...
+// all headers included.
+func NewEmail(msgFields imap.FieldMap) (Email, error) {
+	var email Email
+	// parse the header
+	var message bytes.Buffer
+	message.Write(imap.AsBytes(msgFields["RFC822.HEADER"]))
+	message.Write([]byte("\n\n"))
+	rawBody := imap.AsBytes(msgFields["BODY[]"])
+	message.Write(rawBody)
+	msg, err := mail.ReadMessage(&message)
```
jordan-wright: I'd recommend that we defer all the parsing below this line to the `github.com/jordan-wright/email` package we already use in Gophish. Specifically, the call to `email.NewEmailFromReader` appears to work, which returns an [`email.Email`](https://godoc.org/github.com/jordan-wright/email#Email) instance which has all the fields you'd expect (From, Subject, HTML, Body, etc.). This will allow us to remove a ton of parsing code in here, as well as multiple external dependencies.

jordan-wright:
When it comes to the best way to handle the email reporting, I think we should lean towards doing things directly in the database. Sure, it removes a web access log line, but I think that's more accurate since this wasn't reported through the external API. This also removes some complexity and room for things to fail.

Instead, we could consider how to best add structured log messages to something like the `result.createEvent` or `campaign.AddEvent` functions. I'd view this as something to knock out separate from this PR. For this PR, just adding a simple log line to `result.HandleEmailReport` will be fine and we can figure out the wider details later.

So I'd use `models.GetResult` to grab the result from the `rid`, then use `result.HandleEmailReport` to add the email reported event.

static/js/src/app/settings.js
glennzw: Done. I removed the ticker altogether. I also created a hidden field in the HTML page to store the 'raw' last_login value in order to pass it back to the server in the save operation. In hindsight I maybe should have used Gorm's Update, as I think that only updates changed fields.

Also fixed a bug where one successful login updated all users' successful logins.

Commit: https://github.com/gophish/gophish/pull/1612/commits/2cdc12e94e3ad7473ffc84a0d3a7098521439ab4

glennzw:
Updated the reporting per your suggestions, see https://github.com/gophish/gophish/pull/1612/commits/086517ed8f41164060928e7f3ca1476422e99f9a

I'm not sure if I should populate something in the `EventDetails{}` (for the Payload and Browser fields):

`err = result.HandleEmailReport(models.EventDetails{})`

I log with`§log.Infof`, e.g:  

`time="2019-11-06T13:12:58Z" level=info msg="User gophish.test.9001@gmail.com reported email with rid 7twvLyL"`

I left in the functionality to mark failed reported emails as unread, although I suspect that's far more unlikely than with the web calls.

imap/imap.go
glennzw: The one thing your email struct doesn't have is message UID. I [added this](https://github.com/jprobinson/eazye/blob/master/eazye.go#L188) to the `eazye` email package in order to delete campaign emails (or mark emails as unread that failed to report). Maybe you could add that field to your [email.Email](https://godoc.org/github.com/jordan-wright/email#Email) ?

Also if we're going down this route I wonder if it's worth rewriting the whole of `imap.go` with [go-imap](https://github.com/emersion/go-imap) and your `github.com/jordan-wright/email` 🤔

jordan-wright: I do agree that rewriting might be the best option. But to me the biggest hold up for this particular PR is the number of dependencies it adds just to handle the mail parsing. If we could remove the parsing functionality alone and instead use the existing `github.com/jordan-wright/email` then that'd be better, IMO. We can work on longer term rewrites later.

I mean, honestly if we think the effort is worth it and you'd be open to it, I don't see why we couldn't eventually make a new entire repo at `gophish/imap` that does just that - making IMAP more accessible. I could give write access so you could contribute to that as you'd see fit. Just an option to consider 🤷‍♂ 

For adding the UID, I can see the benefit. The reason I'd err away from having that in the email package (for now, until I've had more time to consider it) is that it's really meant to construct emails that are sent, whereas UID is an IMAP-specific field.

Perhaps the better short-term option would be to make a struct like this:

```
type IMAPEmail struct {
    UID          uint32          `json:"uid"`
    *email.Email
}
```

Then read the `*email.Email` from `NewEmailFromReader`, and use that to construct your `IMAPEmail`, which is used throughout this package for parsing. It'll still have all the other `email.Email` fields and functions, but with the addition of a `UID` field.

imap/monitor.go
```diff
@@ -0,0 +1,194 @@
...
+
+			if rid != "" {
+				rid = rid[5:]
+				log.Infof("User '%s' reported email with rid %s", m.From.Address, rid)
+				result, err := models.GetResult(rid)
+				if err != nil {
+					log.Error("Error reporting GoPhish email with rid ", rid, ": ", err.Error())
+					reportingFailed = append(reportingFailed, m.UID)
+				} else {
+					err = result.HandleEmailReport(models.EventDetails{}) // Not sure if we should populate something into the EventDetails{}
```
jordan-wright: > `// Not sure if we should populate something into the EventDetails{}`

Nah, since there isn't any payload sending an empty `EventDetails` is correct.

imap/imap.go
glennzw: Hi @jordan-wright, thanks for your patience on this - it's been hard to find dedicated time to properly sit down and tackle this. I've now implemented your email library and removed a lot of code that was imported from `eazye`. I've done basic testing and it works well; I haven't tested all error conditions yet but wanted to get it pushed out ASAP.

See: https://github.com/glennzw/gophish/commit/c6a629079fe1198c875a654d6b0e998f840b88a6

jordan-wright:
Thanks @glennzw!

A quick glance looks great, but I'll do a full review here in the next day or two. Thanks, as always, for working so hard on this!

imap/imap.go
```diff
@@ -0,0 +1,346 @@
...
+
+	log "github.com/gophish/gophish/logger"
+	"github.com/gophish/gophish/models"
+	"github.com/jordan-wright/email"
+	"github.com/mxk/go-imap/imap"
+
+	_ "github.com/paulrosania/go-charset/data"
+)
+
+// IMAPEmail Email struct with IMAP UID
```
jordan-wright: I think throughout the `imap` package, there's room to clean up what we call things. For example, here we can just call this `Email`. This makes it easier to use outside of the package since it wouldn't be as repetitive (`imap.Email` vs. `imap.IMAPEmail`).

The same goes for things like:

* `ValidateIMAP` -> `Validate`
* `newIMAPClient` -> `newClient`
* `monitorIMAP` -> `monitor`

static/js/dist/app/settings.min.js
```diff
@@ -1 +1,233 @@
...
+                Swal.fire({
+                    title: "Failed!",
+                    html: "Unable to login to <b>" + $("#imaphost").val() + "</b>.",
+                    type: "error",
+                    showCancelButton: true,
+                    cancelButtonText: "Close",
+                    confirmButtonText: "More Info",
+                    confirmButtonColor: "#428bca",
+                    allowOutsideClick: false,
+                }).then((result) => {
```
jordan-wright: Heads up that we don't currently support ES6 constructs in settings.js. It's a WIP to migrate everything over to the new webpack + Babel workflow. The hard part is making sure that we don't break anything in the process due to things like different `this` scoping.

static/js/dist/app/settings.min.js
```diff
@@ -1 +1,233 @@
-$(document).ready(function(){$("#apiResetForm").submit(function(e){return api.reset().success(function(e){user.api_key=e.data,successFlash(e.message),$("#api_key").val(user.api_key)}).error(function(e){errorFlash(e.message)}),!1}),$("#settingsForm").submit(function(e){return $.post("/settings",$(this).serialize()).done(function(e){successFlash(e.message)}).fail(function(e){errorFlash(e.responseJSON.message)}),!1});var e=localStorage.getItem("gophish.use_map");$("#use_map").prop("checked",JSON.parse(e)),$("#use_map").on("change",function(){localStorage.setItem("gophish.use_map",JSON.stringify(this.checked))})});
\ No newline at end of file
+$(document).ready(function () {
```
jordan-wright: No need to remove it now if it's helping you test, but before we merge I'd ask that you please remove the *.min.js files from the PR. I'll rebuild everything on my end after merging the initial PR.

jordan-wright (changes requested):
Hey @glennzw!

I took a bit of time to go through and mainly look for any obvious styling changes and have added suggestions as comments here. Let me know if you have any questions!

My hope is to do full end-to-end functionality testing sometime this weekend.

controllers/api/imap.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/http"
+	"time"
+
+	ctx "github.com/gophish/gophish/context"
+	"github.com/gophish/gophish/imap"
+	log "github.com/gophish/gophish/logger"
+	"github.com/gophish/gophish/models"
+)
+
+// ImapServerValidate handles requests for the /api/imapserver/validate endpoint
```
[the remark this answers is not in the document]
glennzw: Done

controllers/api/imap.go
```diff
@@ -0,0 +1,64 @@
...
+		err = imap.ValidateIMAP(&s)
+		if err != nil {
+			JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusOK)
+			return
+		}
+		JSONResponse(w, models.Response{Success: true, Message: "Successful login."}, http.StatusCreated)
+	}
+}
+
+// ImapServer handles requests for the /api/imapserver/ endpoint
```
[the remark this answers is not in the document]
glennzw: Done

controllers/api/imap.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/gophish/gophish/models"
+)
+
+// ImapServerValidate handles requests for the /api/imapserver/validate endpoint
+func (as *Server) ImapServerValidate(w http.ResponseWriter, r *http.Request) {
+	switch {
+	case r.Method == "GET":
+		JSONResponse(w, models.Response{Success: false, Message: "Only POSTs allowed"}, http.StatusBadRequest)
+	case r.Method == "POST":
+		s := models.IMAP{}
```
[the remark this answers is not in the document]
glennzw: Good point, changed `s` variable to `im`

imap/imap.go
glennzw: Good point. go-lint had told me off for this. Done.

imap/imap.go
```diff
@@ -0,0 +1,346 @@
...
+	"fmt"
+	"strconv"
+	"time"
+
+	log "github.com/gophish/gophish/logger"
+	"github.com/gophish/gophish/models"
+	"github.com/jordan-wright/email"
+	"github.com/mxk/go-imap/imap"
+
+	_ "github.com/paulrosania/go-charset/data"
```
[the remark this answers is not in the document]
glennzw: Hmm that's a leftover from `eazye` it seems: 

https://github.com/jprobinson/eazye/blob/master/eazye.go#L19

Fairly sure we can remove that as I think it was only used in the [parseBody()](https://github.com/jprobinson/eazye/blob/master/eazye.go#L520) function, which I removed during the cleanup.

static/js/dist/app/settings.min.js
glennzw: Cool updated to `then(function(result) {` syntax.

Also just noticed hitting 'Close' on that dialog wasn't closing it as I wasn't checking the result value, fixed that too.

static/js/dist/app/settings.min.js
glennzw: Ah yes, I'm never sure how to handle `.min.js` files in my commits. Shall I delete the min files in my PR, or check them out from their originals?

imap/imap.go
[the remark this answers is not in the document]
glennzw: I removed the import and everything still seems to hold together.

glennzw:
Thanks @jordan-wright! I think I've addressed the suggestions you had https://github.com/glennzw/gophish/commit/c5f19654306b83e9d20a23edbe7ff72861b14fea

jordan-wright:
Heads up that I encountered a bug where I couldn't save new IMAP settings. It appears the root cause is that we try to send the `LastLogin` property [here](https://github.com/gophish/gophish/pull/1612/files#diff-77eea3462821f273c7ef3b7d20965a41R40) which becomes an empty string. This causes Go to fail when trying to parse it as a time.

Personally, I'm not sure we should keep the lastlogin time if the user changed the settings. To that end, I'd recommend just removing it from the POST body altogether. That seems to fix things for me. 🤷‍♂

jordan-wright:
Another nice to have would be to add an interface called `Client` (or `Provider`... I'm bad with names 😄) in the `imap` package which looks something like this:

```go
type Client interface {
    Close(expunge bool) (cmd *imap.Command, err error)
    Login(username, password string) (cmd *imap.Command, err error)
    Logout(timeout time.Duration) (cmd *imap.Command, err error)
    Select(mbox string, readonly bool) (cmd *imap.Command, err error)
    UIDFetch(seq *imap.SeqSet, items ...string) (cmd *imap.Command, err error)
    UIDSearch(spec ...imap.Field) (cmd *imap.Command, err error)
    UIDStore(seq *imap.SeqSet, item string, value Field) (cmd *imap.Command, err error)
}
```

(Note: I generated this list from whatever functions we currently use from the `imap.Client`, so it's possible this isn't a complete list).

Then, each function that currently accepts an `*imap.Client` (such as `alterEmail`) is changed to receive a `Client`. Right now, there's no tests which isn't a huge deal, but making this change will make it easier to write tests in the future, since we could just make a mock `Client` that can give us whatever responses we want and pass that to the various functions.

imap/imap.go
```diff
@@ -0,0 +1,344 @@
...
+	TLS    bool
+	User   string
+	Pwd    string
+	Folder string
+	// Read only mode, false (original logic) if not initialized
+	ReadOnly bool
+}
+
+// GetAll will pull all emails from the email folder and return them as a list.
+func GetAll(info MailboxInfo, markAsRead, delete bool) ([]Email, error) {
```
jordan-wright: What do you think about making all the functions here that have `MailboxInfo` as their first argument instead be functions off a `*MailboxInfo`, eg:

```go
func (mbox *MailboxInfo) GetAll(markAsRead, delete bool) ([]Email, error) {}
```

It seems like that might be a bit cleaner? Additionally, we could rename `MailboxInfo` to just `Mailbox`, since it seems to logically represent a mailbox. I'm all for simpler struct names where possible!

Both of these are 100% your call- just suggestions from me.

controllers/api/imap.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// IMAPServerValidate handles requests for the /api/imapserver/validate endpoint
+func (as *Server) IMAPServerValidate(w http.ResponseWriter, r *http.Request) {
+	switch {
+	case r.Method == "GET":
+		JSONResponse(w, models.Response{Success: false, Message: "Only POSTs allowed"}, http.StatusBadRequest)
+	case r.Method == "POST":
+		im := models.IMAP{}
+		// Put the request into a page
```
jordan-wright: Small nit - this comment looks like it came from the `Pages` handler.

glennzw:
> Another nice to have would be to add an interface called `Client` (or `Provider`... I'm bad with names 😄) in the `imap` package which looks something like this:
> 
> ```go
> type Client interface {
>     Close(expunge bool) (cmd *imap.Command, err error)
>     Login(username, password string) (cmd *imap.Command, err error)
>     Logout(timeout time.Duration) (cmd *imap.Command, err error)
>     Select(mbox string, readonly bool) (cmd *imap.Command, err error)
>     UIDFetch(seq *imap.SeqSet, items ...string) (cmd *imap.Command, err error)
>     UIDSearch(spec ...imap.Field) (cmd *imap.Command, err error)
>     UIDStore(seq *imap.SeqSet, item string, value Field) (cmd *imap.Command, err error)
> }
> ```
> 
> (Note: I generated this list from whatever functions we currently use from the `imap.Client`, so it's possible this isn't a complete list).
> 
> Then, each function that currently accepts an `*imap.Client` (such as `alterEmail`) is changed to receive a `Client`. Right now, there's no tests which isn't a huge deal, but making this change will make it easier to write tests in the future, since we could just make a mock `Client` that can give us whatever responses we want and pass that to the various functions.

I'm a little new to interfaces and have been trying to fully understand this one, but I've implemented it as you suggested and it seems to work 🙃

controllers/api/imap.go
glennzw: Removed

imap/imap.go
glennzw: Ah yes much neater this way, thanks! Done.

glennzw:
Addressed the new issues, let me know if you find any more 🙂

imap/imap.go
```diff
@@ -0,0 +1,365 @@
...
+	Close(expunge bool) (cmd *imap.Command, err error)
+	Login(username, password string) (cmd *imap.Command, err error)
+	Logout(timeout time.Duration) (cmd *imap.Command, err error)
+	Select(mbox string, readonly bool) (cmd *imap.Command, err error)
+	UIDFetch(seq *imap.SeqSet, items ...string) (cmd *imap.Command, err error)
+	UIDSearch(spec ...imap.Field) (cmd *imap.Command, err error)
+	UIDStore(seq *imap.SeqSet, item string, value imap.Field) (cmd *imap.Command, err error)
+}
+
+// UIDStore stores UIDs
```
jordan-wright: Looks like this bit of commented code slipped into the commit.

imap/imap.go
```diff
@@ -0,0 +1,365 @@
...
+//func UIDStore(seq *imap.SeqSet, item string, value imap.Field) (cmd *imap.Command, err error) {
+//	return nil, nil
+//}
+
+// Email Email struct with IMAP UID
```
jordan-wright:
```suggestion
// Email represents an email.Email with an included IMAP UID
```

imap/imap.go
```diff
@@ -0,0 +1,365 @@
...
+	cmd, err := imap.Wait(client.UIDSearch(specs...))
+	if err != nil {
+		return &imap.Command{}, fmt.Errorf("uid search failed: %s", err)
+	}
+	return cmd, nil
+}
+
+var GenerateBufferSize = 100
+
+//func generateMail(info Mailbox, search string, since *time.Time, markAsRead, delete bool) (chan Response, error) {
```
jordan-wright: Small nit - looks like some commented code slipped in here.

imap/imap.go
```diff
@@ -0,0 +1,365 @@
...
+
+	// get headers and UID for UnSeen message in src inbox...
+	cmd, err := imap.Wait(client.UIDSearch(specs...))
+	if err != nil {
+		return &imap.Command{}, fmt.Errorf("uid search failed: %s", err)
+	}
+	return cmd, nil
+}
+
+var GenerateBufferSize = 100
```
jordan-wright: You can probably make this a `const`

models/imap.go
```diff
@@ -0,0 +1,151 @@
...
+var ErrIMAPUsernameNotSpecified = errors.New("No Username specified")
+
+// ErrIMAPPasswordNotSpecified is thrown when there is no Password specified
+// in the IMAP configuration
+var ErrIMAPPasswordNotSpecified = errors.New("No Password specified")
+
+var ErrInvalidIMAPFreq = errors.New("Invalid polling frequency.")
+
+// TableName specifies the database tablename for Gorm to use
+func (s IMAP) TableName() string {
```
jordan-wright: Throughout this file it looks like there's multiple instances where we call an `IMAP` instance `s`, which probably isn't the best name for it. Could we change these to `im` and, for plural as needed, `ims`? That would be a bit more descriptive when we need to modify the code later on.

jordan-wright:
Hey @glennzw!

This is still looking great. We're definitely in the home stretch. I've added a few more comments for your review, but things are really coming along nicely.

Thank you so much for all your great work on this!

imap/imap.go
glennzw: Done

imap/imap.go
glennzw: Changed

imap/imap.go
glennzw: Done

imap/imap.go
glennzw: Done

models/imap.go
glennzw: Good point, changed.

glennzw:
Addressed the new issues, some sloppy ones on my side there!

jordan-wright:
Hey @glennzw!

Sorry for the delay- things have been quite busy over here. I should have time to review this today or tomorrow. I think the only outstanding item is removing the minified JS from the commits. I'm honestly not sure what the easiest way is to do that. I think there's a way to reset (or just checkout again) the files from master and then re-commit.

I'll do some digging on my end tomorrow if you haven't found an answer by then. I'm sure we can figure it out together 😄 

Thanks so much for your patience!

glennzw:
Hey @jordan-wright! 

No worries at all, that's the beauty of OSS projects, we give when we have time :) I did a checkout against the .min.js files from the original branch I forked in my repo, hopefully that works 😬

```
git checkout 24fe998a3aa04e205900476a9601d481e94d8eea settings.min.js
git checkout 24fe998a3aa04e205900476a9601d481e94d8eea gophish.min.js
git checkout 24fe998a3aa04e205900476a9601d481e94d8eea vendor.min.js
```

controllers/api/imap.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+}
+
+// IMAPServer handles requests for the /api/imapserver/ endpoint
+func (as *Server) IMAPServer(w http.ResponseWriter, r *http.Request) {
+	switch {
+	case r.Method == "GET":
+		ss, err := models.GetIMAP(ctx.Get(r, "user_id").(int64))
+		if err != nil {
+			log.Error(err)
```
jordan-wright: We should probably send back a `JSONResponse` with the error here and the status for an internal server error, then return.

db/db_mysql/migrations/20191022184235_0.9.0_imap.sql
```diff
@@ -0,0 +1,8 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+CREATE TABLE IF NOT EXISTS `imap` (user_id bigint,host varchar(255),port int,username varchar(255),password varchar(255),modified_date datetime,tls boolean,enabled boolean,folder varchar(255,restrict_domain varchar(255),delete_reported_campaign_email boolean,last_login datetime,imap_freq int);
```
jordan-wright:
```suggestion
CREATE TABLE IF NOT EXISTS `imap` (user_id bigint,host varchar(255),port int,username varchar(255),password varchar(255),modified_date datetime,tls boolean,enabled boolean,folder varchar(255),restrict_domain varchar(255),delete_reported_campaign_email boolean,last_login datetime,imap_freq int);
```

It looks like you're missing a parenthesis here causing the SQL to be invalid.

imap/imap.go
```diff
@@ -0,0 +1,353 @@
...
+	seq := &imap.SeqSet{}
+	msgCount := 0
+	for _, rsp := range cmd.Data {
+		for _, uid := range rsp.SearchResults() {
+			msgCount++
+			seq.AddNum(uid)
+		}
+	}
+
+	// nothing to request?! why you even callin me, foolio?
```
jordan-wright: While this made me laugh, let's remove this for now to keep things as clean as possible 😆

imap/imap.go
```diff
@@ -0,0 +1,353 @@
...
+// will expect the message to container the internaldate and the body with
+// all headers included.
+func NewEmail(msgFields imap.FieldMap) (Email, error) {
+
+	rawBody := imap.AsBytes(msgFields["BODY[]"])
+
+	rawBodyStream := bytes.NewReader(rawBody)
+	em, err := email.NewEmailFromReader(rawBodyStream) // Parse with @jordanwright's library
+	if err != nil {
+		log.Error("Unable to parse email")
```
jordan-wright: It seems like we'd probably want to return the error here, right? I could be wrong, so I'd like to get your take on it.

models/imap.go
```diff
@@ -0,0 +1,151 @@
...
+var ErrInvalidIMAPPort = errors.New("Invalid IMAP Port")
+
+// ErrIMAPUsernameNotSpecified is thrown when there is no Username specified
+// in the IMAP configuration
+var ErrIMAPUsernameNotSpecified = errors.New("No Username specified")
+
+// ErrIMAPPasswordNotSpecified is thrown when there is no Password specified
+// in the IMAP configuration
+var ErrIMAPPasswordNotSpecified = errors.New("No Password specified")
+
```
jordan-wright: Small nit - missing comment here.

jordan-wright:
Hey @glennzw,

I left a few comments on the diff- things like a bug in the MySQL migration and some small nits. And it looks like all of the minified JS is gone except for vendor.min.js, so if we could remove that as well that'd be great.

With these things out of the way, I think we'll be in a position where we can merge this in 🎊

imap/imap.go
glennzw: Done 🙃

controllers/api/imap.go
glennzw: Good point, not sure why I didn't. Fixed.

imap/imap.go
glennzw: Right you are!

models/imap.go
glennzw: Fixed

glennzw:
Hi @jordan-wright,

Thanks for spotting those little mistakes, I really should have caught them!

Hmm for the `vendor.min.js` it looks like it matches the branch from which I originally forked (Fix multiple XSS issues):

```
commit 28ba68c9163f5ab02ba75814bac9122205f823f5
Author: Glenn Wilkinson <glenn.wilkinson@gmail.com>
Date:   Thu Aug 29 12:48:05 2019 +0100

    Added functionality to manage IMAP settings via UI

commit 24fe998a3aa04e205900476a9601d481e94d8eea (origin/master, origin/HEAD)
Author: David Maciejak <david.maciejak@gmail.com>
Date:   Sat Aug 24 10:07:15 2019 +0800

    Fix multiple XSS issues in User Management Page (#1547)
    
    If the user name is embedding some JS code, it will be executed on the client side. Note: gophish/static/js/dist/app/users.min.js will need to be regenerated too.
```

(I ran the command `git checkout 24fe998a3aa04e205900476a9601d481e94d8eea vendor.min.js` on my last checkin)

But it looks like the latest modification to `vendor.min.js` is from the `Upgrade SweetAlert2 Dependency (#1583)` commit, so perhaps if I check that file out it'll align better.

I ran `git checkout 6222c5e180a97d0aa3ec345d0c7af06c380f3db5 vendor.min.js ` for my latest commit so let's see.

jordan-wright:
Looks like that worked! The very last thing I noticed was that we'll probably want to update the filename of the SQL migrations. It's my understanding that our migrations library, goose, works based on the timestamp.

Since we added webhook support and those migrations have the timestamp `20191104103306`, I think that folks who use the `master` branch of Gophish wouldn't see these IMAP migrations applied since they have the timestamp `20191022184541`. The fix for this would be to update the date portion of those filenames, e.g. `20200116000000` or so.

Once that's done, things are good to merge! Sorry for all the back and forth. I *really* appreciate all your patience here!

glennzw:
Hi @jordan-wright , whoop almost there! I renamed the two database files 👍

No worries on the back and forth, it's been a great learning experience for me and I've enjoyed it!

jordan-wright:
This works for me! There might be some small tweaks in the future, but that's stuff we can figure out later. I think this is a stellar first pass.

Thank you so much for all the hard work and patience!

glennzw:
Whooop, that's brilliant, I'm so happy to have contributed to the project! Thanks for all the guidance and code reviews! I hope it's a feature people find useful 😃

Also yes I'm sure there will be more tweaks and fixes once people start using it - I'll keep an eye on it!

wshi4440:
I downloaded 0.90 and tried this new feature.  But I got this error "tls: first record does not look like a TLS handshake".  
Did I miss any steps?  Do I need to import server cert to the machine where gophish is running?

PR #1635, opened 2019-10-28
Exit on port binding failure

glennzw:
There have been a few issues raised recently with people trying to bind to port 80 without root permissions, as Gophish currently fails silently on this condition.

For this pull request I wrapped `log.Fatal()` around the various `ListenAndServe` functions. Also removed the `error` return as log.Fatal will immediately exit.


jordan-wright:
Nice! I had thought about how I might do a more complicated graceful shutdown kind of thing, but this seems simpler and should work fine.

Thanks for sending this in!

Issue #1570, opened 2019-09-03
gophish getting crashed automatically


Allen-Paul:
What version of Gophish are you using?: 
0.7.1

Brief description of the issue: 
getting crashed automatically. 

What are you expecting to see happen? 
Work without getting crash, launch and track campaigns.

What are you seeing happen?
When I run GoPhish, after 1 minute it is getting crashed. I started facing this issue from today.
When I went through all the issues related to this, I found that the issue was mostly related to errors in SMTP name, FROM address, and {{.URL}}.
I've made sure it's not caused because of anything mentioned above.

Please provide any terminal output that may be relevant below:

time="2019-09-03T10:25:29Z" level=warning msg="No contact address has been configured."
time="2019-09-03T10:25:29Z" level=warning msg="Please consider adding a contact_address entry in your config.json"
goose: no migrations to run. current version: 20190105192341
time="2019-09-03T10:25:29Z" level=info msg="Starting admin server at https://XXX.XXX.XXX.XXX:3333"
time="2019-09-03T10:25:29Z" level=info msg="Background Worker Started Successfully - Waiting for Campaigns"
time="2019-09-03T10:25:29Z" level=info msg="Starting phishing server at http://XXX.XXX.XXX.XXX:4444"
time="2019-09-03T10:26:29Z" level=warning msg="record not found: template not found for campaign"
time="2019-09-03T10:26:29Z" level=warning msg="record not found: template not found for campaign"
time="2019-09-03T10:26:29Z" level=info msg="Sending emails to mailer for processing" num_emails=2
time="2019-09-03T10:26:29Z" level=warning msg="record not found: template not found for campaign"
time="2019-09-03T10:26:29Z" level=info msg="Sending emails to mailer for processing" num_emails=3
time="2019-09-03T10:26:29Z" level=warning msg="record not found: template not found for campaign"
time="2019-09-03T10:26:30Z" level=warning msg="record not found: template not found for campaign"
time="2019-09-03T10:26:30Z" level=warning msg="record not found: template not found for campaign"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x93d74c]

goroutine 28 [running]:
github.com/gophish/gophish/models.NewPhishingTemplateContext(0xc69460, 0xc0001e8580, 0xc000277be0, 0x1d, 0xc000429487, 0x5, 0xc00042948c, 0x1, 0xc000429498, 0x4, ...)
[...]

PR #1640, opened 2019-10-29
Bugfix Issue 1570

SamSutt:
Regarding Issue 1570 (https://github.com/gophish/gophish/issues/1570) "gophish getting crashed automatically": This could happen to anybody. Any User of an instance could terminate the whole gophish process. This could badly influence other actual running campaigns. This fix will prevent gophish from terminating, when a template is deleted while mails still scheduled.


jordan-wright:
LGTM - thanks for sending this in!

PR #1642, opened 2019-10-31
WIP: Add webhook

GildedHonour:


controllers/api/server.go
```diff
@@ -71,6 +71,8 @@ func (as *Server) registerRoutes() {
 	router.HandleFunc("/import/group", as.ImportGroup)
 	router.HandleFunc("/import/email", as.ImportEmail)
 	router.HandleFunc("/import/site", as.ImportSite)
+  router.HandleFunc("/webhooks/", mid.Use(as.Webhooks, mid.RequirePermission(models.PermissionModifySystem)))
```
jordan-wright: You might want to add `goimports` to your IDE settings to run automatically on save. I've found this helps me from having to go back later and reformat all the files.

controllers/api/server.go
```diff
@@ -71,6 +71,8 @@ func (as *Server) registerRoutes() {
...
+  router.HandleFunc("/webhooks/{id:[0-9]+}/ping", mid.Use(as.PingWebhook, mid.RequirePermission(models.PermissionModifySystem)))
```
jordan-wright: Let's change this to `validate` instead of `ping`. This will match the pattern that we're doing with the IMAP additions [here](https://github.com/gophish/gophish/pull/1612/files#diff-fab0ab83fa5d22c823eaf5fea7c04cabR52)

controllers/api/webhook.go
```diff
@@ -0,0 +1,35 @@
...
+  "github.com/gophish/gophish/models"
+  // "github.com/gorilla/mux"
+)
+
+
+
+func (as *Server) Webhooks(w http.ResponseWriter, r *http.Request) {
+  switch {
+  case r.Method == "GET":
+    wh_s, err := models.GetWebhooks()
```
jordan-wright: Minor nit - I try not to use `_`'s in variable names where possible. In this case, `whs` would be fine.

controllers/route.go
```diff
@@ -106,6 +106,8 @@ func (as *AdminServer) registerRoutes() {
 	router.HandleFunc("/sending_profiles", mid.Use(as.SendingProfiles, mid.RequireLogin))
 	router.HandleFunc("/settings", mid.Use(as.Settings, mid.RequireLogin))
 	router.HandleFunc("/users", mid.Use(as.UserManagement, mid.RequirePermission(models.PermissionModifySystem), mid.RequireLogin))
+	router.HandleFunc("/webhooks", mid.Use(as.Webhooks, mid.RequirePermission(models.PermissionModifySystem), mid.RequireLogin))
+	router.HandleFunc("/webhooks/{id:[0-9]+}/ping", mid.Use(as.PingWebhook, mid.RequirePermission(models.PermissionModifySystem), mid.RequireLogin))
```
jordan-wright: You likely won't need this. Pinging a webhook could be API-only and a button on the webhooks page.

go.mod
```diff
@@ -0,0 +1,28 @@
+module github.com/gophish/gophish
```
jordan-wright: Please remove the `go.sum` and `go.mod` files before the final review. I'll add dependency management at some point, but that should be separate from the work here.

models/webhook.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+  "errors"
+
+  log "github.com/gophish/gophish/logger"
+)
+
+type Webhook struct {
+  Id    int64  `json:"id"`
+  Title string `json:"url"`
+  Url   string `json:"url"`
```
jordan-wright: You'll want to include a `Secret` here as well to handle signing.

models/webhook.go
```diff
@@ -0,0 +1,36 @@
...
+  if err != nil {
+    log.Error(err)
+    return t, err
+  }
+  return wh, nil
+}
+
+// GetUsers returns the users registered in Gophish
+func GetWebhooks(uid int64) ([]Webhook, error) {
+  wh_s := []Webhook{}
```
jordan-wright: Same minor nit as above - `whs` would work fine here.

webhook/webhook.go
```diff
@@ -0,0 +1,42 @@
+package webhook
+
+type Webhook struct {
+  client *http.Client
+}
+
+//TODO
+
+
+func (whook *Webhook) Send(server string, secret []byte, data interface{}) error {
```
jordan-wright: Super minor nit - I'd rename this to `wh` instead of `whook`. Keeps it cleaner and matches what you have in the `models` package above.

webhook/webhook.go
```diff
@@ -0,0 +1,42 @@
...
+  }
+  defer resp.Body.Close()
+
+  if resp.Status >= 400 {
+    //error
+  }
+
+}
+
+func sign(data interface{}) {
```
jordan-wright: I'd make `sign` be a function off `Webhook`:

```
func (wh *Webhook) sign(data interface{}) (string, error) {}
```

Or something similar.

jordan-wright (commented):
Hey @GildedHonour,

I know this is a WIP, just wanted to pass along feedback as I had it. Keep up the great work!

quelsan:
Nice work and good luck, @GildedHonour!

controllers/api/server.go
GildedHonour: Ok. But using verbs in the URL in REST API  is discouraged. It's recommended to use nouns. It'd be "/validation"

go.mod
GildedHonour: Ok

controllers/api/server.go
jordan-wright: I'd prefer to leave it `validate`. I understand that REST API semantics prefer for everything in the URL to be a resource, but to me `validate` reads cleaner in that we're validating a resource, not referring to a new validation.

I appreciate the suggestion, though! Keep up the great work.

GildedHonour:
@jordan-wright for indentation we use tabs in the project, not spaces, right?

GildedHonour:
RE: 

`func (wh *Webhook) Send(server string, secret []byte, data interface{}) error {}`

the params `server` and `secret` won't be needed then because we already have `(wh *Webhook)`

webhook/webhook.go
```diff
@@ -0,0 +1,59 @@
...
+  signat := wh.sign(data, ts)
+  req.Header.Set("X-Gophish-Signature", signat)
+  req.Header.Set("Content-Type", "application/json")
+  resp, err := wh.client.Do(req)
+  if err != nil {
+    log.Error(err)
+    return err
+  }
+  defer resp.Body.Close()
+  if resp.Status >= MinHttpStatusErrorCode {
```
jordan-wright: TIL about `MinHttpStatusErrorCode`. Nice!

jordan-wright: Oh, nvm, I see you defined it. I thought it was part of the standard `net/http` package 😝

jordan-wright:
> @jordan-wright for indentation we use tabs in the project, not spaces, right?

The easiest answer I can give is that we don't worry about indentation ourselves. I strongly recommend leveraging `goimports` into your IDE to run on save. This is a superset of the functionality provided by `gofmt`, which is the only formatting style recommended for Go. I will be awfully hesitant to merge any code which hasn't been formatted this way (it will be a requirement in our CI pipeline in the not too distant future).

> the params server and secret won't be needed then because we already have (wh *Webhook)

So I guess what's confusing to me is how we plan to break this up. On one hand, I see the benefit in us re-using a single `http.Client` that is configured with reasonable default timeouts. They're goroutine-safe, so it makes sense for us to just have one global one that we send requests through.

However, you say that we have `wh *Webhook`, but this currently only wraps the `http.Client`. In the existing implementation the actual server/secret details will be stored in `models.Webhook`. This isn't a bad thing, but maybe the solution is to do some renaming.

For example, maybe `models.Webhook` should be something like `models.WebhookOptions` or `models.WebhookConfiguration`. Alternatively, we could keep it `models.Webhook` and change `webhook.Webhook` to be `webhook.Transport` or `webhook.Sender` since it's job is to be a server-wide singleton sender that all webhook events go through.

Then, the function signature would look like this (in the case of `webhook.Sender`:

```go
func (sender *Sender) Send(opts models.WebhookOptions, event interface{})
```

Or something similar.

The other thing to consider is the plumbing for testing. It's probably best to leverage interfaces for this. There are a couple of options that I see:

* Changing the `http.Client` in the (current) `webhook.Webhook` to be an interface we define that has only the needed operations and we stub out those for tests.
* Creating an interface called `Sender` that exposes a single `Send` method with the signature above. We could then name the current `webhook.Webhook` struct something like `HTTPSender`, and then create a `MockSender` in tests that mocks out the `Send` method to just store what requests would have been sent for introspection later.

Mainly just thinking out loud here 😄

GildedHonour:
Re: whitespaces vs tabs - alright. But I don't use an IDE. For formatting I use "gofmt". Before merging we'll sort it out.

Re: `webhook.Webhook` - I'll rename it to `webhook.Sender`

Re: for testing - let me see.

GildedHonour:
"validate" vs "ping" webhook in the URL:

In other models we have a method "validate" to validate the names of a models the insert and update operations. Won't it be ambigious to use the URL "/validate" to ping a remote server?

GildedHonour:
I've implemented the CRUD operations on the html page partially.

jordan-wright (commented):
Hey @GildedHonour!

Just wanted to drop a few comments your way. Keep up the great work!

webhook/webhook.go
```diff
@@ -0,0 +1,90 @@
...
+  }
+  jsonData, err := json.Marshal(data)
+  if err != nil {
+    log.Error(err)
+    return err
+  }
+  req, err := http.NewRequest("POST", url, bytes.NewBuffer(jsonData))
+  ts := int32(time.Now().Unix())
+  signat, err := sign(secret, data, ts)
+  req.Header.Set("X-Gophish-Signature", signat)
```
jordan-wright: I'd like us to make the `X-Gophish-Signature` string a constant.

controllers/api/campaign.go
```diff
@@ -41,8 +42,28 @@ func (as *Server) Campaigns(w http.ResponseWriter, r *http.Request) {
 		if c.Status == models.CampaignInProgress {
 			go as.worker.LaunchCampaign(c)
 		}
-		JSONResponse(w, c, http.StatusCreated)
-	}
+
+    //TODO
```
jordan-wright: Just as a heads up, I'd recommend we move the logic to send a webhook event to `models/campaign.go` in `AddEvent`, since this is what is called anytime a campaign event happens.

controllers/api/webhook.go
```diff
@@ -0,0 +1,112 @@
...
+    vars := mux.Vars(r)
+    id, _ := strconv.ParseInt(vars["id"], 0, 64)
+    wh, err := models.GetWebhook(id)
+    if err != nil {
+      log.Error(err)
+      JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusInternalServerError)
+      return
+    }
+
+    httpCnt := &http.Client{} //TODO add timeout and other stuff
```
jordan-wright: I don't think we should create a new `http.Client` and `webhook.Transport` every time a webhook is used. Instead, here's what I'd recommend:

* We create the `Sender` interface in `webhook/webhook.go` as mentioned in another comment.
* We update the `AdminServer` and `api.Server` to have a `Webhook` field that is a `webhook.Sender`.
* We create a `DefaultSender` in `webhook/webhook.go` that is an `HTTPSender` with a properly configured `http.Client`. You can see an example of this in the [net/http package](https://golang.org/src/net/http/client.go#L107)
* We use this `DefaultSender` when creating the admin server. That way, the same sender is reused for webhook calls.
* We have a `WithWebhook` option for the admin server and api server that sets the `Sender`. This will help us out during testing, since we can use it to set the server's webhook provider to a mock sender.

controllers/api/campaign.go
GildedHonour: I wanted to do that initially. 

An issue would be that `AddEvent` does not only what it's supposed to do, but an additional thing. Which would bring issues: 

a) in testing it

b) in situations when we need only `AddEvent` and not (`AddEvent` + send webhook)

Therefore, I'd either a) create `AddEventEx` (AddEvent + webhook) b) call "sendWebhook" in the places where `AddEvent` and where we need a webhook to be sent.

jordan-wright: On second thought, I agree with what you're saying. I think part of the problem is that there is no way right now for `AddEvent` to know which webhook sender to use, requiring us to change the function signature in a very non-ideal way.

Instead, let's just go the manual route and add the `api.Webhook.Send` (using the `Webhook` property introduced in https://github.com/gophish/gophish/pull/1642#discussion_r344469030) wherever we add campaign events. If we want to reduce the amount of code repetition, we could just add a new private method off the API like `api.addCampaignEvent(campaign)` that runs both `AddEvent` and `api.Webhook.Send`, adjusting existing references accordingly.

webhook/webhook.go
```diff
@@ -0,0 +1,90 @@
...
+    errMsg := fmt.Sprintf("http status of response: %s", resp.Status)
+    log.Error(errMsg)
+    return errors.New(errMsg)
+  }
+  return nil
+}
+
+
+//TODO
+func sign(secret string, data interface{}, ts int32) (string, error) {
```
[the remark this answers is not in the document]
GildedHonour: Okay, I've removed a timestamp.

We send body in JSON to a remote server. That is, `json.Marshal(data/payload)` will have to be called anyway, either in `SendWebhook` or anywhere in other methods which call `SendWebhook`. How do you say we can avoid a redundant conversion to JSON?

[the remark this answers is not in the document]
jordan-wright: I think we should just call it in `Send`, and then pass that data (right now this is `jsonData`) to the `sign()` function as a `[]byte`. This way, we only marshal once for the webhook, and it makes the `sign` function a bit more focused, since it's really there to sign bytes.

GildedHonour:
I've implemented webhook for "campaing created and completed". Not tested yet

controllers/api/webhook.go
GildedHonour: Done. I've not implemented `WithWebhook` yet

GildedHonour:
@jordan-wright please check out the approach I've used for api/campaign. If all is good, I'll continue using it for the other types of events.

jordan-wright:
I don't think this is the design that I'd go for. Ideally, the events your receive in your webhook would be the exact events that are stored in the database. This way, you're seeing the same information just getting it pushed to you instead of having to poll. In this case, we're creating an entirely different kind of payload.

The problem is that I don't have any easy solution. Going back to the conversation we had a couple of days ago I'd love it if it were as simple as calling something like `webhook.Send(e)` in `campaign.AddEvent`. That would make sure that we captured all events and sent out the exact events that are stored to the database.

The downside to this approach is that right now we call `AddEvent` from our `models/` package - in `campaign.go` and `result.go`. In this case, we don't have our `webhook.Sender` to use.

Thinking more about it, while it isn't the easiest solution, here's what I'd do:

* In the `Handle*` functions in `result.go`, we called `r.createEvent`. We could change this to `r.CreateEvent` (to make it an exported function) and move the calls to it outside of the `Handle*` functions and into the `phish.go` places where those functions are called. This would let us keep those events and send them via the webhooks.
* In `campaign.go`, the only place we call `AddEvent` is in campaign creation. We should add an event for campaign completion, but I'd view that out of scope for this PR - I can add that later. For now, we could move that call out to `api/campaigns.go` where we create the campaign and do the same thing- keep the event and use it when sending the webhook.

This changes where events are added right now, but it's far more important to me to have webhook events mirror what's in the database than it is to have the `AddEvent` calls done from the `models/` package (since it's exported anyway 😛)

As for the servers, this would turn the sending function into something like this (untested):

```go
func (as *Server) sendWebhooks(event *models.Event) {
  whs, err := models.GetActiveWebhooks()
  if err != nil {
    log.Errorf("error getting webhooks: %v", err)
    return
  }
  for _, wh := range whs {
    go func(wh models.Webhook) {
        as.webhook.Send(wh.Url, wh.Secret, event)
    }(wh)
  }
}
```

I know this is a bit of a departure from where things are now, but I think it will result in a stronger, more consistent design. 

I really appreciate you continuing the great work and for being so patient with us as we work through the kinks. I wish things were easier (you can blame past @jordan-wright for not thinking about these things earlier! 😄) but I'm excited about where we'll end up.

GildedHonour:
> The problem is that I don't have any easy solution. Going back to the conversation we had a couple of days ago I'd love it if it were as simple as calling something like webhook.Send(e) in campaign.AddEvent. That would make sure that we captured all events and sent out the exact events that are stored to the database.
> 
> The downside to this approach is that right now we call AddEvent from our models/ package - in campaign.go and result.go. In this case, we don't have our webhook.Sender to use.

Yes. But after that we decided, in case of `campaign`, :

1) not to "sendWebhooks" from within a model - https://github.com/gophish/gophish/pull/1642#discussion_r344470837

2) instead "sendWebhooks" from api/campaign - https://github.com/gophish/gophish/pull/1642#discussion_r344469030

What's an issue with this?

GildedHonour:
> The downside to this approach is that right now we call AddEvent from our models/ package - in campaign.go and result.go. In this case, we don't have our webhook.Sender to use.


Well, we introduced `webhook.Sender` in the first place. And we can get rid of it. 

Sending a webhook boils down to creating an instance of it and simply sending it. It doesn't necessitate to create and send it via `webhook.Sender`. We can create a helper method in webhook which creates a DefaultHttpClient and use that.

jordan-wright:
> Yes. But after that we decided, in case of `campaign`, :

I'm still of the opinion that sending the webhook from within the `models/` package probably isn't the right approach. However, I'm also very strongly in favor of making sure that the events we send are the exact events that wind up in the database.

This means there's a harder way which will probably wind up with a better architectural flow, or an easier way which will require touching less code.

## The Harder Way

My previous proposal aims to address both issues. Instead of trying to send the webhook in the `models/` package everywhere that `AddEvent` is called, I'm proposing to move the `AddEvent` (or the helper for results called `createEvent`) calls out of the `models/` package and into the `api/` package, or others as needed. If we do this, then after `AddEvent` is called, we'll have a handle to the event that was just stored in the db (since we pass it as a pointer). We can then take that event and send it out.

Here's what this would look like in practice:

* First, we change `models/result.go: createEvent` to `CreateEvent` (making it exported so we can use it from the `api/` package.
* Then, we take calls to `createEvent` (like [this one](https://github.com/gophish/gophish/blob/master/models/result.go#L99)) and move them into the `controllers` package where the `Handle*` equivalent is called (in this case, [here](https://github.com/gophish/gophish/blob/master/controllers/phish.go#L153))

This means the flow would now look like this in `phish.go`:

```go
event, err := r.CreateEvent(models.EventOpened, d)
if err != nil {
	log.Error(err)
	// Handle the error, returning an HTTP error
}

err = as.Webhook.Send(event)
if err != nil {
	log.Error(err)
	// Handle the error, returning an HTTP error
}

// For now, I'm passing "d", which is the EventDetails but realistically we won't need it anymore,
// since HandleEmail
err = rs.HandleEmailOpened(d)
if err != nil {
	log.Error(err)
}
```

The hard part here is that I haven't fully thought out how we would handle the Email* events, like HandleEmailSent, HandleEmailError, etc. since the calls to those are done from `models/maillog.go`, which is called from the `worker` package. To make this work, we could start by adding a webhook Sender to the [`DefaultWorker`](https://github.com/gophish/gophish/blob/master/worker/worker.go#L21). This could then be passed down to the `MailWorker` to be used as needed. It's not ideal, but keeps the architecture how we want it.

This leads to the easier way, which may not be as architecturally nice.

## The Eas(ier) Way

The easier way would be to actually call `webhook.Send()` from the models package. I know, it's not ideal, but I think we can make it work for now.

Specifically, we would make a sender singleton (like the `DefaultSender` you had mentioned, which is also a pattern used in the [`net/http`](https://golang.org/src/net/http/client.go#L107) package), then call `models.SendWebhooks()` (which would contain the code from the above comment at https://github.com/gophish/gophish/pull/1642#issuecomment-553203858) from the models package anywhere where `CreateEvent` is called. This would call `webhook.Send` which would subsequently use the `DefaultSender`.

We should be ok when it comes to testing, since there won't be any active webhooks by default. This will let us test with/without webhooks as needed.

Sure, this isn't _ideal_ architecturally but at this point it's more important to me that the events that are sent are the events in the db. If we need to make some architectural tradeoffs for the sake of having this happen (without having to change a ton of code!) then I'm ok with it 😄 

I know this is a lot, so let me know if you have any questions!

GildedHonour:
I've implemented it via the easy path, for now. For the campaign model.

In the "models" package we import "webhook" package and call "sender.sendWebhooks". Because of this, we no longer can refer to "models" from within the "webhook" package. Otherwise, it'd cause a circle dependency error and won't compile. 

Therefore:


      //models/campaign.go

      //[..............]
      
      pl := map[string]string {
          "event_name": "create_campaign",
          "data": "TODO",
      }



      //we can't move this code to the 'webhook' package

      whs, err := GetActiveWebhooks()
      if err != nil {
        whsInfo := make(map[string]string)
        for _, wh := range whs {
           whsInfo[wh.Url] = wh.Secret
        }
        webhook.SendAll(whsInfo, pl)
      } else {
        log.Error("GetActiveWebhooks")
      }






This this of code would have to be repeated in each file of the "models" package -- each file/model we want to send webhooks for. Is there any solution? I can create a helper method in each "models" file, yet there'll be repetition still.

jordan-wright:
I think you're on the right track. You've added the logic to `campaign.AddEvent` which I think is perfectly reasonable so that the logic only needs to be in one place.

In terms of how we can make this work without circular dependency issues, I'd say we should have the logic look something like this in the `AddEvent` function:

```go
whs, err := GetActiveWebhooks()
if err != nil {
	log.Error("GetActiveWebhooks")
} else {
	// Maybe we can think of a better name? Receiver? Destination?
	whEvents := []webhook.Configuration{}
	for _, wh := range whs {
		whEvents = append(whEvents, webhook.Configuration{
			URL: wh.URL,
			Secret: wh.Secret,
		})
	}
	webhook.SendAll(whEvents, e)
}
```

Then, in the `webhook` package, we would update the signature for `SendAll` to look like this:

```go
func SendAll(whEvents []Configuration, data interface{}) {...}
```

We would also need to create the right `webhook.Configuration` struct containing the `URL` and `Secret` fields. I'm open to better names if we come up with one 😄 

That _should_ handle the logic side of things. I'll leave code-comments for the rest of the review.

Keep up the great work, and please continue to pass over questions as you have them!

webhook/webhook.go
```diff
@@ -0,0 +1,104 @@
...
+  hash1 := hmac.New(sha256.New, []byte(secret))
+  _, err := hash1.Write(data)
+  if err != nil {
+    return "", err
+  }
+  hexStr := hex.EncodeToString(hash1.Sum(nil))
+  return hexStr, nil
+}
+
+func interfaceToBytes(data interface{}) ([]byte, error) {
```
jordan-wright: We should be able to remove this function since we don't need it anymore.

webhook/webhook.go
```diff
@@ -0,0 +1,104 @@
...
+
+type defaultSender struct {
+  client *http.Client
+} 
+
+var senderInstance Sender
+func newDefaultSender() Sender {
+  sn := &defaultSender{}
+  sn.client = &http.Client{
+    Timeout: DefaultTimeoutSeconds,
```
jordan-wright: Another thing we might consider is adding a configuration which refuses to follow redirects, though I could see this being a stretch goal for later.

webhook/webhook.go
```diff
@@ -0,0 +1,104 @@
...
+
+type Sender interface {
+  Send(url string, secret string, data interface{}) error
+}
+
+type defaultSender struct {
+  client *http.Client
+} 
+
+var senderInstance Sender
```
jordan-wright: Instead of the pattern where you check if `senderInstance` is `nil` then call `newDefaultSender`, you might instead just instantiate it here. You can see an example in the [`net/http` package](https://golang.org/src/net/http/client.go#L107). This would look something like:

```
var senderInstance = &defaultSender{
    client: &http.Client{
        Timeout: DefaultTimeoutSeconds,
    }
}
```

Then you can remove `newDefaultSender` and all the logic in the functions below.

jordan-wright (commented):
Here is just some more feedback from skimming the latest code.

webhook/webhook.go
GildedHonour: We need it

Captura de pantalla 2019-11-18 a la(s) 11 46 54

webhook/webhook.go
GildedHonour: Noted

webhook/webhook.go
GildedHonour: Done

webhook/webhook.go
jordan-wright: Why aren’t we just signing `jsonData`?

models/campaign.go
```diff
@@ -608,5 +629,24 @@ func CompleteCampaign(id int64, uid int64) error {
 	if err != nil {
 		log.Error(err)
 	}
+
+  //INFO send webhook
+  pl := map[string]interface{} {
```
jordan-wright: I’d hold off on sending campaign completion events for now. The proper solution would be to run “AddEvent” when a campaign is completed, though I’d see that out of scope for this PR- it’s something I can fix shortly after this is merged, then our webhook via AddEvent will fire.

models/campaign.go
```diff
@@ -157,6 +158,26 @@ func (c *Campaign) UpdateStatus(s string) error {
 func (c *Campaign) AddEvent(e *Event) error {
 	e.CampaignId = c.Id
 	e.Time = time.Now().UTC()
+
+  //INFO send webhook
+  pl := map[string]interface{} {
```
jordan-wright: Instead of making a new map, I’d still strongly recommend just sending the event itself to SendAll. That way, the event sent to the webhook is the same one returned by the API and stored in the DB.

jordan-wright: Also note that since we’re putting the call here, you won’t need to call the webhook anywhere else - all events are meant to go through here. :)

webhook/webhook.go
GildedHonour: Because the Go library function `hmac.Write()` requires `[]byte`

jordan-wright: I’m not at my desk so I can’t confirm that I’m not missing something, but that should be what jsonData is, since [thats what json.Marshal returns](https://golang.org/pkg/encoding/json/#Marshal).

GildedHonour: Right. Fixed

models/campaign.go
GildedHonour: * sending an event - done

* can you elaborate on "you won’t need to call the webhook anywhere else - all events are meant to go through here. :)" ?

models/campaign.go
GildedHonour: removed it

GildedHonour:
In the "models" there's no validation of the user's input. We don't do it in the app, right?

GildedHonour:
Found that yes.

GildedHonour:
The basic functionality has been done. What's left are: 

* displaying the status of a webhook on UI in red or green colours. 

* reloading a page or a row or cell after a ping has returned and if a webhook changes its status.

webhook/webhook.go
```diff
@@ -0,0 +1,97 @@
...
+  client: &http.Client{
+    Timeout: time.Second * DefaultTimeoutSeconds,
+    CheckRedirect: func(req *http.Request, via []*http.Request) error {
+      return http.ErrUseLastResponse
+    },
+  },
+}
+
+type EndPoint struct {
+  Url string
```
jordan-wright: Minor nit - can we please make this `URL` to match `models/campaign.go`?

models/result.go
```diff
@@ -65,6 +66,7 @@ func (r *Result) HandleEmailSent() error {
 	r.SendDate = event.Time
 	r.Status = EventSent
 	r.ModifiedDate = event.Time
+	sendWebhooks(event)
```
jordan-wright: You should be able to remove all of these, since each `Handle*` function calls `r.createEvent` which [calls `campaign.AddEvent`](https://github.com/gophish/gophish/blob/52ed2eabe843033a03281a51b252c86d21092b57/models/result.go#L55). Since we're already sending the webhook in `AddEvent`, we'd wind up sending it twice as things are now.

models/campaign.go
```diff
@@ -157,6 +158,26 @@ func (c *Campaign) UpdateStatus(s string) error {
...
+  whs, err := GetActiveWebhooks()
+  if err != nil {
+		whEndPoints := []webhook.EndPoint{}
+	  for _, wh := range whs {
+			whEndPoints = append(whEndPoints, webhook.EndPoint {
+			  Url: wh.Url,
+			  Secret: wh.Secret,
+			})
+	  }
+	  pl := map[string]interface{} {
```
jordan-wright: As mentioned before, let's please just send the event as is. This is still wrapping it in a struct, which isn't necessary. In this case, we should just run `webhook.SendAll(whEndpoints, e)`. This will subsequently marshal the event into JSON and send it out, so what's received by the webhook would essentially be the same thing that's returned via the API.

models/result.go
```diff
@@ -212,3 +220,23 @@ func GetResult(rid string) (Result, error) {
 	err := db.Where("r_id=?", rid).First(&r).Error
 	return r, err
 }
+
+//INFO send 'Event' to all the active webhooks
```
jordan-wright: Once we remove the unnecessary calls to `sendWebhooks`, we should be able to remove this function entirely.

GildedHonour: done

models/campaign.go
GildedHonour: done

models/result.go
GildedHonour: done

webhook/webhook.go
GildedHonour: renamed to URL

GildedHonour:
I'd like to merge the PR asap so that we'll close this task with Al-TSN. 

@jordan-wright what's left to implement? Tests, some polishing of the code, something on UI perhaps.

And functionality-wise?

config.json
```diff
@@ -1,7 +1,7 @@
 {
 	"admin_server": {
 		"listen_url": "127.0.0.1:3333",
-		"use_tls": true,
+		"use_tls": false,
```
jordan-wright: We'll want to reset this to `true` before merging.

controllers/api/campaign.go
```diff
@@ -41,8 +41,9 @@ func (as *Server) Campaigns(w http.ResponseWriter, r *http.Request) {
 		if c.Status == models.CampaignInProgress {
 			go as.worker.LaunchCampaign(c)
 		}
-		JSONResponse(w, c, http.StatusCreated)
-	}
+    // as.sendWebhooks("create_campaign", c)
```
jordan-wright: You can remove this line and re-run `goimports`.

controllers/api/server.go
```diff
@@ -6,6 +6,7 @@ import (
 	mid "github.com/gophish/gophish/middleware"
 	"github.com/gophish/gophish/models"
 	"github.com/gophish/gophish/worker"
+	// "github.com/gophish/gophish/webhook"
```
jordan-wright: You can remove this line

controllers/api/webhook.go
```diff
@@ -0,0 +1,117 @@
+package api
+
+//TODO
```
jordan-wright: You can remove this line

controllers/api/webhook.go
```diff
@@ -0,0 +1,117 @@
...
+    err = models.PutWebhook(&wh2)
+    if err != nil {
+      JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusBadRequest)
+      return
+    }
+    JSONResponse(w, wh2, http.StatusOK)
+  }
+}
+
+func (as *Server) PingWebhook(w http.ResponseWriter, r *http.Request) {
```
jordan-wright: If we're calling the route `/validate` (which I think is the right option), I'd like for us to be consistent here and rename this to `ValidateWebhook`

controllers/api/webhook.go
```diff
@@ -0,0 +1,117 @@
...
+    vars := mux.Vars(r)
+    id, _ := strconv.ParseInt(vars["id"], 0, 64)
+    wh, err := models.GetWebhook(id)
+    if err != nil {
+      log.Error(err)
+      JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusInternalServerError)
+      return
+    }
+
+    //TODO update it here inplace?
```
jordan-wright: Possible dangling comment? Looks like we do update the `IsActive` property if the response checks out.

db/db_sqlite3/migrations/20191104103306_0.7.1_create_webhooks.sql
```diff
@@ -0,0 +1,14 @@
+
```
jordan-wright: We'll want to rename this to `0.9.0` instead of `0.7.1`

controllers/api/webhook.go
```diff
@@ -0,0 +1,117 @@
...
+  "net/http"
+  "strconv"
+  "encoding/json"
+
+  "github.com/gophish/gophish/models"
+  "github.com/gorilla/mux"
+  log "github.com/gophish/gophish/logger"
+  "github.com/gophish/gophish/webhook"
+)
+
```
jordan-wright: Please add a docstring here in the format `Webhooks ...`, such as `Webhooks handles the CRUD operations for managing Gophish webhooks` or similar.

controllers/api/webhook.go
```diff
@@ -0,0 +1,117 @@
...
+    }
+    err = models.PostWebhook(&wh)
+    if err != nil {
+      JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusBadRequest)
+      return
+    }
+    JSONResponse(w, wh, http.StatusCreated)
+  }
+}
+
```
jordan-wright: Please add a docstring here in the format `Webhook ...`, such as `Webhook handles the CRUD operations for managing existing individual Gophish webhooks` or similar.

models/campaign.go
jordan-wright: This doesn't appear done - it looks like you're still sending `pl` to `SendAll`.

models/campaign.go
```diff
@@ -157,6 +158,26 @@ func (c *Campaign) UpdateStatus(s string) error {
...
+			  URL: wh.URL,
+			  Secret: wh.Secret,
+			})
+	  }
+	  pl := map[string]interface{} {
+	    "data": e,
+	  }
+	  webhook.SendAll(whEndPoints, pl)
+	} else {
+		log.Error("GetActiveWebhooks")
```
jordan-wright: Please make this a more descriptive error message

models/webhook.go
```diff
@@ -0,0 +1,68 @@
+package models
+
+import (
+  "errors"
+
+  log "github.com/gophish/gophish/logger"
+)
+
```
jordan-wright: Please add comments for all exported functions and variables in this file. Running `golint` will help you out here.

templates/base.html
```diff
@@ -65,7 +65,10 @@
     {{template "body" .}}
     <!-- Placed at the end of the document so the pages load faster -->
     <script src="/js/dist/vendor.min.js"></script>
-    <script src="/js/dist/app/gophish.min.js"></script>
+
+    <!-- TODO - use "min" -->
```
jordan-wright: Before merging, we should change the reference to `min.js`, but please don't actually commit your copy of `.min.js`- I'll compile that myself and commit it after merging these changes.

templates/nav.html
```diff
@@ -28,6 +28,11 @@
                     <a href="/users">User Management<span class="nav-badge badge pull-right">Admin</span></a>
                 </li>
                 {{end}}
+                {{if .ModifySystem}}
```
jordan-wright: Since "User Management` is already wrapped in an `if .ModifySystem`, we can remove this and just move the `li` into the block above. So it'd look like:

```
{{if .ModifySystem}}
<li>
    <a href="/users">User Management<span class="nav-badge badge pull-right">Admin</span></a>
</li>
<li>
    <a href="/webhooks">Webhooks<span class="nav-badge badge pull-right">Admin</span></a>
</li>
{{end}}
```

webhook/doc.go
```diff
@@ -0,0 +1,32 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+// Package worker contains the functionality for the background worker process.
```
jordan-wright: We'll want to update this comment

webhook/webhook.go
```diff
@@ -0,0 +1,96 @@
...
+  "errors"
+  "encoding/json"
+  "bytes"
+  "time"
+
+  log "github.com/gophish/gophish/logger"
+)
+
+const (
+  DefaultTimeoutSeconds = 10
```
jordan-wright: Please add comments for all exported functions and variables in this file. Running `golint` can help you out here.

webhook/webhook.go
```diff
@@ -0,0 +1,96 @@
...
+    },
+  },
+}
+
+type EndPoint struct {
+  URL string
+  Secret string
+}
+
+func SendSingle(endPoint EndPoint, data interface{}) error {
```
jordan-wright: Personally, if we're calling `senderInstance.Send` here, I'd just rename this function to `Send` instead of `SendSingle`. I think it's cleaner.

jordan-wright (changes requested):
Hey @GildedHonour,

I'll do some functional testing tonight or tomorrow, but things are looking good. At this point, I think the bulk of what's left is cleanup work. Things like:

* Adding comments to all exported functions and variables (`golint` helps a ton with this!)
* Removing dangling comments and `TODO`'s
* Running `goimports` on every file
* Adding tests
* Removing unneeded files, like `go.sum`, `go.mod`, and `.DS_Store`

But I think we're largely in the home stretch here. Let me know if you have any questions, and keep up the good work!

webhook/webhook.go
GildedHonour: In that case we'll have 2 methods `Send` with almost the same signature, and interface with `Send` too. Do you think it'll be a good idea? 

I've called it `SendSingle` to distinguish it from `SendAll` and keep all that logical.

On the other hand, do we need to keep the method `Send` public  at all which now implements the interface `Sender`? If not, I'd rename it to `sendImpl`

webhook/webhook.go
GildedHonour: made them private

webhook/doc.go
GildedHonour: How should it be updated?

controllers/api/webhook.go
GildedHonour: in "models" we have method "Validate" for each model that make validatation on the fields of a model, in "models/webhook.go" it exist also and does the same thing. 

Therefore, to remove possible confusion, I've renamed `PingWebhook` to `ValidateEndPoitn`.  Right?

webhook/webhook.go
GildedHonour: Correction: `func (ds defaultSender) Send(endPoint EndPoint, data interface{}) error {` doesn't even implement the interface `Sender`. We may want to remove the interface then. However, I said that the interface `Send` would be helpful when creating the tests for webhooks, right?

jordan-wright: It's quite ok to have both a package level `Send` and a default implementation level `Send`. If anything, I think it might help avoid ambiguity.

We can use [`net/http`](https://golang.org/pkg/net/http/) as our inspiration here- it has both a package level `Get`, `Post`, etc. that use the default client, but it also has individual client `Get`, `Post`, etc. This makes it clear to the user that you _can_ make your own client and use those methods, otherwise you can use the top-level methods which use the default client. We should follow the same pattern here.

webhook/webhook.go
jordan-wright: Let's keep them exported. My suggestion was that we need to add comments for them so that the documentation renders them correctly.

webhook/doc.go
jordan-wright: Right now, it contains documentation for the `worker` package, while this is meant to be a docstring for the `webhook` package.

controllers/api/webhook.go
jordan-wright: I guess I'm not super firm either way. To me, it's just fine calling something `Validate` in the `controllers` package and calling something `Validate` in the `models` package, even when they do different things since the packages themselves serve different purposes. This also has the benefit of making it easier to reason about the URL path -> HTTP Handler.

GildedHonour:
I fixed all that yesterday. I might've missed something, will double check. Now I'm working on the tests.

jordan-wright:
It seems that there's still a bunch of documentation missing. Every exported function needs to have a comment above it in the form `// FunctionName ...` - `golint` can help you catch these. Additionally, it looks like the `.DS_Store`, `go.sum`, `go.mod`, etc. files are still included in the diff. We'll want to remove those.

Looking forward to seeing the tests!

controllers/api/campaign.go
```diff
@@ -127,7 +127,8 @@ func (as *Server) CampaignComplete(w http.ResponseWriter, r *http.Request) {
 	id, _ := strconv.ParseInt(vars["id"], 0, 64)
 	switch {
 	case r.Method == "GET":
-		err := models.CompleteCampaign(id, ctx.Get(r, "user_id").(int64))
+		userId := ctx.Get(r, "user_id").(int64)
```
jordan-wright: These changes aren't needed for this functionality - can we please revert them to keep this PR as focused as possible?

controllers/route.go
```diff
@@ -234,6 +235,13 @@ func (as *AdminServer) UserManagement(w http.ResponseWriter, r *http.Request) {
 	getTemplate(w, "users").ExecuteTemplate(w, "base", params)
 }
 
+// Webhook is an admin-only handler that handles webhooks
```
jordan-wright: In this case, your comment starts with `Webhook ...`, while it should be `Webhooks` - the name of the function. `golint` can help you track down these errors.

GildedHonour: fixed

GildedHonour:
Ok, adding more of the documentation and tests.

`golint` suggests that `Id` should be `ID`, whereas accross the project we have `Id` everywhere. That is, I'll keep it as `Id`, right?

jordan-wright:
For now, keep it “Id”. It’s on my todo list to do a refactor to change it all to “ID”.

GildedHonour:
I've added a test for `Send()` with a mock sender. In the project there's we have `mailer` which is similar to `webhook` in terms of testing and creating mocks. But not precisely.

GildedHonour:
Is that the way you meant when we were talking about an interface with `Send()` methond which would be used for testing?

webhook/webhook.go
```diff
@@ -0,0 +1,102 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	log "github.com/gophish/gophish/logger"
+)
+
+const (
+
+	//DefaultTimeoutSeconds is amount of seconds of timeout used by HTTP sender
```
jordan-wright: Minor nit - all of these should be `// Comment`, not `//Comment` (note the space). I understand there are a couple of instances in the codebase where I make the same mistake, but best to fix it now rather than having to go back and fix it later.

jordan-wright:
After looking into it a bit more, I think testing this might be tricky. While we should keep the interface, we don't currently have a way to use it. I can't quite remember _why_ we don't make it a configurable attribute on the server anymore (I seem to recall a circular dependency, but 🤷‍♂), but that's something we can fix down the line.

For now, we probably don't even need the mock sender. The best we can probably do is create an instance of [`httptest.NewServer`](https://golang.org/pkg/net/http/httptest/#NewServer), passing it a handler that does whatever validation on the received data we care about (such as verifying the signature, verifying the number of events received, etc.).

You can find an example on how to use the `httptest.Server` [here](https://play.golang.org/p/GrxhEq5YDU8).

GildedHonour:
Okay. But

> passing it a handler that does whatever validation on the received data we care about (such as verifying the signature, verifying the number of events received, etc.).

The thing is, the validation that take place a remote server is a concern of that server. Even if a server does it improperly, skips it, etc... - Gophish shouldn't care, as long as it sends data to that server in a correct way.  Right?

webhook/webhook.go
GildedHonour: fixed

jordan-wright:
> Gophish shouldn't care, as long as it sends data to that server in a correct way. Right?

With these tests, the first goal is to confirm that the signature we're sending is the one we expect to send. Basically writing how validation is expected to be done and confirming it works. Yes, Gophish itself doesn't do this in production but this serves as a sanity-check to ensure that both now and in the future, we're generating valid signatures when it's expected.

For the number of expected events, this is basically saying that we should validate what payload we expect to receive on the wire. Like signatures, this gives us confidence that the code does what's advertised now and in the future.

Finally, there's another benefit to showing an example of validation - it gives us a place to point people when they ask how they can do it themselves.

GildedHonour:
Updated.

I created 2 tests for "Send" that do testing for sending http request in a slighly different way.

Also, on one hand, the method "sign" is private because it doesn't need to be public because it's called only from "Send" function internally. One the other hand, it can't be tested in a simple way because it's private.

webhook/webhook_test.go
```diff
@@ -0,0 +1,189 @@
...
+		assert.Equal(s.T(), contTypeJsonHeader, "application/json")
+
+		body, err := ioutil.ReadAll(r.Body)
+		s.Nil(err)
+
+		var d2 map[string]interface{}
+		err = json.Unmarshal(body, &d2)
+		s.Nil(err)
+
+		key1RealVal := d2["key1"]
```
jordan-wright: I'm not familiar enough with the details of `assert` to know if you could just do `assert.Equal(s.T(), d1, d2` but it might be worth a shot.

webhook/webhook_test.go
```diff
@@ -0,0 +1,189 @@
...
+	d1 := map[string]string {
+		"a1": "a11",
+		"a2": "a22",
+		"a3": "a33",
+	}
+	err := mcSnd.Send(endp1, d1)
+	s.Nil(err)
+}
+
+func (s *WebhookSuite) TestSendReal() {
```
jordan-wright: What are you actually testing here? This is essentially copy/pasting all the code from the signing/sending routines into a test. It's better to call those functions directly.

webhook/webhook_test.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	hClient := &http.Client{}
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Println("[test] running the server...")
+
+		realSign := r.Header.Get(webhook.SignatureHeader)
+		assert.Equal(s.T(), expectedSign, realSign)
+
+		neHeader := r.Header.Get("non-existing-header")
```
jordan-wright: You don't need this specific condition.

jordan-wright (commented):
Hey there!

So I think you can remove `TestSendReal` altogether, since it's not really testing any part of the `webhook` package from what I can tell. Instead, it's just repeating the code, which won't help us catch issues early when the functionality inevitably changes over time.

> Also, on one hand, the method "sign" is private because it doesn't need to be public because it's called only from "Send" function internally. One the other hand, it can't be tested in a simple way because it's private.

Sure you can! 😄 As long as your tests are in the same package (they are) then you have access to private functions. That's one of the biggest reasons for keeping tests in the same package vs. moving them to a `package_test` kind of package. So testing `sign` directly is what I'd do.

And in these tests, you hardcode the signature. This is ok when you want to test, for example, that a particular header is set. However when testing the signature function directly, I'd use it as an opportunity to show the reader _how_ a valid signature is derived. This could look (roughly) like this:

```go
func TestSignatureGeneration() {
    secret := "12345"
    payload := []byte("This is a sample payload")
    mac := hmac.New(sha256.New, secret)
    mac.Write(payload)
    expected := mac.Sum(nil)
    
    got, err := sign(secret, payload)
    s.Nil(err)
    assert.True(s.T(), hmac.Equal(expected, got))
}
```

This both validates that `sign` is giving us a correct signature, but also has the benefit of giving us a place to point people when they ask how to validate signatures in their application.

webhook/webhook_test.go
GildedHonour: Actually, I have an error that there's no access to the "sign" method; looking into it.

GildedHonour: fixed

GildedHonour:
Updated the tests

GildedHonour:
Removed the redundant files.

GildedHonour:
@jordan-wright have you been able to check out the tests and the rest of the code?

jordan-wright:
No, I likely won’t be able to until the end of this week.

static/js/src/app/gophish.js
```diff
@@ -223,6 +223,28 @@ var api = {
...
+        get: function() {
+            return query("/webhooks/", "GET", {}, false)
+        },
+        post: function(webhook) {
+            return query("/webhooks/", "POST", webhook, false)
+        },
+    },
+    webhookId: {
+        get: function(id) {
+            return query(`/webhooks/${id}`, "GET", {}, false)
```
jordan-wright: This particular file isn't currently built with Babel compilation, so we shouldn't use ES6 directives here. If you could, please change these to use the same double-quote string styles as the other endpoints.

The thing to check is whether or not running `gulp` works as expected.

models/webhook.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"errors"
+
+	log "github.com/gophish/gophish/logger"
+)
+
+// Webhook represents the webhook model
+type Webhook struct {
+	Id       int64  `json:"id" gorm:"column:id; primary_key:yes"`
+	Title    string `json:"title"`
```
jordan-wright: It might be a small thing, but can we call this `Name` instead of `Title`? This would have it match the rest of the models in Gophish. I only noticed it when I was trying to test it out and it was a bit jarring moving from "name" to "title".

templates/webhooks.html
```diff
@@ -0,0 +1,72 @@
...
+            <div class="modal-body" id="modal_body">
+                <div class="row" id="modal.flashes"></div>
+                <label class="control-label" for="title">Title:</label>
+                <div class="form-group">
+                    <input type="text" class="form-control" placeholder="Title" id="title" required autofocus />
+                </div>
+
+                <label class="control-label" for="url">Url:</label>
+                <div class="form-group">
+                    <input type="text" class="form-control" placeholder="Url" id="url" required />
```
jordan-wright: Can we change this placeholder to `http://example.com/webhook` or something to indicate what the field should look like?

db/db_sqlite3/migrations/20191104103306_0.9.0_create_webhooks.sql
```diff
@@ -0,0 +1,14 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+CREATE TABLE IF NOT EXISTS "webhooks" (
+    "id" integer primary key autoincrement,
+    "title" varchar(255),
+    "url" varchar(1000),
+    "secret" varchar(255)
```
jordan-wright: You're missing the `is_active` column in this migration.

GildedHonour: I'm not, you're looking at the old code.

jordan-wright: Also, why are you using an integer field for `is_active`? We use boolean fields in migrations now and this seems like another good use case for that.

templates/webhooks.html
GildedHonour: fixed

models/webhook.go
GildedHonour: fixed

static/js/src/app/gophish.js
GildedHonour: will change

db/db_sqlite3/migrations/20191104103306_0.9.0_create_webhooks.sql
GildedHonour: is_active is boolean now

static/js/src/app/gophish.js
GildedHonour: fixed. 

I need to install gulp to check that

GildedHonour: running `gulp` has worked well.

GildedHonour:
```
$ gulp
[10:04:42] Using gulpfile ~/projects/go_lang/gophish/gulpfile.js
[10:04:42] Starting 'default'...
[10:04:42] Starting 'vendorjs'...
[10:04:42] Starting 'scripts'...
[10:04:42] Starting 'styles'...
[10:04:44] Finished 'scripts' after 1.32 s
[10:04:53] Finished 'styles' after 10 s
[10:04:53] Finished 'vendorjs' after 10 s
[10:04:53] Finished 'default' after 10 s
```

db/db_sqlite3/migrations/20191104103306_0.9.0_create_webhooks.sql
GildedHonour: I see. There're 2 migrations, in one of them it's missing

jordan-wright: This is the latest code right? Like, the code right here that we're commenting on is for the SQLite migrations, and it doesn't have `is_active`. MySQL does, but SQLite doesn't.

GildedHonour:
```
SQLite does not have a separate Boolean storage class. Instead, Boolean values are stored as integers 0 (false) and 1 (true).
```

https://www.sqlite.org/datatype3.html



Sqlite doesn't support boolean, why are we using them in all the migration for Sqlite?

jordan-wright: It doesn't have a separate concept for booleans, but does support a `boolean` data type- it just represents them internally as integers:

```
sqlite> .schema webhooks
CREATE TABLE IF NOT EXISTS "webhooks" (
    "id" integer primary key autoincrement,
    "title" varchar(255),
    "url" varchar(1000),
    "secret" varchar(255),
    "is_active" boolean
);
sqlite> select * from webhooks;
1|Test Webhook|http://127.0.0.1:9999/webhook|mysecret|0
```

I use them both for clearer database structure, since it's clear that this is a true/false kind of field rather than an integer, as well as to try and avoid any possible weird issues with the underlying ORM, gorm.

GildedHonour: Fixed

config.json
jordan-wright: Also just a reminder that we'll want to reset this to `true` before merging.

GildedHonour: fixed

webhook/webhook.go
```diff
@@ -0,0 +1,102 @@
...
+// Send contains the implementation of sending webhook to an EndPoint
+func (ds defaultSender) Send(endPoint EndPoint, data interface{}) error {
+	jsonData, err := json.Marshal(data)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+
+	req, err := http.NewRequest("POST", endPoint.URL, bytes.NewBuffer(jsonData))
+	signat, err := sign(endPoint.Secret, jsonData)
```
jordan-wright: After reviewing some of the common webhook server implementations, I think we should tweak this implementation slightly. Namely, we should add a prefix to the signature to indicate what hashing algorithm was used. This would make it like:

```
X-Gophish-Signature: sha256=hash
```

You can see an example of this [here](https://developer.github.com/webhooks/#example-delivery). The benefit this gives us is making it much easier to move to different hashing functions in the future as needed.

jordan-wright: For now, we can keep this simple - I'd implement this is to make a new constant, `Sha256Prefix = "sha256"`, and then adjust the signature that's set.

GildedHonour: fixe

GildedHonour:
Other fixes?

jordan-wright:
Doing a bit more testing, I think the `is_active` field should be set by the user. Right now, it appears to be set entirely from the Gophish end when the user attempts to send a ping to the endpoint.

Instead, I think the user should be able to enable/disable the webhook as needed. This could be a checkbox in the modal popup that sends the value in the JSON payload. Then, the validation doesn't need to set the value - only send the request and give back the result.

I have plans to do more testing tomorrow evening or so, but I think we're in the home stretch here! Maybe just a few very minor tweaks, then we'll be ready to merge.

GildedHonour:
Okay. Fixed

GildedHonour:
@jordan-wright what's left to fix?

static/js/src/app/webhooks.js
```diff
@@ -0,0 +1,182 @@
...
+        is_active: $("#is_active").is(":checked"),
+    };
+    if (id != -1) {
+        wh.id = id;
+        api.webhookId.put(wh)
+            .success(function(data) {
+                dismiss();
+                load();
+                $("#modal").modal("hide");
+                successFlash(`Webhook "${wh.name}" has been updated successfully!`);
```
jordan-wright: We'll want to escape this to prevent xss. You can use `escapeHtml` for this.

static/js/src/app/webhooks.js
```diff
@@ -0,0 +1,182 @@
...
+            .error(function(data) {
+                modalError(data.responseJSON.message)
+            })
+    } else {
+        api.webhooks.post(wh)
+            .success(function(data) {
+                load();
+                dismiss();
+                $("#modal").modal("hide");
+                successFlash(`Webhook "${wh.name}" has been created successfully!`);
```
jordan-wright: We'll want to escape this to prevent xss. You can use `escapeHtml` for this.

static/js/src/app/webhooks.js
```diff
@@ -0,0 +1,182 @@
...
+};
+
+const deleteWebhook = (id) => {
+    var wh = webhooks.find(x => x.id == id);
+    if (!wh) {
+        return;
+    }
+    Swal.fire({
+        title: "Are you sure?",
+        text: `This will delete the webhook '${wh.name}'`,
```
jordan-wright: We'll want to escape this to prevent xss. You can use `escapeHtml` for this.

static/js/src/app/webhooks.js
```diff
@@ -0,0 +1,182 @@
...
+    })
+};
+
+const pingUrl = (btn, whId) => {
+    dismiss();
+    btn.disabled = true;
+    api.webhookId.ping(whId)
+        .success(function(wh) {
+            btn.disabled = false;
+            successFlash(`Ping of "${wh.name}" webhook succeeded.`);
```
jordan-wright: We'll want to escape this to prevent xss. You can use `escapeHtml` for this.

static/js/src/app/webhooks.js
```diff
@@ -0,0 +1,182 @@
...
+        })
+        .error(function() {
+            btn.disabled = false;
+            var wh = webhooks.find(x => x.id == whId);
+            if (!wh) {
+                return
+            }
+            errorFlash(`Ping of "${wh.name}" webhook failed.`)
```
jordan-wright: If we're returning an error from the webhook ping, we should display it to the user here.

templates/webhooks.html
```diff
@@ -0,0 +1,77 @@
...
+                </div>
+
+                <label class="control-label" for="secret">Secret:</label>
+                <div class="form-group">
+                    <input type="text" class="form-control" placeholder="Secret" id="secret" required />
+                </div>
+
+                <label class="control-label" for="secret">Is active:</label>
+                <div class="form-group">
+                    <input type="checkbox" class="" id="is_active" value="true" />
```
jordan-wright: We have a way to make pretty checkboxes that I'd encourage you to use.  You can find an example [here](https://github.com/gophish/gophish/blob/master/templates/landing_pages.html#L65).

webhook/webhook_test.go
```diff
@@ -0,0 +1,119 @@
...
+		assert.Equal(s.T(), d1, d2)
+	}))
+
+	defer ts.Close()
+	endp1 := EndPoint{URL: ts.URL, Secret: secret}
+	err := Send(endp1, d1)
+	s.Nil(err)
+}
+
+func (s *WebhookSuite) TestSignature() {
```
jordan-wright: It's still not clear what this is testing. This doesn't appear to invoke any of the code in `webhook.go`.

jordan-wright (changes requested):
Hey there!

I've done another pass through the code itself and left some comments. We're definitely in the final stretch of things. Once we have these taken care of, I'll do a final pass using the webhooks locally and, if things are in good shape, we'll merge this in.

controllers/api/webhook.go
```diff
@@ -0,0 +1,96 @@
...
+		vars := mux.Vars(r)
+		id, _ := strconv.ParseInt(vars["id"], 0, 64)
+		wh, err := models.GetWebhook(id)
+		if err != nil {
+			log.Error(err)
+			JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusInternalServerError)
+			return
+		}
+		err = webhook.Send(webhook.EndPoint{URL: wh.URL, Secret: wh.Secret}, "")
+		if err == nil {
```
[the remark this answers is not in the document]
GildedHonour: How will flipping the conditionals clean up the code? We're expect the condition without an error first. And that's what I'm calling here first. And there have to exist 2 conditions here. 

```
		err = webhook.Send(webhook.EndPoint{URL: wh.URL, Secret: wh.Secret}, "")
		if err == nil {
			JSONResponse(w, wh, http.StatusOK)
		} else {
			JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusBadRequest)
		}
```


How do you say we could do this?

```
		err = webhook.Send(webhook.EndPoint{URL: wh.URL, Secret: wh.Secret}, "")
		if err != nil {
			JSONResponse(w, models.Response{Success: false, Message: err.Error()}, http.StatusBadRequest)
		} 
```

Return nothing in case of "no error"?

models/campaign.go
```diff
@@ -157,6 +158,21 @@ func (c *Campaign) UpdateStatus(s string) error {
 func (c *Campaign) AddEvent(e *Event) error {
 	e.CampaignId = c.Id
 	e.Time = time.Now().UTC()
+
+	whs, err := GetActiveWebhooks()
+	if err != nil {
```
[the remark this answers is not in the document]
GildedHonour: fixed

models/webhook.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// DeleteWebhook deletes an existing webhook in the database.
+// An error is returned if a webhook with the given id isn't found.
+func DeleteWebhook(id int64) error {
+	err := db.Where("id=?", id).Delete(&Webhook{}).Error
+	return err
+}
+
+func validate(wh *Webhook) error {
```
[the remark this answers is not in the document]
GildedHonour: fixed

static/js/src/app/webhooks.js
GildedHonour: Then why not escape it upon writting the database?

webhook/webhook_test.go
GildedHonour: removed

models/webhook.go
```diff
@@ -0,0 +1,80 @@
...
+	if wh.URL == "" {
+		return errors.New("URL can't be empty")
+	}
+	if wh.Name == "" {
+		return errors.New("Name can't be empty")
```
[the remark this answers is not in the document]
GildedHonour: fixed

models/webhook.go
```diff
@@ -0,0 +1,80 @@
...
+// DeleteWebhook deletes an existing webhook in the database.
+// An error is returned if a webhook with the given id isn't found.
+func DeleteWebhook(id int64) error {
+	err := db.Where("id=?", id).Delete(&Webhook{}).Error
+	return err
+}
+
+func validate(wh *Webhook) error {
+	if wh.URL == "" {
+		return errors.New("URL can't be empty")
```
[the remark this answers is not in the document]
GildedHonour: fixed

templates/webhooks.html
GildedHonour: fixed

controllers/api/webhook.go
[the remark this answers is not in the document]
jordan-wright: My recommendation is really just encouraging the pattern we use everywhere else in the code, which is also what Go itself recommends. That is, a pattern like:

```
if err != nil {
    handle the error
    return
}
continue execution
```

The reason this is encouraged is that it keeps the nesting to a minimum so we don't wind up with things like:

```
if err == nil {
    big block of code that happens if things go as planned
    if err == nil {
        another big block of code
        etc, etc
    } else {
        handle the error
    }
} else {
    handle the error
}
```

So in my proposal we're handling the error first, then if the error was nil, you'll find the `JSONResponse(w, wh, http.StatusOK)` at the end of the suggested code- this would be the default thing to run if no errors occurred. That's the way all of the other HTTP handlers work- this is the only one that has a different pattern when it doesn't need to.

static/js/src/app/webhooks.js
jordan-wright: Because then there would be a need to unescape any time you're using the data not from the UI, such as in the API.

Instead, the [recommended](https://security.stackexchange.com/questions/32394/when-to-escape-user-input/32396#32396) [pattern](https://stackoverflow.com/a/3648443) is to store the data in its original form in the database, then do contextual escaping as needed.

And note that while it's a bit of a manual process right now, my hope is to eventually move to something like React which automatically escapes values for us.

controllers/api/webhook.go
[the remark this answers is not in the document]
GildedHonour: Agree. Fixed

static/js/src/app/webhooks.js
GildedHonour: fixed

static/js/src/app/webhooks.js
GildedHonour: fixed

GildedHonour:
@jordan-wright what's left to fix?

jordan-wright:
I’ll let you know when I have a chance to review it. Should be soon.

jordan-wright:
I think we're at a point where I'm comfortable merging this in. There's some super minor tweaks left to do, but I can handle those.

Thanks for your patience and all your hard work! I don't have an ETA on the next release, but it will probably be in the next few weeks or so. I'm hoping that I can get #1612 and my work on password policy improvements merged, then I'll cut the new release.

Thanks again!

Issue #1624, opened 2019-10-21
API: Creating user should return initial API key


edermi:
What version of Gophish are you using?: 

0.8

Brief description of the issue: 

I'm building scripts to setup Gophish instances for multiple users.
When creating a user [via API](https://docs.getgophish.com/api-documentation/user-management), it is not possible to interact with the API using the created user. This makes it impossible to use Gophish without manual interaction (getting the API key once via Browser and web interface). 

What are you expecting to see happen? :

Since the user creating another user is in possession of the initial password, there would be no harm in just returning the API key for the created user in the JSON response. 

What are you seeing happen?

The only way to interact with the new user is to fetch the API key by browsing the web interface.

jordan-wright:
This seems reasonable, I bet we can make that happen.

PR #1663, opened 2019-11-16
1624 - Return API Key on Password Change

edermi:
Fix for #1624 

The idea is to create a second user type that allows to marshal the API key to JSON but is not serialized to the database.

In case a user is created or altered via API, the regular `models.User` object is converted to a `models.ChangedUser` before serialization, making the API key available to the issuer of the request.

Quick demo:

~~~bash
~  curl -k -XGET 'https://localhost:3333/api/users/' -H 'Authorization: b07c553b82b6aedfa4c7107c8f2b81854dc4d4c85e66c09fe77055a285c660d7'
[
  {
    "id": 1,
    "username": "admin",
    "role": {
      "slug": "admin",
      "name": "Admin",
      "description": "System administrator with full permissions"
    }
  }
]
~  curl -k -XPOST 'https://localhost:3333/api/users/' -H "Content-Type: application/json" -H 'Authorization: b07c553b82b6aedfa4c7107c8f2b81854dc4d4c85e66c09fe77055a285c660d7' -d '{"role": "admin", "username": "test", "password": "NotSoSecure123!"}'
{
  "id": 2,
  "username": "test",
  "api": "f0ad4dc056eb47041ea3293cb8bce2399fc81ec26e9a3a8cdc0c5f6aabfed140",
  "role": {
    "slug": "admin",
    "name": "Admin",
    "description": "System administrator with full permissions"
  }
}
~  curl -k -XGET 'https://localhost:3333/api/users/' -H 'Authorization: f0ad4dc056eb47041ea3293cb8bce2399fc81ec26e9a3a8cdc0c5f6aabfed140'
[
  {
    "id": 1,
    "username": "admin",
    "role": {
      "slug": "admin",
      "name": "Admin",
      "description": "System administrator with full permissions"
    }
  },
  {
    "id": 2,
    "username": "test",
    "role": {
      "slug": "admin",
      "name": "Admin",
      "description": "System administrator with full permissions"
    }
  }
]
~~~

I tried to keep the change as small and concise as possible, feedback is nevertheless appreciated.
I'd love to see this merged, after approval I'd update the API documentation as well to reflect the change.

### DIFF
--- a/models/user.go
+++ b/models/user.go
@@ -21,20 +21,6 @@ type User struct {
 	RoleID   int64  `json:"-"`
 }
 
-// Changed User is identical to a regular user except that it contains the API key in
-// the serialized response. This is desirable in case GoPhish is used via API with multiple
-// users, otherwise there are no means of retrieving the API key of a newly created user.
-// This type is only supposed to be used for JSON serialization, so there is no need for 
-// having database associations.
-type ChangedUser struct {
-	Id       int64  `json:"id"`
-	Username string `json:"username"`
-	Hash     string `json:"-"`
-	ApiKey   string `json:"api"` 
-	Role     Role   `json:"role"`
-	RoleID   int64  `json:"-"`
-}
-
 // GetUser returns the user that the given id corresponds to. If no user is found, an
 // error is thrown.
 func GetUser(id int64) (User, error) {
--- a/models/user.go
+++ b/models/user.go
@@ -16,7 +16,7 @@ type User struct {
 	Id       int64  `json:"id"`
 	Username string `json:"username" sql:"not null;unique"`
 	Hash     string `json:"-"`
-	ApiKey   string `json:"api" sql:"not null;unique"`
+	ApiKey   string `json:"api_key" sql:"not null;unique"`
 	Role     Role   `json:"role" gorm:"association_autoupdate:false;association_autocreate:false"`
 	RoleID   int64  `json:"-"`
 }
--- a/models/user.go
+++ b/models/user.go
@@ -16,7 +16,7 @@ type User struct {
 	Id       int64  `json:"id"`
 	Username string `json:"username" sql:"not null;unique"`
 	Hash     string `json:"-"`
-	ApiKey   string `json:"-" sql:"not null;unique"`
+	ApiKey   string `json:"api" sql:"not null;unique"`
 	Role     Role   `json:"role" gorm:"association_autoupdate:false;association_autocreate:false"`
 	RoleID   int64  `json:"-"`
 }
--- a/models/user.go
+++ b/models/user.go
@@ -21,6 +21,20 @@ type User struct {
 	RoleID   int64  `json:"-"`
 }
 
+// Changed User is identical to a regular user except that it contains the API key in
+// the serialized response. This is desirable in case GoPhish is used via API with multiple
+// users, otherwise there are no means of retrieving the API key of a newly created user.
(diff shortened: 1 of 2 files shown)


jordan-wright:
Hey @edermi!

Thanks so much for sending this in! I think this is a nice way to accomplish what we're going for. Great work!

Another option here would be to just return the API key in the JSON payload for the `User` model. I'd like to get your take on what we'd be preventing with your approach over the simpler one to make sure there isn't something I'm missing.

Thinking through threat models, it seems that:

* If a compromised admin wants a user's API key, they can just modify the user somehow and they'll get it.
* If a user's API key was compromised, the attacker has it anyway
* If a user's password was compromised, the attacker can retrieve the API key.

Is there any threats which your approach mitigates while just returning the API key would be impacted? I'm always down for the simplest solution possible, but I want to make sure I'm not missing an obvious attack vector here.

edermi:
Hi @jordan-wright,

thanks for already taking a look, I appreciate your feedback. In fact I assumed that any user has access to the `/api/users/` endpoint - that would be problematic. This, and the fact that I was not sure where else `User`s are marshaled and reflected back in responses made me go for the more defensive approach that does not break any previous assumptions or constraints unknown to me.

I checked and noticed that the API server takes care of validating the user's permissions, so I'm also fine with this approach if you prefer it. 

With your approach, the third curl request looks like this:

~~~bash
 ~  curl -k -XGET 'https://localhost:3333/api/users/' -H 'Authorization: f0ad4dc056eb47041ea3293cb8bce2399fc81ec26e9a3a8cdc0c5f6aabfed140'
[
  {
    "id": 1,
    "username": "admin",
    "api": "b07c553b82b6aedfa4c7107c8f2b81854dc4d4c85e66c09fe77055a285c660d7",
    "role": {
      "slug": "admin",
      "name": "Admin",
      "description": "System administrator with full permissions"
    }
  },
  {
    "id": 2,
    "username": "test",
    "api": "f0ad4dc056eb47041ea3293cb8bce2399fc81ec26e9a3a8cdc0c5f6aabfed140",
    "role": {
      "slug": "admin",
      "name": "Admin",
      "description": "System administrator with full permissions"
    }
  }
]
~~~

As you can see, administrators get access to all API keys. As you said, this also works with my approach but the administrator first needs to alter a user, so there is no difference in capabilties of an admin.

I will modify the pull request to your simpler approach if you prefer it over mine.

jordan-wright:
Totally understand (and appreciate) wanting to start as defensively as possible!

Since it looks like the simpler approach of just adjusting the JSON tag for that field keeps the same threat model in mind, let just go with that.

edermi:
Okay, I'm also fine with that. 
Thanks for your time and reviewing my changes.

models/user.go
```diff
@@ -16,7 +16,7 @@ type User struct {
 	Id       int64  `json:"id"`
 	Username string `json:"username" sql:"not null;unique"`
 	Hash     string `json:"-"`
-	ApiKey   string `json:"-" sql:"not null;unique"`
+	ApiKey   string `json:"api" sql:"not null;unique"`
```
jordan-wright: As a small nit, could we make this `api_key`?

jordan-wright:
Hey @edermi,

Thanks for changing up the PR! Super clean and easy - I dig it. I left a really minor request that if we could rename it to `api_key` vs just `api`, I'd appreciate it!

models/user.go
edermi: Of course!

jordan-wright:
Perfect, thank you!

Issue #1538, opened 2019-08-15
Password policy and login management


eickeh:
What version of Gophish are you using?: 0.7.1

Brief description of the issue: 
**Thanks again for your great work!**
We have several suggestions to improve the password policy and login management of GoPhish. In this thread we would like to discuss several ways to implement it. We are also open to provide some pull requests after some final decisions have been made.
If you're more comfortable with splitting this into several issues, I'll be happy to open them.

Here are our findings:
1) At the moment, there are no restrictions concerning the password complexity. I suggest a minimum length of 12 characters optionally also with enforced numbers.

2) To prevent session abuse one could enhance the functionality to log out a user automatically after some time of inactivity. At the moment, the standard value is at 5 days, which is a lot. A standard value could be 30 minutes. Moreover one could add an option to change this value or to deactivate it completely.

3) This is the most complicated point. At the moment, there is no protection against password brute forcing. To make such attacks more difficult, one could add progressive delays upon failed logins. 
An example implementation could do the following: After e.g. 3 failed login attempts for an user, one has to wait a minute to try again. If he/she fails 3 times again afterwards, he has to wait twice the time before trying again. This should have an upper bound (lets say 3 hours) to prevent that the account is practically locked forever.
With a view to the upcoming RBAC support, one could also think about locking (non admin) accounts and an option for the admin to re enable them.

jordan-wright:
Hi @eickeh,

I'd be fine accepting PR's for these, thanks for sending in the suggestions!

Let's shoot for the first two to start. I agree that we'll want to consider account lockout more carefully, and I think taking care of 1 and 2 will mitigate many of the threats that would require account lockouts.

For the password policy, I'd prefer to not be too prescriptive. I think just a reasonable length would be sufficient - 12 seems perhaps a bit much to me (don't get me wrong - I use a password manager so I wouldn't care, I'm just thinking about the average admin) but I'm open to hear more about your reasoning.

If we're on this subject, I'd also like to make it so the default password must be changed after the first login. If that's something you'd like to take on while you're in this space, it'd be appreciated! I'm just thinking of a flow that's something like:

* Add a new database schema that adds a boolean `password_change_required` field to users
* Set it to `true` for the initial admin
* On a successful login, if it's true, then set a session variable indicating a password reset needs to happen and redirect the admin to the change password page (I forget if we have one of these already)
* When the password is reset, update the database and remove that session flag
[...]

Issue #1213, opened 2018-09-27
Feature request: Altering random ID-link


CyberM070:
**Thanks for reaching out! We're happy to help resolve issues as quickly as possible.**

**Please use this template when creating a new issue. If you do not follow this template format, your issue may be closed without comment.**

Before filing a new issue, please use the search bar at the top of the browser to search for similar issues. Also, please make sure you have read our [documentation](https://getgophish.com/documentation/), which covers how to use Gophish.

What version of Gophish are you using?: 0.7.1

Brief description of the issue: Feature request

What are you expecting to see happen? : At this moment the generated random ID looks like this 
?rid=SN4rXjd or something similar. Is there anyway to be able to alter the random ID link, or add something to it? Let's say that I have a domain name: www.organizationX.com, and I would like to send a phishing mail which has a survey request as theme. I would like to show the landing page link as: www.organizationX.com/phishingsurvey followed by the random ID that is generated. This makes it more realistic and harder for employees to recognize. Another cool feature would be to use employee numbers as random ID since these are familiar for them, and more personalized. If you would like to do something similar at this moment you should host your website and create webpages et cetera, which could save a lot of time when imbedded in the tool.  

What are you seeing happen?

If this question is related to email templates or landing pages not working as expected, please provide your template or landing page below:

```html
Insert code here
```

Please provide any terminal output that may be relevant below:

```
Insert output here
```

Please provide as many steps as you can to reproduce the problem:
 * Step 1 - 
 * Step 2 -

rubberducky69:
You could use a subdomain instead, www.phishingsurvey.organizationx.com

CyberM070:
> You could use a subdomain instead, [www.phishingsurvey.organizationx.com](http://www.phishingsurvey.organizationx.com)

Yes but then you would have to setup webhosting for that domain which is not convenient at all times.

S0larflare:
You can already do this, just type what you want the link to be when you set up the campaign

Sent from my iPhone

On 28 Sep 2018, at 08:00, CyberM070 <notifications@github.com<mailto:notifications@github.com>> wrote:


You could use a subdomain instead, www.phishingsurvey.organizationx.com<http://www.phishingsurvey.organizationx.com>

Yes but then you would have to setup webhosting for that domain which is not convenient at all times.

jordan-wright:
Hi @CyberM070, thanks for reaching out!

[...]

Issue #346, opened 2016-08-22
allow to host the app in a subfolder


mattoufoutu:
Currently, it is only possible to deploy GoPhish at "/", with no support (at least none I found) for hosting in a subfolder.

Example: I have a server on which I have many web applications, all of them behind an nginx server. So I  tell nginx that "/gophish" and "/gophishadmin" should be proxy_pass'd respectively to the phishing site and to the admin interface. The problem here is that GoPhish has no way of knowing it is hosted under "/gophishadmin" instead of the usual "/", making every request to it fail because of erroneous URL paths.

It would be nice to be able to handle this with a "prefix" option (or whatever name you give it) for both the phishing site and the admin interface, so GoPhish would generate its links accordingly and would be able to work even if not hosted at a webserver's root.

jordan-wright:
Could something like [nginx url rewriting](https://serverfault.com/questions/379675/nginx-reverse-proxy-url-rewrite) help with this? I don't know that it is in the current plan to support hosting out of a configurable subdirectory.

mattoufoutu:
I'll have a look into nginx url rewriting thx.
On the other hand, if this isn't in your plans, would you still be willing
to accept a PR if I implement this by myself?

Le mer. 24 août 2016 06:23, Jordan Wright <notifications@github.com> a
écrit :

jordan-wright:
Sorry for the delay - I would absolutely accept a pr that addresses this problem in an easy to use way.

jordan-wright:
I'm going to go ahead and close this out for now, but if you open a PR, please reference this and we'll make sure it all gets taken care of.

Thanks again for this report!

PR #1681, opened 2019-11-27
Allow CORS to be configured for API requests

mdeboer:
Currently we're building an application that calls the Gophish API using XHR and ran into CORS. This PR adds proper support for CORS with preflight and I have added configuration options to customise the allowed origins as well as to toggle debug mode. CORS is only handled for the API endpoints and I have set the default to allow all origins.

It should probably be mentioned in the documentation that this should be set to a more strict set of hosts or even to the hostname of the Gophish server.

It would be really helpful if this functionality was added so we don't have to rely on the webserver configuration injecting the CORS headers.

Please let me know if anything is missing.

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,7 +3,11 @@
 		"listen_url": "127.0.0.1:3333",
 		"use_tls": true,
 		"cert_path": "gophish_admin.crt",
-		"key_path": "gophish_admin.key"
+		"key_path": "gophish_admin.key",
+		"api": {
+			"cors_allowed_origins": ["*"],
+			"cors_debug": false
+		}
 	},
 	"phish_server": {
 		"listen_url": "0.0.0.0:80",
--- a/config/config.go
+++ b/config/config.go
@@ -5,12 +5,19 @@ import (
 	"io/ioutil"
 )
 
+// ApiServer represents the API server configuration details
+type ApiServer struct {
+	CorsAllowedOrigins []string `json:"cors_allowed_origins"`
+	CorsDebug          bool     `json:"cors_debug"`
+}
+
 // AdminServer represents the Admin server configuration details
 type AdminServer struct {
-	ListenURL string `json:"listen_url"`
-	UseTLS    bool   `json:"use_tls"`
-	CertPath  string `json:"cert_path"`
-	KeyPath   string `json:"key_path"`
+	ListenURL string    `json:"listen_url"`
+	UseTLS    bool      `json:"use_tls"`
+	CertPath  string    `json:"cert_path"`
+	KeyPath   string    `json:"key_path"`
+	ApiConf   ApiServer `json:"api"`
 }
 
 // PhishServer represents the Phish server configuration details
--- a/controllers/api/api_test.go
+++ b/controllers/api/api_test.go
@@ -43,7 +43,7 @@ func (s *APISuite) SetupSuite() {
 	// static assets
 	err = os.Chdir("../")
 	s.Nil(err)
-	s.apiServer = NewServer()
+	s.apiServer = NewServer(conf.AdminConf.ApiConf)
 }
 
 func (s *APISuite) TearDownTest() {
--- a/controllers/api/server.go
+++ b/controllers/api/server.go
@@ -79,9 +79,9 @@ func (as *Server) registerRoutes() {
 	c := cors.New(cors.Options{
 		AllowedOrigins: as.config.CorsAllowedOrigins,
 		AllowedMethods: []string{"GET", "POST", "OPTIONS"},
-		AllowedHeaders: []string{"Origin", "X-Requested-With", "Content-Type", "Accept"},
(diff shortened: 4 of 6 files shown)


jordan-wright:
Hi @mdeboer!

Thanks for sending this in. While this PR looks like great work, I'd prefer if we took a simpler route. For background, I'm typically very picky when it comes to what becomes configurable, since the power of Gophish is it's simplicity and reasonable defaults. To that end, I don't want the CORS configuration to be something that's configurable within Gophish.

Instead, if the current wildcard origins isn't working, we should adjust the way the headers get applied to fix that, since it's the intention. I'd prefer if we do that using the existing implementation instead of introducing another dependency just for the sake of adding CORS headers.

If people want to scope down the whitelist of CORS origins, I'd consider that a custom case that should be implemented using a reverse proxy in front of Gophish.

mdeboer:
I strongly disagree as properly handling CORS requires some effort. It's not simply adding one header with a wildcard (although that would work) and I strongly suggest people to narrow their allowed origins.

Imho it is better to have a small and afaik commonly used and tested library to handle all CORS headers and preflight requests than to reinvent the wheel. It is small and easy to implement with at bare minimum two lines of code.

Using a reverse proxy for this is what I consider a last resort to be quite honest.

jordan-wright:
Thanks for the quick reply! I appreciate the concern, and I agree that folks looking to build client-side applications on top of Gophish should restrict the allowed origins as a best practice. However, I'm not willing to make that configurable in the `config.json`.

_Note: For the following, I'll be the first to say that I'm not an expert on CORS, so please do correct me where I'm wrong_

In general, _most_ people will just use the admin panel where CORS isn't a concern since you're hitting the same origin. Then, a subset of those folks will write server-side scripts to consume the API, where CORS isn't a concern. It's only an (IMO) even smaller subset of folks who are planning to write client-side integrations with the Gophish API who would need to consider creating a whitelist of origins for CORS.

And even then, the threat model seems pretty limited. It's my understanding that an attacker would either:

* Need to have a valid API key and get you to visit a malicious domain. But by then, they already have an API key so they don't need user interaction unless it's some kind of pivot to an internal host or something.
* Need to gain client-side code execution via something like XSS on the custom app that then leverages the API key (since it's probably an accessible variable) to then send an authenticated request.

Unless I'm missing something, to me these are both pretty narrow attack vectors. If this is within someone's threat model, that's when I'd encourage them to put a reverse proxy in front of Gophish that locks it down however they'd like.

But again, I could be missing something, so I'd be interested to hear your perspective!

mdeboer:
The dependency is very small, does not add complexity but instead removes some complexity from your code, is well tested and in the smallest case (without configuration options) can be implemented with 9 lines in a single file.

So even though you can question the use of CORS, the changes required to implement it are so small it took me longer to write this than to fix this and I'm no Go guru.

What would you say about removing the lines from the `config.json` and making those values absolutely optional, possibly not even included in the documentation?

jordan-wright:
It still seems that there is only a small number of folks for which this mitigates a potential threat, right? What's the attack vector I'm missing in https://github.com/gophish/gophish/pull/1681#issuecomment-559909285?

Unless I'm missing something there (which I totally could be) it seems this only gives flexibility to a small number of people. While making this undocumented could help the average user not have to care about CORS, it doesn't change the fact that I now have to maintain that code.

Sure, one could argue that it's not _that_ big of a deal since it's only a couple of configuration items, but they add up. It's the same reason why I don't make the number of bcrypt rounds, rate limits, or password policy configurable (#1538). Or allow Gophish to be configured to run in a subfolder (#346). Or changing the `rid` name or format (#1213).

Instead, I prefer that we keep the CORS policy simple. If someone has a use case where they need more, they will need to either recompile Gophish or run it behind a reverse proxy.

Now, to the question of using an external dependency. I'm not strictly opposed to using this one. You're right - it's well tested and fairly self-contained. At the same time, I don't think it's critically needed. Instead, since we're taking the route of having a single CORS policy within Gophish, we could just as easily modify the HTTP headers we already send to cause CORS to work as expected. So I'm on the fence about it - I wouldn't be opposed to merging a PR that just moves CORS support to leverage that package, keeping the same policy in-place, but I would also be a-ok with just updating our approach to send the right headers.

As always, I appreciate the great discussion!

mdeboer:
Well you're right about the attack vectors probably, I'm not a complete security wizard. We just have different views on configurability and maintainability I suppose.

As I really need this, I'll stick to my fork which makes it easier for me to backport changes and implement possibly other specific requirements. I'll leave this open for a while in case you change your mind.

Thanks for elaborating! 👍🏻

jordan-wright:
Hey there!

I'm going to go ahead and close this one out. I do appreciate you sending it in, and encourage you to continue sending in any ideas that you think would benefit Gophish!

Thank you for taking the time to send up this PR!

PR #1720, opened 2020-01-15
Get send_by_date

JoseW3b:
Get send_by_date field when query on GetCampaignSummary and GetCampaignSummaries


jordan-wright:
LGTM - thanks for taking the time to send this in!

Issue #1708, opened 2020-01-07
Block Auto Fill on "Send Emails By" Field on new Campaign Screen


SteveT76:
What version of Gophish are you using?: 
0.7.1

Brief description of the issue:
Using Chrome with Auto address fill in enabled. When trying to enter a date/time into the 'Send Emails By Date' field, Chrome sees this as an email field and displays a list of 'auto complete addressed over the date/time picker. So unable to select a date/time.

What are you expecting to see happen? :
To see the date/time picker 

What are you seeing happen?
A chrome address selection list displays over the top of the date/time picker.

Possible Solution:
Add autocomplete="no" html directive to the field
![Capture](https://user-images.githubusercontent.com/43610298/71909146-4bb88d80-3167-11ea-9f63-1853934ae67c.PNG)

glennzw:
I fixed this issue in [this](https://github.com/gophish/gophish/pull/1557) pull request with [this](https://github.com/gophish/gophish/pull/1557/files#diff-4db40aae803c78ba7e28b2f767cf04f1R111) code:

```<input type="text" class="form-control" id="send_by_date" autocomplete="random-data"/> <!--Chrome ignores autocomplete="off".-->```

Just need to add `autocomplete="random-data"`.

@jordan-wright I could add a pull request for just that issue if you like.

jordan-wright:
> 

Wow. TIL. That's just great.

@glennzw would you mind making this a separate PR? Probably best to keep things separate for tracking purposes if it's not too much trouble.

glennzw:
Oops I hadn't seen your comment here @jordan-wright, just sent in a pull request fixing this. https://github.com/gophish/gophish/pull/1749

jordan-wright:
Closed via #1749. Thanks @glennzw!

PR #1749, opened 2020-02-07
Fixed text autocomplete dropping down in Chrome

glennzw:
As discussed in https://github.com/gophish/gophish/issues/1708#issuecomment-571864987 and https://github.com/gophish/gophish/pull/1557


jordan-wright:
Thanks @glennzw!

Issue #1726, opened 2020-01-21
Debugging sending performance issue


edermi:
What version of Gophish are you using?: 
c553d7d9255bd686f4e169f9471da5ece0aefcb7

Brief description of the issue: 
As several users have already reported, sending performance of Gophish degrades for a large number of targets.

What are you expecting to see happen? :
Fast mail sending even for a large number of targets.

What are you seeing happen?

I did some deep dive and identified the root cause of the issue. I took care to write down all steps in order to allow you to reproduce the issue with minimal manual work. Grab a drink and take a seat..

## Setup

- The test is conducted on a Debian Buster VM running on my local machine. All actions take place locally, therefore it is **not** required to have Internet connectivity (apart from installing required packages and getting sources), an external mail server, domain configured etc. The used Go version is `go1.11.6 linux/amd64`.
- `go get` Gophish from c553d7d9255bd686f4e169f9471da5ece0aefcb7
- `go build` Gophish
- Download latest MailHog release (https://github.com/mailhog/MailHog/releases/download/v1.0.0/MailHog_linux_amd64)
- Run mailhog with default options
- Verify web interface works (http://localhost:8025)
- To verify the mail server works and to get an idea for its performance, I use the following Python script that sends 5k Mails and measures times:

~~~python
import smtplib
from email.message import EmailMessage
from timeit import default_timer as timer

start = timer()

s = smtplib.SMTP('localhost:1025')

for number in range(0, 5000):
    msg = EmailMessage()
    msg.set_content("Mail from python")
    msg['Subject'] = f'Test {number}'
    msg['From'] = "sender@example.com"
    msg['To'] = "receiver@example.com"

    s.send_message(msg)

s.quit()
end = timer()
print(end - start) 
~~~

On my small test VM (2 notebook cores, <4GB RAM) this takes about 24 seconds:

~~~bash
$ python3 --version
Python 3.7.3
$ python3 python-gophishdbg-sendmail.py
23.703909114999988
~~~

- In mailhog, we can see that all mails have been received by mailhog and no mail was lost. 

![verify_mailhog_works](https://user-images.githubusercontent.com/7430871/72819016-c061f080-3c6c-11ea-8793-3a5fdb716400.png)
- Since this is a rather simple use case and a lab setup involving no authentication, only small message sizes, in memory storage and no network latency, I'd say that it would be OK for Gophish to send 5k mails in less than 5 minutes.

[...]

PR #1753, opened 2020-02-16
Improve the Performance of Sending Emails

jordan-wright:
This PR improves the performance of sending emails. This works by caching just the relevant fields of a campaign once per batch of maillogs. This saves multiple database lookups and, as a result, reduces memory/CPU usage and sends emails significantly faster.

Some details can be found in https://github.com/gophish/gophish/issues/1726#issuecomment-578588154.


edermi:
> Thanks for the follow-up! Just curious, if you sort the results in the UI by status, do you see any results with “Error”? This will help track down if they ever tried to send or if there was an error when sending. 

No, all mails have a green "Email Sent" badge.

jordan-wright:
@edermi - Sorry for the delayed response.

That's definitely odd. I'm going to go ahead and merge this in for now since it seems to improve speed and have the same results we saw earlier regarding some emails not making it through.

If you get a chance, could you run the campaign again, dumping all the Gophish output to a file you can send my way? I think that'll help us track down which emails aren't getting sent and why. We can then tackle any bugs as needed.

Thanks for all your patience here!

edermi:
No worries, thanks for implementing and merging the change.

I'll try to take a closer look again as soon as possible.

edermi:
So, I found some time to test again today.

I built gophish from master today, used my scripts to fill gophish with test data and sent the mails, but only 4989 leave the server. Here is the log of the mail sending: https://gist.github.com/edermi/f66ab8d4e719229ed031fd57900725eb

I don't see any errors there, and these are in fact 4989 log lines indicating a mail was sent. Mailhog only receives 4989 mails but the set of targets is exactly 5k mail addresses.

Issue #1226, opened 2018-10-08
Reported button


securitygeneration:
It would be nice to add a button that marks a particular user as having reported an email as phishing. While I know the current functionality is designed to be integrated into other reporting tools, the majority of 'reports' are still by email to the IT or security team, and it would be nice to be able to log that within GoPhish for metrics purposes.

dsmurf:
Where do you want the report button to be?
you can program it in html.

securitygeneration:
I envisage it should live at the bottom of the user detail timeline. And when you click it, a Reported entry get entered into the timeline, and the button disappears (no point in multiple reports).

![image](https://user-images.githubusercontent.com/1046239/46656711-9931fd80-cba6-11e8-9293-7df90786a66b.png)

dsmurf:
Ahhh you want a button in the admin panel where a admin can mark that a user reported the mail.

securitygeneration:
Yeah. I think this should be fairly trivial to implement.

jordan-wright:
I like the idea in your screenshot, @securitygeneration. I think this is a good place to put a general "Actions" section that can be populated with various actions, as needed.

For example, in addition to marking the user as reported, I'd like to also offer a button that can re-send the email in the case of errors.

The thing is- I don't want to just have this point to the `/report` endpoint, since that might be misleading. For example, we collect the user-agent and IP address of the person who hit that endpoint so that might cause the logs to appear mismatched. Instead, I'm considering establishing API endpoints at `campaigns/<id>/results/<rid>/report`, `campaigns/<id>/results/<rid>/resend`, etc. We can then find a way in the logs and UI to indicate that an admin manually performed that action on behalf of the user.

Curious to hear y'all's thoughts!

dsmurf:
Jordan i think this would be awesome,.Would love that.

securitygeneration:
Resending would indeed be nice to have.

I agree that you'd want to be able to track an email as 'reported' without logging the rest of the info normally captured from the reporting user (IP, user agent, etc).

mcvic1rj:
@jordan-wright a suggestion for the new report endpoint is to allow it to take a datetime object as well. That way you can insert the event, and have it displayed in the correct location in the timeline.

dsmurf:
Is there any timeline on this?

blacklist-arcc:
[...]

PR #1772, opened 2020-03-04
add "mark as reported" to results table

FLX-0x00:
Reference: #1226 
The small icon in the result view (reported column) is now clickable if the email is not marked as reported. On click you have to confirm a dialogue which will mark the selected email as reported directly. We are facing some situations where this feature can be useful for example if a user report a phishing attempt via phone call. Tested on our servers, works great!


static/js/src/app/campaign_results.js
```diff
@@ -925,7 +925,26 @@ function refresh() {
...
+    if (r == true) {
+        api.campaignId.get(cid).success((function(c) {
+            report_url = c.url;
+            report_url += report_url.endsWith("/") ? "" : "/";
+            report_url += "report?rid=" + rid;
+            $.ajax({
+                url: report_url,
+                method: "GET",
+                success: function(data) {
+                    return; // wait for response
```
jordan-wright: What do you think about moving the call to refresh here, that way it fires as soon as the `success` callback is fired?

jordan-wright:
Hey there!

I'll give this a shot over the next couple of days. Thanks for sending it in!

Could you do me a favor and remove the .min.js file? I prefer to generate those myself since they're difficult to review in a PR.

I've also added a little comment that might make it such that the `refresh` call will fire a bit faster vs using `setTimeout`. Let me know what you think!

static/js/src/app/campaign_results.js
blacklist-arcc: ofc, that will be a better code style! The timout / sleep function is important for the refresh to work, not the best solution i think, but works good. I will make a commit

FLX-0x00 (commented):
ofc, that will be a better code style! The timout / sleep function is important for the refresh to work, not the best solution i think, but works good. Without the timout, the refresh is not working...I will make a commit.

jordan-wright:
Doing some local testing, I didn't see that the `setTimeout` call is needed. And it _shouldn't_ be needed, since the `success` callback only fires after the call to `/report` is completed. So I'd recommend removing it.

And one last thing - would we please change the confirmation to leverage the SweetAlert2 library? This is how we do confirmations in other parts of Gophish. You can find an example of that here https://github.com/gophish/gophish/blob/master/static/js/src/app/campaign_results.js#L128

Let me know if you have any questions! This will be a good stop gap until we can figure out a longer term solution for doing a number of actions per recipient (resending emails, etc.)

blacklist-arcc:
All right. Sound nice... I'll give it a shot and make some tests in the next days/pentests. Thanks for your input!

blacklist-arcc:
Done. Tested in productive environment. Works as expected without any issues.

jordan-wright:
LGTM - thank you!

blacklist-arcc:
The strict Content-Security-Policy do not allow ajax requests from https to http. If you are using gophish with ssl and your landingpage not using ssl, the report-request will fail. So be sure to setup all your resources with https.

PR #1773, opened 2020-03-04
Fixing issues - get values from non objects

hwu16:
After call gets Template, Page, SMTP and returns not found, trying to display error message using the name values of the non-objects.


jordan-wright:
LGTM - thanks for sending this in!

PR #1774, opened 2020-03-04
Remove Unused Variable

hwu16:
The timeline_series_data variable created twice before using it.


jordan-wright:
Sure enough! I have no idea why that's in there, but it looks like a bug to me.

Thanks for sending this in!

PR #1786, opened 2020-03-15
Updated the Ansible role

PaulSec:
Hi!

I updated the Ansible role to reflect: 
- changes on the `config.json` file
- make sure that ufw is installed on the remote machine before any operations
- generate a SSL certificate on the fly to serve the app
- use the Ansible capabilities module instead of some shell-fu
- created a gophish service which runs out of the box

Have a great day!

### DIFF
--- a/ansible-playbook/roles/gophish/tasks/main.yml
+++ b/ansible-playbook/roles/gophish/tasks/main.yml
@@ -2,22 +2,27 @@
   hostname:
     name: "{{ hostname }}"
 
+- name: Ensure ufw is installed on the machine
+  package:
+    name: ufw
+    state: present
+
 - name: Allow TCP 22 for SSH.
   ufw:
     rule: allow
-    port: 22
+    port: '22'
     proto: tcp
 
 - name: Allow TCP 80 for Gophish.
   ufw:
     rule: allow
-    port: 80
+    port: '80'
     proto: tcp
 
 - name: Allow TCP 443 for Gophish.
   ufw:
     rule: allow
-    port: 443
+    port: '443'
     proto: tcp
 
 - name: Enable ufw.
@@ -34,11 +39,55 @@
   apt:
     upgrade: safe
 
+- name: Ensure /etc/ssl/csr folder exists
+  file:
+    path: /etc/ssl/csr
+    state: directory
+    mode: '0755'
+  
+- name: Ensure /etc/ssl/private folder exists
+  file:
+    path: /etc/ssl/private
+    state: directory
+    mode: '0755'
+
+- name: Ensure /etc/ssl/crt folder exists
+  file:
+    path: /etc/ssl/crt
+    state: directory
+    mode: '0755'
+
 - name: Install specified packages.
   apt:
-    pkg: "{{ item }}"
+    pkg: "{{ install_packages }}"
     state: latest
-  with_items: "{{ install_packages }}"
+
+- name: adding existing user '{{ gophish_user }}' to group ssl-cert
+  user:
+    name: '{{ gophish_user }}'
+    groups: ssl-cert
+    append: yes
+
+- name: Ensure the cryptography Python package is installed
+  pip:
+    name: cryptography
+
+- name: Generate an OpenSSL private key with the default values (4096 bits, RSA)
+  openssl_privatekey:
+    path: "{{ gophish_ssl_cert_path }}"
+
+- name: Generate an OpenSSL Certificate Signing Request
+  openssl_csr:
+    path: "{{ gophish_csr_path }}"
+    privatekey_path: "{{ gophish_ssl_cert_path }}"
+    common_name: "{{ gophish_domain }}"
+
+- name: Generate a Self Signed OpenSSL certificate
+  openssl_certificate:
+    path: "{{ gophish_crt_path }}"
+    privatekey_path: "{{ gophish_ssl_cert_path }}"
+    csr_path: "{{ gophish_csr_path }}"
+    provider: selfsigned
 
 - name: Update postfix main.cf configuration file.
   template:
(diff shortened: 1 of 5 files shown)


ansible-playbook/roles/gophish/tasks/main.yml
```diff
@@ -34,11 +39,45 @@
...
+    mode: '0755'
+
+- name: Ensure /etc/ssl/crt folder exists
+  file:
+    path: /etc/ssl/crt
+    state: directory
+    mode: '0755'
+
+- name: Generate an OpenSSL private key with the default values (4096 bits, RSA)
+  openssl_privatekey:
```
jordan-wright: It looks like the [`openssl_*`](https://docs.ansible.com/ansible/latest/modules/openssl_privatekey_module.html) modules require either the `cryptography` or `pyOpenSSL` Python packages to work. When doing this on a vanilla Debian system, I got an error since neither of these were installed.

It seems that we could add `python-pip` to the `{{install_packages}}`, bump up the "Install specified packages." task up before the cert generation tasks, and add a task that installs the `cryptography` Python module.

Something like:

```yml
- name: Ensure the cryptography Python package is installed
  pip:
    name: cryptography
```

ansible-playbook/roles/gophish/tasks/main.yml
```diff
@@ -80,29 +133,48 @@
...
-    path: /home/{{ gophish_user }}/gophish
+    path: /home/{{ gophish_user }}/gophish_deploy
     owner: "{{ gophish_user }}"
     group: "{{ gophish_user }}"
     recurse: True
 
-- name: Allow gophish binary to bind to privileged ports using setcap.
-  shell: setcap CAP_NET_BIND_SERVICE=+eip /home/{{ gophish_user }}/gophish/gophish
+- name: Ensure gophish binary is allowed to bind to privileged ports using setcap
+  capabilities:
```
jordan-wright: As part of this, could we add `libcap2-bin` to the default `{{ install_packages }}`? In a vanilla Ubuntu system, I got an error that `getcap` wasn't available.

jordan-wright (changes requested):
Hey @PaulSec!

Thanks for sending this in. This is looking great!

I've added a couple of comments. If we could get those knocked out, I'll be happy to merge this in.

ansible-playbook/roles/gophish/tasks/main.yml
PaulSec: I added it!

ansible-playbook/roles/gophish/tasks/main.yml
PaulSec: You're right, I also added it :)

PaulSec:
Btw, just so you know, I also added a nginx to serve as a reverse proxy and made the CN (for the certificate + nginx) a variable that you can change in `vars/main.yml`

PaulSec:
Moreover, I updated the role so that it spins up a nginx as a reverse proxy for the gophish server. 
The admin panel will only available on localhost.

jordan-wright:
Sounds great! I’ll give this a shot ASAP and will merge if everything checks out.

Thanks for all the help on this!

jordan-wright:
This is the error I'm getting on a fresh Debian install:

```
root@gophish:/home/jordan# service gophish status
● gophish.service - gophish
   Loaded: loaded (/etc/systemd/system/gophish.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Tue 2020-03-17 22:18:12 CDT; 3min 5s ago
  Process: 461 ExecStart=/home/jordan/gophish_deploy/gophish (code=exited, status=1/FAILURE)
 Main PID: 461 (code=exited, status=1/FAILURE)

Mar 17 22:18:11 gophish systemd[1]: Started gophish.
Mar 17 22:18:12 gophish gophish[461]: time="2020-03-17T22:18:12-05:00" level=warning msg="No contact address has been configured."
Mar 17 22:18:12 gophish gophish[461]: time="2020-03-17T22:18:12-05:00" level=warning msg="Please consider adding a contact_address entry in your config.json"
Mar 17 22:18:12 gophish gophish[461]: goose: no migrations to run. current version: 20200116000000
Mar 17 22:18:12 gophish gophish[461]: time="2020-03-17T22:18:12-05:00" level=info msg="Starting admin server at https://127.0.0.1:3333"
Mar 17 22:18:12 gophish gophish[461]: time="2020-03-17T22:18:12-05:00" level=fatal msg="open /etc/ssl/private/gophish.pem: permission denied"
Mar 17 22:18:12 gophish systemd[1]: gophish.service: Main process exited, code=exited, status=1/FAILURE
Mar 17 22:18:12 gophish systemd[1]: gophish.service: Unit entered failed state.
Mar 17 22:18:12 gophish systemd[1]: gophish.service: Failed with result 'exit-code'.
```

The error appears to be stemming from the fact that I'm using an unprivileged user, which is trying to access `/etc/ssl/private` which is locked down:

```
jordan@gophish:~$ ls -alh /etc/ssl/
total 52K
drwxr-xr-x  6 root root     4.0K Mar 17 22:15 .
drwxr-xr-x 81 root root     4.0K Mar 17 22:19 ..
drwxr-xr-x  2 root root      20K Mar 17 22:16 certs
drwxr-xr-x  2 root root     4.0K Mar 17 22:16 crt
drwxr-xr-x  2 root root     4.0K Mar 17 22:16 csr
-rw-r--r--  1 root root      11K Nov 28  2018 openssl.cnf
drwx--x---  2 root ssl-cert 4.0K Mar 17 22:16 private
```

Perhaps it makes sense for us to ensure that the `{{ gophish_user }}` is in the `ssl-cert` group?

jordan-wright (changes requested):
Requesting changes due to the comment above.

jordan-wright:
Oops! Accidentally closed this out and reopened. My bad!

PaulSec:
You are totally right Jordan. I changed that to make sure that the user is in the `ssl-cert` group using: 

```yaml
- name: adding existing user '{{ gophish_user }}' to group ssl-cert
  user:
    name: '{{ gophish_user }}'
    groups: ssl-cert
    append: yes
```

Thanks again for the review! I tested it on a vanilla Debian 10.

ansible-playbook/roles/gophish/tasks/main.yml
```diff
@@ -34,11 +39,55 @@
...
+    state: directory
+    mode: '0755'
+
+- name: Ensure /etc/ssl/crt folder exists
+  file:
+    path: /etc/ssl/crt
+    state: directory
+    mode: '0755'
+
+- name: adding existing user '{{ gophish_user }}' to group ssl-cert
```
jordan-wright: Hey @PaulSec! For whatever reason, I'm getting an error that the `ssl-cert` group doesn't exist. However, if I move the package installation step before this one, things work great. Can we make that change, then we should be in good shape to merge!

PaulSec: Haha, all good then! (it's weird it worked out of the box for me on a Debian 10). All resolved :)

jordan-wright:
Looks like this works great. Thanks @PaulSec!

PR #1791, opened 2020-03-24
IMAP update; new library and attachment support

glennzw:
This pull request replaces the antiquated `github.com/mxk/go-imap` library with the actively maintained `github.com/emersion/go-imap`. 

I've also added support for reporting campaign emails as attachments.


jordan-wright:
🎉 Awesome! I’ll give this a look over ASAP. Thanks @glennzw!

imap/imap.go
```diff
@@ -150,203 +65,181 @@ func Validate(s *models.IMAP) error {
...
 			if err != nil {
-				responses <- Response{Err: fmt.Errorf("unable to remove seen flag: %s", err)}
-				return
+				return emails, err
 			}
-		}
 
-		if delete {
-			err = deleteEmail(client, imap.AsNumber(msgFields["UID"]))
+			// Reload the reader 🔫
```
jordan-wright: Let's avoid having emojis in the comments if possible 😄 Go totally supports it, but I'd like to keep things easy.

imap/imap.go
```diff
@@ -150,203 +65,181 @@ func Validate(s *models.IMAP) error {
...
-	if plus {
-		flg = "+FLAGS"
-	}
-	fSeq := &imap.SeqSet{}
-	fSeq.AddNum(UID)
-	_, err := imap.Wait(client.UIDStore(fSeq, flg, flag))
-	if err != nil {
-		return err
+		} // On to the next email
+	} else {
```
jordan-wright: If we're not going to keep this log, let's remove this `else`.

imap/monitor.go
```diff
@@ -192,3 +200,43 @@ func checkForNewEmails(im models.IMAP) {
...
+		if a.Header.Get("Content-Type") == "message/rfc822" || (len(a.Filename) > 3 && a.Filename[len(a.Filename)-4:] == ".eml") {
+
+			//Let's decode the email
+			rawBodyStream := bytes.NewReader(a.Content)
+			attachementemail, err := email.NewEmailFromReader(rawBodyStream)
+			if err != nil {
+				return rids, err
+			}
+
+			for _, r := range goPhishRegex.FindAllStringSubmatch(string(attachementemail.Text)+string(attachementemail.HTML), -1) {
```
jordan-wright: This is quite a long line. Can we break this up into multiple operations? Like, creating the string, then having `goPhishRegex.FindAllStringSubmatch(content)` or similar?

jordan-wright (changes requested):
Hey @glennzw!

Had some time to look into this a bit. I'd like to rename `imapclient` to the camelCase `imapClient`. I added some suggestions, but then I realized it'll just be easier for you to find/replace 😄 

I added a few other little comments and questions. I think this is looking good! Here in the next couple of days I'll start actually testing it out and I'm sure I'll have more questions for you.

As a side note, I sure do like PR's that remove more code than they add 😛 Great work!

imap/imap.go
```diff
@@ -150,203 +65,181 @@ func Validate(s *models.IMAP) error {
...
-		// http://mailman13.u.washington.edu/pipermail/imap-protocol/2014-October/002355.html
-		// http://stackoverflow.com/questions/26262472/gmail-imap-is-sometimes-returning-bad-results-for-fetch
-		if _, ok := msgFields["RFC822.HEADER"]; !ok {
-			continue
-		}
+	if len(seqs) > 0 {
+		seqset := new(imap.SeqSet)
+		seqset.AddNum(seqs...)
+		section := &imap.BodySectionName{}
+		items := []imap.FetchItem{imap.FetchEnvelope, imap.FetchFlags, imap.FetchInternalDate, section.FetchItem()} // Check this
```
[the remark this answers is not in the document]
glennzw: Ah just a note to myself to make sure I was selecting the appropriate email parts. Removed.

imap/imap.go
glennzw: 🙃

imap/imap.go
```diff
@@ -150,203 +65,181 @@ func Validate(s *models.IMAP) error {
...
+	defer imapclient.Logout()
 
-func (mbox *Mailbox) generateMail(search string, since *time.Time, markAsRead, delete bool) (chan Response, error) {
-	responses := make(chan Response, GenerateBufferSize)
-	client, err := mbox.newClient()
+	seqSet := new(imap.SeqSet)
+	seqSet.AddNum(seqs...)
+
+	item := imap.FormatFlagsOp(imap.AddFlags, true)
+	flags := []interface{}{imap.DeletedFlag}
```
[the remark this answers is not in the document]
glennzw: Hmm I'm not sure, changed.

glennzw:
Great thanks for the feedback @jordan-wright ! Addressed your issues and pushed changes.

glennzw:
I've also added support for invalid TLS certificates on IMAP servers, but I'll wait until this is merged into master as there are some UI and db changes.

glennzw:
IMAP messages can be identified by either sequence number (SeqNum) or unique id (UID). The code above fetches by sequence number, which is fine, but as they change when the inbox changes I'd like to change to UID, but I'm having trouble with the `FetchUid` function. If anyone has some IMAP experience, or just wants to help me out, here's the issue I filed that I'm stuck on:

https://github.com/emersion/go-imap/issues/353

S0larflare:
I’ll take a look tomorrow at my code as I remember having a similar problem with UID from the emersion IMAP

Sent from my iPhone

On 22 Apr 2020, at 13:57, Glenn Wilkinson <notifications@github.com> wrote:

﻿

IMAP messages can be identified by either sequence number (SeqNum) or unique id (UID). The code above fetches by sequence number, which is fine, but as they change when the inbox changes I'd like to change to UID, but I'm having trouble with the FetchUid function. If anyone has some IMAP experience, or just wants to help me out, here's the issue I filed that I'm stuck on:

emersion/go-imap#353<https://github.com/emersion/go-imap/issues/353>

glennzw:
Extra info; this condition happens with Gmail, but on Outlook the Uid function works fine. Weird. Might have to crack open some socat tunnels and inspect traffic 🙄

glennzw:
> I’ll take a look tomorrow at my code as I remember having a similar problem with UID from the emersion IMAP

I found a post from [2012](https://support.google.com/mail/forum/AAAAK7un8RUj-snr8u0p6s/?hl=en-GB) of someone having a similar issue with GMail, the suggestion being to send a NOOP (no op) command to the server. This has fixed the issue.

```
go func() {
	done <- c.Fetch(seqset, items, messages)
}()

c.Noop()
```

IMAP is a horrible piece of technology.

glennzw:
Hey @jordan-wright; I've decided I'm going to stick with SeqNums over UIDs for now. There's a slight advantage to using UIDs for some other gophish features that I'm working on, but not essential for now.

Would appreciate a PR when you get a chance 🙂

imap/imap.go
```diff
@@ -150,203 +63,176 @@ func Validate(s *models.IMAP) error {
...
 
-	if seq.Empty() {
-		return
-	}
+	defer imapClient.Logout()
 
-	fCmd, err := imap.Wait(client.UIDFetch(seq, "INTERNALDATE", "BODY[]", "UID", "RFC822.HEADER"))
+	// Search for unread emails
+	criteria := imap.NewSearchCriteria()
+	criteria.WithoutFlags = []string{"\\Seen"}
```
jordan-wright, line 133: I like to defer to package constants where possible. It looks like the package has a [`SeenFlag="\\Seen`](https://godoc.org/github.com/emersion/go-imap#pkg-constants) that we might be able to use here.

Here's what that might look like (please double-check me though 😄)

```suggestion
	criteria.WithoutFlags = []string{imap.SeenFlag}
```

imap/imap.go
```diff
@@ -150,203 +63,176 @@ func Validate(s *models.IMAP) error {
...
-
-		// make sure is a legit response before we attempt to parse it
-		// deal with unsolicited FETCH responses containing only flags
-		// I'm lookin' at YOU, Gmail!
-		// http://mailman13.u.washington.edu/pipermail/imap-protocol/2014-October/002355.html
-		// http://stackoverflow.com/questions/26262472/gmail-imap-is-sometimes-returning-bad-results-for-fetch
-		if _, ok := msgFields["RFC822.HEADER"]; !ok {
-			continue
-		}
+	if len(seqs) > 0 {
```
jordan-wright, line 139: When possible, I recommend catching and handling conditions early, since that helps avoid long nested blocks.

Here, we can do something like this:

```go
if len(seqs) == 0 {
    return emails, nil
}
// At this point, we know that we have emails we can handle so we can continue as normal.
// The benefit is that we've saved one level of indentation which is cleaner to read.
seqset := new(imap.SeqSet)
...
```

jordan-wright:
Hey @glennzw, just checking in. I'll be working to review this over the next couple of days. My time for PR reviews is spread pretty thin, so I appreciate you bearing with me 😄 

I'll keep you updated!

imap/imap.go
glennzw: Done.

imap/imap.go
glennzw: Good point, done!

imap/monitor.go
```diff
@@ -147,41 +150,46 @@ func checkForNewEmails(im models.IMAP) {
...
-				log.Infof("User '%s' reported email with rid %s", m.Email.From, rid)
-				result, err := models.GetResult(rid)
-				if err != nil {
-					log.Error("Error reporting GoPhish email with rid ", rid, ": ", err.Error())
-					reportingFailed = append(reportingFailed, m.UID)
-				} else {
-					err = result.HandleEmailReport(models.EventDetails{})
+			if err != nil {
+				log.Errorf("Error searching email for rids from user '%s': %s", m.Email.From, err.Error())
+			} else {
```
jordan-wright, line 157: What do you think about removing this `else` and moving this back up a level?

imap/monitor.go
```diff
@@ -147,41 +150,46 @@ func checkForNewEmails(im models.IMAP) {
...
+					// In the future this should be an alert in Gophish
+					log.Infof("User '%s' reported email with subject '%s'. This is not a GoPhish campaign; you should investigate it.", m.Email.From, m.Email.Subject)
+				}
+				for rid := range rids {
+					log.Infof("User '%s' reported email with rid %s", m.Email.From, rid)
+					result, err := models.GetResult(rid)
 					if err != nil {
-						log.Error("Error updating GoPhish email with rid ", rid, ": ", err.Error())
+						log.Error("Error reporting GoPhish email with rid ", rid, ": ", err.Error())
+						reportingFailed = append(reportingFailed, m.SeqNum)
```
jordan-wright, line 167: Should we just `continue` here and bump the `else` up a level?

imap/monitor.go
```diff
@@ -147,41 +150,46 @@ func checkForNewEmails(im models.IMAP) {
...
 					} else {
-						if im.DeleteReportedCampaignEmail == true {
-							campaignEmails = append(campaignEmails, m.UID)
+						err = result.HandleEmailReport(models.EventDetails{})
+						if err != nil {
+							log.Error("Error updating GoPhish email with rid ", rid, ": ", err.Error())
```
jordan-wright, line 171: Likewise, should we just `continue` here and bump the `else` up a level?

imap/monitor.go
```diff
@@ -192,3 +200,45 @@ func checkForNewEmails(im models.IMAP) {
 		log.Debug("No new emails for ", im.Username)
 	}
 }
+
+// returns a slice of gophish rid paramters found in the email HTML, Text, and attachments
+func checkRIDs(em *email.Email) (map[string]int, error) {
```
jordan-wright, line 205: What's the use case you see where more than one `rid` can be reported per email? I can't really see that happening, but I could be missing something.

If we don't anticipate that happening, this might be a bit easier to implement.

imap/monitor.go
```diff
@@ -192,3 +200,45 @@ func checkForNewEmails(im models.IMAP) {
...
+
+	rids := make(map[string]int)
+
+	// Check Text and HTML
+	emailContent := string(em.Text) + string(em.HTML)
+	for _, r := range goPhishRegex.FindAllStringSubmatch(emailContent, -1) {
+		newrid := r[len(r)-1]
+		if _, ok := rids[newrid]; ok {
```
jordan-wright, line 213: Small nit: we can make this a bit cleaner. Something like:

```go
rid := r[len(r)-1]
if _, ok := rids[newrid]; !ok {
    rids[newrid] = 0
}
rids[newrid]++
```

imap/monitor.go
```diff
@@ -192,3 +200,45 @@ func checkForNewEmails(im models.IMAP) {
...
+			rids[newrid]++
+		} else {
+			rids[newrid] = 1
+		}
+	}
+
+	//Next check each attachment
+	for _, a := range em.Attachments {
+		if a.Header.Get("Content-Type") == "message/rfc822" || (len(a.Filename) > 3 && a.Filename[len(a.Filename)-4:] == ".eml") {
```
jordan-wright, line 222: I think that you might prefer the `filepath.Ext` function to grab the extension and check if it's `.eml`. That would be cleaner than checking the filename length and last 4 characters.

```go
ext := filepath.Ext(a.Filename)
if a.Header.Get("Content-Type") == "message/rfc822" || ext == ".eml" {
...
}
```

imap/monitor.go
```diff
@@ -192,3 +200,45 @@ func checkForNewEmails(im models.IMAP) {
...
+
+			//Let's decode the email
+			rawBodyStream := bytes.NewReader(a.Content)
+			attachementEmail, err := email.NewEmailFromReader(rawBodyStream)
```
jordan-wright, line 226: Small nit: `attachmentEmail` typo.

imap/monitor.go
```diff
@@ -192,3 +200,45 @@ func checkForNewEmails(im models.IMAP) {
...
+			if err != nil {
+				return rids, err
+			}
+
+			emailContent := string(attachementEmail.Text) + string(attachementEmail.HTML)
+			for _, r := range goPhishRegex.FindAllStringSubmatch(emailContent, -1) {
```
jordan-wright, line 232: Since you're repeating this code here, you might consider breaking it out into a function. In fact, I think that `checkRIDs` might actually be a good name for that function since it actually checks for the RIDs in content, and we can rename this one to be `matchEmail` or something - I'm open to suggestions!

imap/imap.go
```diff
@@ -150,203 +63,178 @@ func Validate(s *models.IMAP) error {
...
 
-func deleteEmail(client Client, UID uint32) error {
-	return alterEmail(client, UID, "\\DELETED", true)
-}
+		emtmp := Email{Email: em, SeqNum: msg.SeqNum} // Not sure why msg.Uid is always 0, so swapped to sequence numbers
+		emails = append(emails, emtmp)
 
-func removeSeen(client Client, UID uint32) error {
-	return alterEmail(client, UID, "\\SEEN", false)
+	} // On to the next email
```
jordan-wright, line 209: We can remove this comment

jordan-wright (changes requested):
Hey @glennzw!

I finally had a chance to dig into this a little bit. This is looking great! Doing some testing, things seem to work which is awesome.

The comments below are really just some tweaks that might make the code a little bit cleaner. Let me know if you have any questions!

imap/imap.go
```diff
@@ -150,203 +63,178 @@ func Validate(s *models.IMAP) error {
...
-	if seq.Empty() {
-		return
-	}
 
-	fCmd, err := imap.Wait(client.UIDFetch(seq, "INTERNALDATE", "BODY[]", "UID", "RFC822.HEADER"))
-	if err != nil {
-		responses <- Response{Err: fmt.Errorf("unable to perform uid fetch: %s", err)}
-		return
-	}
+		//Remove CR characters, see https://github.com/jordan-wright/email/issues/106
```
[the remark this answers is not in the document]
glennzw: Yes it was one email from Google. I've removed this while we figure out what's going on.

imap/imap.go
```diff
@@ -150,203 +63,178 @@ func Validate(s *models.IMAP) error {
...
-		if _, ok := msgFields["RFC822.HEADER"]; !ok {
-			continue
+		rawBodyStream := bytes.NewReader(buf)
+		em, err = email.NewEmailFromReader(rawBodyStream) // Parse with @jordanwright's library
+		if err != nil {
+			return emails, err
 		}
 
-		email, err = NewEmail(msgFields)
+		// Reload the reader
```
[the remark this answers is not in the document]
glennzw: Ah brilliant, nice addition to your library! Always feels good to remove a chunk of superfluous code.

imap/monitor.go
glennzw: If I understand correctly, remove the `else` and change the error check to have a continue to step over to the next email?

```
if err != nil {
	log.Errorf("Error searching email for rids from user '%s': %s", m.Email.From, err.Error())
	continue
} 
if len(rids) < 1 { .......
```

imap/monitor.go
glennzw: Done.

imap/monitor.go
glennzw: Done.

imap/monitor.go
glennzw: So I was trying to cover as many eventualities as possible - perhaps an email has an rid parameter that's not from gophish (some other developer created links with the same rid parameter name/form), or someone forwards two different campaign emails in one.

Happy to remove and simplify if you'd prefer though. We could change the function to singular `checkRID` and just return the first one it finds.

imap/monitor.go
glennzw: Ah, much neater pattern, thanks!

imap/monitor.go
glennzw: Much better, thanks!

imap/monitor.go
glennzw: Eagle eyes, well spotted 🦅

imap/monitor.go
jordan-wright: Right. It looks to me that this would fire if something broke when checking for the regex in the email content.

This would mean that we can't trust the result of `rids`. At that point, since the rest of the loop isn't applicable, we should just continue on with the next email.

imap/monitor.go
jordan-wright: Another question for you. If `HandleEmailReport` fails and gives an error, should we append it to `reportingFailed`?

imap/monitor.go
glennzw: I did think about this when coding it, and have given it some more thought - it's a little tricky as I'm using a map to maintain the unique list of rids.  I think the easiest solution is to implement the two separate functions as you suggest and pass the rid map so we can update it.

Well, I implemented that, let me know if that's OK or if you have a neater idea :)

Also, I do realise we're not checking attachments of attachments, etc. I can save that for later.

imap/monitor.go
```diff
@@ -147,41 +150,46 @@ func checkForNewEmails(im models.IMAP) {
...
 					}
+
 				}
-			} else {
-				// In the future this should be an alert in Gophish
-				log.Debugf("User '%s' reported email with subject '%s'. This is not a GoPhish campaign; you should investigate it.\n", m.Email.From, m.Email.Subject)
 			}
+
 			// Check if any emails were unable to be reported, so we can mark them as unread
 			if len(reportingFailed) > 0 {
```
jordan-wright, line 183: Right now, this fires inside the `for _, m := range msgs {` loop, but it seems as though that's a bug.

Instead, I think the `for` loop should close before we mark any emails as unread or delete them. Imagine the scenario where I get two emails marked as reported in the same run of the monitor. In that case, the first email will be deleted, then in the next iteration of the loop, Gophish will try to delete both the first and the second emails.

This looks like this:

```
time="2020-06-13T08:34:05-05:00" level=info msg="User 'XXXX' reported email with rid FcHeBrc"
time="2020-06-13T08:34:05-05:00" level=info msg="Deleting 1 campaign emails"
time="2020-06-13T08:34:06-05:00" level=info msg="User 'XXXX' reported email with rid 7eXoql6"
time="2020-06-13T08:34:06-05:00" level=info msg="Deleting 2 campaign emails"
```

imap/monitor.go
```diff
@@ -147,41 +150,46 @@ func checkForNewEmails(im models.IMAP) {
...
+						reportingFailed = append(reportingFailed, m.SeqNum)
 					} else {
-						if im.DeleteReportedCampaignEmail == true {
-							campaignEmails = append(campaignEmails, m.UID)
+						err = result.HandleEmailReport(models.EventDetails{})
+						if err != nil {
+							log.Error("Error updating GoPhish email with rid ", rid, ": ", err.Error())
+						} else {
+							if im.DeleteReportedCampaignEmail == true {
+								campaignEmails = append(campaignEmails, m.SeqNum)
```
jordan-wright, line 174: You might consider renaming this to `deleteEmails` or some other name that indicates the emails are scheduled to be deleted. Right now, that's the only reason we have this list.

By doing that, you can drop the `im.DeleteReportedCampaignEmail == true` check later on, since you're already checking it here.

That is, unless we had other plans for the campaign emails besides just deleting them. 😄

imap/imap.go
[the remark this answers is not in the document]
glennzw: So on my first  test attachment after remove the CR cleanup I got an error:

`time="2020-06-13T14:36:15+01:00" level=error msg="malformed MIME header line: \r--000000000000a0ce9605a19d6401"`

Previously I'd seen this on GMail with a Google email, but this attachment was on Outlook.com and a crafted phishing email I had from another user on here. Perhaps we can leave the CR cleanup for now and remove once we find a fix?

imap/monitor.go
glennzw: Oh gosh, sloppy! Fixed.

imap/monitor.go
glennzw: Other plans? Well I was going to add support to Tweet out emails before being deleted, but fair enough I'll rename ;-)

imap/imap.go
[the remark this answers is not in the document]
jordan-wright: Sure, that's reasonable!

imap/imap.go
```diff
@@ -150,203 +61,150 @@ func Validate(s *models.IMAP) error {
...
-	flg := "-FLAGS"
-	if plus {
-		flg = "+FLAGS"
+// newClient will initiate a new IMAP connection with the given creds.
+func (mbox *Mailbox) newClient() (*client.Client, error) {
+	var imapClient *client.Client
+	var err error
+	if mbox.TLS {
+		imapClient, err = client.DialTLS(mbox.Host, new(tls.Config))
+		if err != nil {
```
jordan-wright, line 189: Small nit - we can move the `if err != nil` check outside the `if mbox.TLS` conditional to avoid repeating ourselves.

```go
if mbox.TLS {
    imapClient, err = client.DialTLS(mbox.Host, new(tls.Config))
} else {
    imapClient, err = client.Dial(mbox.Host)
}
if err != nil {
    return imapClient, err
}
```

imap/monitor.go
```diff
@@ -147,48 +151,91 @@ func checkForNewEmails(im models.IMAP) {
...
+		if _, ok := rids[newrid]; !ok {
+			rids[newrid] = 0
+		}
+		rids[newrid]++
+	}
+}
+
+// returns a slice of gophish rid paramters found in the email HTML, Text, and attachments
+func matchEmail(em *email.Email) (map[string]int, error) {
+
```
jordan-wright, line 220: Super small nit: Can we remove any blank lines between the function definition and the start of the function?

imap/monitor.go
```diff
@@ -147,48 +151,91 @@ func checkForNewEmails(im models.IMAP) {
...
+			}
+		}
+		
 	} else {
 		log.Debug("No new emails for ", im.Username)
 	}
 }
+
+func checkRIDs(em *email.Email, rids map[string]int){
+
```
jordan-wright, line 206: Super small nit: Can we remove any blank lines between the function definition and the start of the function?

imap/monitor.go
```diff
@@ -147,48 +151,91 @@ func checkForNewEmails(im models.IMAP) {
...
+	// Check Text and HTML
+	emailContent := string(em.Text) + string(em.HTML)
+	for _, r := range goPhishRegex.FindAllStringSubmatch(emailContent, -1) {
+		newrid := r[len(r)-1]
+		if _, ok := rids[newrid]; !ok {
+			rids[newrid] = 0
+		}
+		rids[newrid]++
```
jordan-wright, line 214: So I thought a bit more about this, and this is really just being used as a set. To that end, we don't actually need to keep track of the number of times we've seen this rid.

This let's us make things even easier, since the ++ line isn't need. To make things _even simpler_ we can use the fact that in Go, retrieving a value using a key that doesn't exist gives you the zero value for that value's type. You can see an example [here](https://blog.golang.org/maps#TOC_4.).

This means that we can just change `rids` to be a `map[string]bool` and then do something like:

```go
if !rids[newrid] {
    rids[newrid] = true
}
```

which is much easier to read.

As a side note, some people may prefer to use `map[string]struct{}` and do the normal `!ok` check since empty `struct{}`'s take up 0 bytes, but personally I like the readability of using a bool, and we're not storing a lot of items here so the cost is negligible, IMO.

imap/monitor.go
```diff
@@ -147,48 +151,91 @@ func checkForNewEmails(im models.IMAP) {
...
+	}
+}
+
+// returns a slice of gophish rid paramters found in the email HTML, Text, and attachments
+func matchEmail(em *email.Email) (map[string]int, error) {
+
+	rids := make(map[string]int)
+	checkRIDs(em, rids)
+
+	//Next check each attachment
```
jordan-wright, line 224: Small nit: if possible, we should try to have a space between the `//` and the comment (the one below also has this issue)

jordan-wright:
Hey @glennzw! This is looking great!

I've added a few more comments. Also I think there are some merge conflicts that we'll need to resolve before we can let the CI pipeline run which handles testing, making sure that `gofmt` has been run, etc.

jordan-wright:
I will say that one thing I really like as we're working through this is that we're actually steadily removing code, which is always a nice thing to see 😄 This PR will be awesome for Gophish.

imap/monitor.go
glennzw: I like this approach a lot, done!

glennzw:
Hi @jordan-wright ! Made those little changes, thanks for the review! Yeah I do like removing code rather than adding code :D

jordan-wright:
Hey @glennzw!

This LGTM! All that's left is to resolve the merge conflicts and I'll get this merged in.

Thanks for all the hard work on this! 💚

glennzw:
Great! I hope I did the merge conflict correctly;

```
git fetch upstream
git merge master
git push
```

You're very welcome, this has been a fun project to contribute to! 💚

glennzw:
> Great! I hope I did the merge conflict correctly;
> 
> ```
> git fetch upstream
> git merge master
> git push
> ```

Oh, actually, I don't think this is what you were asking me to do 😬

How do I resolve the conflicts?

jordan-wright (approved):

jordan-wright:
LGTM! Thanks for the great work and continued patience, @glennzw! 🎊

glennzw:
> LGTM! Thanks for the great work and continued patience, @glennzw! 🎊

Awesome! Thanks for all the detailed PR'ing - I learned a lot 🕺

PR #1812, opened 2020-04-11
Added ability to allow admin to 'su' to other accounts

glennzw:
## Scenario: 
As an admin you can create user accounts, but if those users change their passwords you're unable to authenticate to their accounts (e.g. for debugging / other assistance). The only current solution would be for the user to share their password, or for you as the admin to change their password.

## Solution:
This pull request adds a new button in the User Management section that allows admin users to immediately 'su' to that user account.

Screenshot 2020-04-11 at 16 08 17

### DIFF
--- a/controllers/route.go
+++ b/controllers/route.go
@@ -4,7 +4,6 @@ import (
 	"compress/gzip"
 	"context"
 	"crypto/tls"
-	"fmt"
 	"html/template"
 	"net/http"
 	"net/url"
@@ -180,8 +179,6 @@ func newTemplateParams(r *http.Request) templateParams {
 // Base handles the default path and template execution
 func (as *AdminServer) Base(w http.ResponseWriter, r *http.Request) {
 	params := newTemplateParams(r)
-
-	fmt.Println(params)
 	params.Title = "Dashboard"
 	getTemplate(w, "dashboard").ExecuteTemplate(w, "base", params)
 }
--- a/controllers/route.go
+++ b/controllers/route.go
@@ -4,6 +4,7 @@ import (
 	"compress/gzip"
 	"context"
 	"crypto/tls"
+	"fmt"
 	"html/template"
 	"net/http"
 	"net/url"
@@ -130,7 +131,7 @@ func (as *AdminServer) registerRoutes() {
 	router.HandleFunc("/settings", mid.Use(as.Settings, mid.RequireLogin))
 	router.HandleFunc("/users", mid.Use(as.UserManagement, mid.RequirePermission(models.PermissionModifySystem), mid.RequireLogin))
 	router.HandleFunc("/webhooks", mid.Use(as.Webhooks, mid.RequirePermission(models.PermissionModifySystem), mid.RequireLogin))
-	router.HandleFunc("/superlogin", mid.Use(as.SuperLogin, mid.RequirePermission(models.PermissionModifySystem), mid.RequireLogin))
+	router.HandleFunc("/impersonate", mid.Use(as.Impersonate, mid.RequirePermission(models.PermissionModifySystem), mid.RequireLogin))
 	// Create the API routes
 	api := api.NewServer(api.WithWorker(as.worker))
 	router.PathPrefix("/api/").Handler(api)
@@ -179,6 +180,8 @@ func newTemplateParams(r *http.Request) templateParams {
 // Base handles the default path and template execution
 func (as *AdminServer) Base(w http.ResponseWriter, r *http.Request) {
 	params := newTemplateParams(r)
+
+	fmt.Println(params)
 	params.Title = "Dashboard"
 	getTemplate(w, "dashboard").ExecuteTemplate(w, "base", params)
 }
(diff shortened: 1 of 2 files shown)


controllers/route.go
```diff
@@ -265,6 +266,23 @@ func (as *AdminServer) Webhooks(w http.ResponseWriter, r *http.Request) {
 	getTemplate(w, "webhooks").ExecuteTemplate(w, "base", params)
 }
 
+// SuperLogin allows an admin to login to a user account without needing the password
+func (as *AdminServer) SuperLogin(w http.ResponseWriter, r *http.Request) {
```
jordan-wright, line 270: Could we rename this to `Impersonate`?

controllers/route.go
```diff
@@ -265,6 +266,23 @@ func (as *AdminServer) Webhooks(w http.ResponseWriter, r *http.Request) {
...
+
+	if r.Method == "POST" {
+		username := r.FormValue("username")
+		u, err := models.GetUserByUsername(username)
+		if err != nil {
+			log.Error(err)
+		} else {
+			session := ctx.Get(r, "session").(*sessions.Session)
```
jordan-wright, line 278: We should probably return a flash message to indicate the operation was successful.

controllers/route.go
```diff
@@ -265,6 +266,23 @@ func (as *AdminServer) Webhooks(w http.ResponseWriter, r *http.Request) {
...
 }
 
+// SuperLogin allows an admin to login to a user account without needing the password
+func (as *AdminServer) SuperLogin(w http.ResponseWriter, r *http.Request) {
+
+	if r.Method == "POST" {
+		username := r.FormValue("username")
+		u, err := models.GetUserByUsername(username)
+		if err != nil {
+			log.Error(err)
```
jordan-wright, line 276: A better pattern is to abort early. E.g. something like:

```go
if err != nil {
    log.Error(err)
    http.Error(w, http.StatusNotFound, err.Error())
    return
}
session := ctx....
```

static/js/src/app/users.js
```diff
@@ -115,6 +115,35 @@ const deleteUser = (id) => {
...
+        type: "warning",
+        animation: false,
+        showCancelButton: true,
+        confirmButtonText: "Swap User",
+        confirmButtonColor: "#428bca",
+        reverseButtons: true,
+        allowOutsideClick: false,
+    }).then((result) => {
+        if (result.value) {
+          fetch('/superlogin', {
```
jordan-wright, line 135: Perhaps a different approach would be to have an HTML form that points to `/impersonate` in the table itself, and the modal just submits that form? I think we have a similar pattern elsewhere, but I could be wrong. Just something to consider, that way we're not having to encode the CSRF token like this 🤷 But if that's not feasible, then approach will work fine.

jordan-wright (changes requested):
Hey @glennzw!

Thanks for sending this in. I could see this being useful, and this approach seems fine. Perhaps in the future, we could offer a way to get back out of the impersonation without having to logout/login. But that's a future problem to solve 😄 

I made a few notes in the PR. If you could take a look, I'd appreciate it! Thanks again for taking the time to knock this out!

controllers/route.go
glennzw: Done!

jordan-wright: I don't show it as done- for some reason it still shows as `SuperLogin` to me.

glennzw: Ah sorry I got carried away marking it as resolved here, I resolved it locally - but will push all the fixes at once.

controllers/route.go
glennzw: I'm struggling to get the flash message right. I can see two options:

### JavaScript
After submitting the `/impersonate` form we stay on the `/users` page and flash a message at the top of the screen indicating success, and then redirect to the dashboard after a few seconds:

```
successFlash("Successfully changed user to  " + user.username + ". Redirecting to dashboard.")
setTimeout(function(){ window.location.href = "/" }, 2000);
```

### Via session Flash messages
Alternatively on the server side we add a flash message to be displayed after the redirect away from `/users` (say, the dashboard). But I can't seem to figure this out. In the `Impersonate` function in `route.go` I add:

`Flash(w, r, "success", "You have successfully changed user.")`

But this doesn't seem to be unpacked and displayed for any page except for `/login` (when indicating you've logged out, or that your user/pass were incorrect). So I *think* I'd need to add some of the same logic from `Login` to `Base` (unpacking the Flash message), but I haven't been able to get this to work yet.

Just thought I'd checkin to let you know where I'm at, in case I'm missing something, or that the simple JavaScript approach is fine.

glennzw: A third option would be to use a second popup message from the Sweetalert dialog indicating success or not, e.g:

Screenshot 2020-04-12 at 20 24 37

I think I prefer this option the best.

jordan-wright: I’m good with this!

static/js/src/app/users.js
glennzw: So I could use a form by putting this HTML in the users.html:

```
<!-- Form for Impersonate-->
<form id="form_imp" style="visibility: hidden" action="/impersonate" method="POST">
    <input name="username" type="text" value="monzo">
    <input id="csrf_imp" name="csrf_token" type="text">
    <input type="submit" value="Submit">
</form> 
```

and then in the users.js:

```
$("#user_imp").val(user.username)
$("#csrf_imp").val(csrf_token)
$("#form_imp").submit()
```

The problem with this would fall back to the above issue of figuring out how to display a message; if we perform a submit we'll load a whole new page before getting a change to pop up a dialogue, so I think we'll need to stick with the javascript approach.

controllers/route.go
glennzw: Good point, done.

controllers/route.go
glennzw: Super, I'll push that!

glennzw:
> Hey @glennzw!
> 
> Thanks for sending this in. I could see this being useful, and this approach seems fine. Perhaps in the future, we could offer a way to get back out of the impersonation without having to logout/login. But that's a future problem to solve 😄

Yeah I started off on a more complex idea of generating one-time login links that are stored in the DB and when clicked log in as the predetermined user and then expire - but it started getting complicated so I went with this easier solution for now :)

controllers/route.go
```diff
@@ -178,6 +180,8 @@ func newTemplateParams(r *http.Request) templateParams {
 // Base handles the default path and template execution
 func (as *AdminServer) Base(w http.ResponseWriter, r *http.Request) {
 	params := newTemplateParams(r)
+
+	fmt.Println(params)
```
jordan-wright, line 184: Looks like we can delete this debug line

glennzw: Ah whoops, good catch. Removed.

jordan-wright:
LGTM- thanks @glennzw!

Issue #1779, opened 2020-03-09
RFC: Enabling Additional Template Fields


Al-TSN:
**Executive Summary:**
The need to provide “extreme customization” when sending phishing simulation has become one of our top requested features.  This is due to the rise in spearphishing events, it is no longer enough to customize a message with the recipient’s name and possibly position.  Our users want to see messages that reflect the recipient’s organization, management structure, birthdate, etc.

This RFC will propose a solution for increasing the flexibility of GoPhish’s email template framework.  This solution will allow for utmost flexibility without constraining the system into a fixed number of templated fields.  Understanding that each environment, in fact each phishing campaign could require a different set of templated fields.  Therefore designing a feature with a fixed number of field would constrain growth, prevent reusability and become confusing to manage as templating requirements change.

**High Level Design:**
_Targets table_:
The targets table contains the recipients for each campaign, targets are reused as long as every first_name, last_name, email and position is unique.  If any value changes a new record in the targets table is created and associated with a specific group.

To accommodate the requirements for additional template fields, an approach that could be taken is to determine a list potential fields; manager, ceo, birthday, dollar amount, etc. adding those as columns.  As mentioned earlier, this would be difficult as it would lock every GoPhish user into a specific set of fields that may or not be appropriate.  Therefore, many people would start to repurpose fields, e.g. manager -> spouse, birthday -> start date.

An alternative solution would be to use generic field name, e.g. field1, field2, .. field20.  Again, this would have a fixed number of fields, in addition, every administrator would need to keep track of field mapping.  This could be rectified by having an additional table of field mapping to targets, this would require deeper development and an overall change to the DB structure.

**Proposed solution**, add an additional field to the targets table, it can be called extended_template. This would be a text field and the contents would be a JSON string with a key pair of template_field -> value.  This would allow every organization to define the templates fields important to their business case.  Additionally, the fields could be different for each template tied to a campaign. An example would be as follows:
{
	“ceo” : “Dani Smith”,
	“request” : “Amazon Gift Card”,
	“amount” : “$100.00”,
	“number” : “10”,
	“manager” : “Sarah Jane”
}

This is an example of the email template:
{{.first_name}}:

{{.ceo}} has asked me to procure {{.number}} {{.request}}s for {{.amount}} for a give a way at the next staff meeting.  I have set up a special link for you to place the order. Just click on here {{.url}} and use your company provided credit card.

Thanks for doing this.
{{.manager}}

[...]

PR #1813, opened 2020-04-16
adding extended_templates feature

ahmedkhalid1710:
more details can be found here: https://github.com/gophish/gophish/issues/1779
please review.. thanks

### DIFF
--- a/controllers/phish.go
+++ b/controllers/phish.go
@@ -265,11 +265,13 @@ func (ps *PhishingServer) PhishHandler(w http.ResponseWriter, r *http.Request) {
 // connection. This usually involves writing out the page HTML or redirecting
 // the user to the correct URL.
 func renderPhishResponse(w http.ResponseWriter, r *http.Request, ptx models.PhishingTemplateContext, p models.Page) {
+	// Getting all template params.. current + extended.
+	allTemplateParams := models.GetAllTemplateParams(ptx)
 	// If the request was a form submit and a redirect URL was specified, we
 	// should send the user to that URL
 	if r.Method == "POST" {
 		if p.RedirectURL != "" {
-			redirectURL, err := models.ExecuteTemplate(p.RedirectURL, ptx)
+			redirectURL, err := models.ExecuteTemplate(p.RedirectURL, allTemplateParams)
 			if err != nil {
 				log.Error(err)
 				http.NotFound(w, r)
@@ -280,7 +282,7 @@ func renderPhishResponse(w http.ResponseWriter, r *http.Request, ptx models.Phis
 		}
 	}
 	// Otherwise, we just need to write out the templated HTML
-	html, err := models.ExecuteTemplate(p.HTML, ptx)
+	html, err := models.ExecuteTemplate(p.HTML, allTemplateParams)
 	if err != nil {
 		log.Error(err)
 		http.NotFound(w, r)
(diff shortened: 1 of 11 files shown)


jordan-wright:
Thanks for sending this in! I'll get a chance to review ASAP.

jordan-wright:
Hey there!

My apologies for the delay in getting back to you. My time has been little and far between for PR reviews, and I wanted to think through this a little bit more.

I appreciate you sending in the first pass! This approach definitely works, but I'd like to propose some different ways of doing things that might give us a cleaner solution. Here are some things I'm concerned about with the existing approach:

* We remove the template validation checking and set the `.Option("missingkey=zero")` option. I get why you're choosing to do this, but it's quite likely that someone will accidentally mess up a well-known template variable. Having this be blanked out will almost certainly result in issues where people don't understand why part of their email is missing. I'd like to keep this validation in place.
* We introduce another layer for managing template parameters in `GetAllTemplateParams`, which adds complexity. Additionally, it creates a separate map that we have to maintain as the `BaseRecipient` structure changes. I'd like to avoid this if possible.

So here's the approach I'd recommend which _seems_ to work after some limited testing. The idea is that we create a type for our custom variables which acts like a map, but serializes correctly into the database as a byte slice.

First, in `groups.go` (or maybe a new `types.go`?) we define a JSON type which is an alias of a `map[string]string`. We then implement the `sql.Scanner` and `sql.Valuer` interfaces to handle the (de)serialization. This way it comes in and is used as a map, but saves to the database as a []byte. Something like this:

```go
type JSON map[string]string

// Value marshals the JSON payload into a []byte to be stored in the database
func (p JSON) Value() (driver.Value, error) {
	return json.Marshal(p)
}

// Scan implements the sql.Scanner interface, decoding the stored value from
// the database.
func (p *JSON) Scan(value interface{}) error {
	b, ok := value.([]byte)
	if !ok {
		return errors.New("JSON type assertion to []byte failed")
	}

	return json.Unmarshal(b, &p)
}
```

Then add that to the model:

```go
// BaseRecipient contains the fields for a single recipient. This is the base
// struct used in members of groups and campaign results.
type BaseRecipient struct {
	Email            string      `json:"email"`
	FirstName        string      `json:"first_name"`
	LastName         string      `json:"last_name"`
	Position         string      `json:"position"`
	ExtendedTemplate JSON `json:"extended_template,omitempty"`
}
```

Honestly, I think that's about it. Then, let's say I have a recipient with custom parameters that look like this:

```json
{ "custom": "Custom!" }
```

You can reference it in a template like this:

```
{{.ExtendedTemplate.custom}}
```

If this value doesn't exist, then Go by default sets it to the string `<no value>` which personally I think is reasonable.

The benefits are that we don't need the `GetAllTemplateParams`, we don't need to zero out template values using that option, and we can keep our existing validation checks to catch when, say, someone uses `{{.EtxendedTemplate}}`, catching those issues early.

You might give that a shot and let me know if it works for you! It seems to work just fine in my limited testing, but I'd like some feedback to make sure I'm not missing something.

Thanks again for taking the time to send this in!

Al-TSN:
Hey @jordan-wright - Thanks for the feedback, your suggestions look totally reasonable.  Unfortunately, priorities have shifted and we won't be able to finish this up in the near term.  Hopefully someone would be able to pickup the ball and run with it.  I believe this will be a great enhancement to the product.

Best,
Al

jordan-wright:
No worries @Al-TSN, thanks for the heads up! If I get some time then I’ll knock this out myself. It looks like all the pieces are there and things work, it’s just a matter or testing everything.

PR #1817, opened 2020-04-22
Add capability to run the binary in a mode

mitprasoon:
These commit includes changes to start the server as one of admin (also
IMAP) or phish server. Before this change the servers used to run in
monolith this change will decouple the two core component i.e. admin
and phish server so that they can be run independently.

This will help where admin and phish server are required to run
saperately e.g. phish server runs in a DMZ.

The available modes are `admin`, `phish` and `all`. Running the binary
in the `admin` mode will start the admin and IMAP server, while running
the binary in the `phish` mode will start the phish server. `all` mode,
which is also the default mode will start admin, IMAP and phish servers.
e.g. `go run gophish.go --mode admin`

### DIFF
--- a/gophish.go
+++ b/gophish.go
@@ -26,6 +26,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/signal"
@@ -40,9 +41,17 @@ import (
 	"github.com/gophish/gophish/models"
 )
 
+const (
+	modeAll   string = "all"
+	modeAdmin string = "admin"
+	modePhish string = "phish"
+)
+
 var (
 	configPath    = kingpin.Flag("config", "Location of config.json.").Default("./config.json").String()
 	disableMailer = kingpin.Flag("disable-mailer", "Disable the mailer (for use with multi-system deployments)").Bool()
+	mode          = kingpin.Flag("mode", fmt.Sprintf("Run the binary in one of the modes (%s, %s or %s)", modeAll, modeAdmin, modePhish)).
+			Default("all").Enum(modeAll, modeAdmin, modePhish)
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


gophish.go
```diff
@@ -102,18 +111,27 @@ func main() {
 	phishServer := controllers.NewPhishingServer(phishConfig)
 
 	imapMonitor := imap.NewMonitor()
-
-	go adminServer.Start()
-	go phishServer.Start()
-	go imapMonitor.Start()
+	if *mode == "admin" || *mode == "all" {
+		log.Infof("Stating the admin and IMAP server, server mode is %s", *mode)
```
jordan-wright, line 115: Can we please remove this log line? It's not required.

gophish.go
```diff
@@ -102,18 +111,27 @@ func main() {
...
+		go adminServer.Start()
+		go imapMonitor.Start()
+	}
+	if *mode == "phish" || *mode == "all" {
+		log.Infof("Stating the phish server, server mode is %s", *mode)
```
jordan-wright, line 120: Can we please remove this log line? It's not required.

jordan-wright (changes requested):
Hi there!

Thanks for sending this in. I can see the utility of having a command-line parameter like this. I'm not in love with the `--mode=phish` kind of API, but looking around I didn't immediately find other software which use a different pattern to compare against, so this will be fine for now.

I've added a couple of small comments - please remove those log lines and I'll merge this in.

Thanks again!

gophish.go
mitprasoon: Removed

gophish.go
mitprasoon: Removed

mitprasoon:
Hello @jordan-wright ,

I have addressed your review comments. 

Appreciate your time!

jordan-wright (approved):

edermi:
Just out of curiosity, can you elaborate more on the use case? How does an admin server communicate with a separate phish server? Or does this implicitly require to share the database between the two systems?

Also, some documentation regarding this feature may be good for anybody not following the development closely.

jordan-wright:
Documentation is forthcoming, but yes, the admin and phish servers will have to share a database.

PR #1830, opened 2020-05-03
Load datatable rows all at once

mitprasoon:
The javascripts code which load datatables were drawing the table rows
one at a time. This operation is time intensive and caused the tables
with a couple of thousand (e.g. ~5000 targets associated with a group)
entries to take upto 150 seconds to load. This also caused the pages to
timeout while waiting for data tables to load. This issue has been fixed
by drawing all rows at once.

### DIFF
--- a/static/js/src/app/campaigns.js
+++ b/static/js/src/app/campaigns.js
@@ -360,6 +360,7 @@ $(document).ready(function () {
                         [1, "desc"]
                     ]
                 });
+                campaignRows = []
                 $.each(campaigns, function (i, campaign) {
                     campaignTable = campaignTableOriginal
                     if (campaign.status === "Completed") {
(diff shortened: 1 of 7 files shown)


jordan-wright (changes requested):
Hey there!

Thanks for sending this in. I’m excited to give this a shot.

In the meantime, to start getting things ready to merge, could you please remove the .min.js files from this PR, as well as any console.log debug lines you’ve added?

I’ll give this a shot ASAP. Thanks again!

mitprasoon:
Hello @jordan-wright ,

I have removed minified JS from the PR, also I have removed the console.log statement from groups.js

jordan-wright:
I'll be honest- I don't see much of a performance difference on my Mac, but it does seem like it'd be preferred to only draw the table once vs. drawing it for every row.

Thanks for sending this in!

PR #1838, opened 2020-05-15
Use GroupsSummary to create and copy campaign

mitprasoon:
The Groups (get all groups and associated targets) call is used while
loading modal for creating and copying a campaign. As Groups API gets
all the associated targets for a groups as well, it slows system
considerably if there are large number of groups and targets (~200
groups each with ~100-10000 targets).
As targets are not really needed in this workflow, this call can be
replaced by GroupsSummary call.

### DIFF
--- a/static/js/src/app/campaigns.js
+++ b/static/js/src/app/campaigns.js
@@ -162,15 +162,16 @@ function deleteCampaign(idx) {
 }
 
 function setupOptions() {
-    api.groups.get()
-        .success(function (groups) {
+    api.groups.summary()
+        .success(function (summaries) {
+            groups = summaries.groups
             if (groups.length == 0) {
                 modalError("No groups found!")
                 return false;
             } else {
                 var group_s2 = $.map(groups, function (obj) {
                     obj.text = obj.name
-                    obj.title = obj.targets.length + " targets"
+                    obj.title = obj.num_targets + " targets"
                     return obj
                 });
                 console.log(group_s2)


static/js/src/app/campaigns.js
```diff
@@ -162,8 +162,9 @@ function deleteCampaign(idx) {
 }
 
 function setupOptions() {
-    api.groups.get()
-        .success(function (groups) {
+    api.groups.summary()
+        .success(function (groupsSummary) {
```
jordan-wright, line 166:
```suggestion
        .success(function (summaries) {
```

static/js/src/app/campaigns.js
```diff
@@ -162,8 +162,9 @@ function deleteCampaign(idx) {
...
+            groups = groupSummary.groups
```
jordan-wright, line 167:
```suggestion
            groups = summaries.groups
```

jordan-wright (changes requested):
Hi there!

I like the idea behind this change, but as it stands this doesn't work for two reasons:
* A mismatch between `groupSummary` and `groupsSummary`- I've added suggestions to rename to `summaries` to be a bit cleaner
* A bit later, we call `obj.targets.length`. There is no `targets` attribute on a group summary. Instead, we need to change this to `obj.num_targets`.

After those changes, this seems to work as expected. If you could knock those out, then I'll get this merged.

static/js/src/app/campaigns.js
mitprasoon: Addressed

static/js/src/app/campaigns.js
mitprasoon: Addressed

mitprasoon:
Hi @jordan-wright ,

I have addressed your suggestions. 

Thanks for your time and feedback!

jordan-wright:
LGTM. Thanks @mitprasoon!

Issue #1587, opened 2019-09-13
Discussion: Changing GoPhish Architecture / Simulation Library


glennzw:
I find that the way landing pages/email templates/sending profiles is structured is a little counter-intuitive and disconnected. I think these should be collapsed into one 'thing', of which the name I've been using is 'simulation' or 'phishing simulation'. If I want to send, say, a DropBox phish it's going to be a DropBox email, DropBox landing page, and sent from a dropbox looking email. I don't see the benefit of having these three things separate and then having to match them up when sending a campaign.

At the least, I'd favour collapsing the sender email address/name from Sending Profile into Email Template. I think a Sending Profile should just be the SMTP details. I have one SMTP server, and it feels redundant having to create 10 Sending Profiles with identical SMTP settings for 10 different phishing templates.

It's also a little hard to get a feel for what a campaign will look like; I need to manually go to three different tabs and click edit buttons and source buttons to see what a campaign will actually look like.

To illustrate how I envisage my proposal I've created a 'Simulation Library' tab in GoPhish:

<img width="1438" alt="simLib" src="https://user-images.githubusercontent.com/3966613/64864272-a82ef480-d62e-11e9-9d25-2a3a74f5d7b4.png">

The library aggregates Email Templates, Landing Pages, and Sending Profiles that have the same *name* into cards in the Simulation Library tab. The images are loaded from `/endpoint/library/<NAME>/landing.png` and `/endpoint/library/<NAME>/email.png` (if you click the image it'll flip between a picture of the landing page and the email).

Clicking `Test` allows you to send the Simulation to a single email address:  

<img width="845" alt="testSim" src="https://user-images.githubusercontent.com/3966613/64864561-4de26380-d62f-11e9-9fab-e1ee96918791.png">

Clicking `Launch` leads you through a minimal Wizard to get the campaign launched:  

<img width="1115" alt="simWiz" src="https://user-images.githubusercontent.com/3966613/64864939-1c1dcc80-d630-11e9-806d-0e16e01dedbe.png">

Here's a demo video: [GoPhish Simulation Library Demo](https://www.youtube.com/watch?v=bBLRNKTm7w8)

I've also restructured the menu on the left. The `GoPhish Main` menu is your high level options; your dashboard, simulation library, and staff (targets and groups). This section is for day to day phishing campaign management, and many users would only spend their time here.

The `Build Simulations` menu is for 'power users' who want to add / manage simulations, and then settings and help at the bottom.

[...]

PR #1846, opened 2020-05-22
New 'Redirect to' functionality was added

LaggerP:
Now **gophish** is able to redirect to a _second url_ if the Landing Page creator want it.

The functionality works as follows:

1. the victim send data in the Landing Page form.

2. gophish redirect to the first redirect url.

3. now we got all data from the victim... but if the victim send data a second time, gophish redirect to the second url.

**This is very usefull for empresarial campaigns when they want to raise awareness about personal data safety.**

Example:

Current flow:
Form Landing Page > send data > redirect first url 

New flow:
Form Landing Page > send data > redirect first url > _(opcional*)_ send data for a second time > redirect to second url.

*is opcional because if the form landing creator let the second url input empty, gophish ignore this.

image


jordan-wright:
Hi there!

Thanks for sending this in. Instead of adding more redirects, I'd rather find time to rethink how we handle building campaign flows. There has been some design work around this in #223 and #1587. I don't have a timeline on when these would get done, but I'd rather focus effort there.

That said, I'm going to go ahead and close this one out. I do appreciate you sending this in, though! Please keep the ideas coming!

kevlar-ashu:
Hey @LaggerP,
In mine case second redirect isn't working. Getting HTTP ERROR 405 and it's saying this page isn't working right now.
After debugging what's come in my mind is that renderPhishResponse got called twice. One for default landing page html and second is for first redirect url. So, for the second redirect url it's not getting called.

PR #1883, opened 2020-07-01
Add ability to provide initial API key on service standup

stusmall:
This commit follows the same pattern used by GOPHISH_INITIAL_ADMIN_PASSWORD but lets you set the initial API key.  This doesn't change the security model any since someone could still automate their way to an API key with the initial password, this just makes the process easy and more reliable.

I'm working in an environment where I want to completely automate everything around gophish.  The service standup, campaigns, etc.  This change lets me do this and stick only to documented APIs.


jordan-wright:
Works for me! Thanks for sending this in.

Issue #1851, opened 2020-05-30
Display of reported non-campaign emails (request for comment)


glennzw:
I added the ability to display reported non-campaign emails on top of my [recent IMAP improvements](https://github.com/gophish/gophish/pull/1791). Any emails emails forwarded to your IMAP account that are not from gophish, and therefore require further investigation, would appear here. The emails are imported into the gophish db, as opposed to just being fetched on the fly and displayed. This is what it looks like:

![Screenshot 2020-05-30 at 19 48 41](https://user-images.githubusercontent.com/3966613/83336835-966daa00-a2ae-11ea-8e5f-3f62fa277abd.png)

Attachments can be viewed via the dropdown:

![Screenshot 2020-05-30 at 19 39 15](https://user-images.githubusercontent.com/3966613/83336686-44785480-a2ad-11ea-9aed-adef7a1bebe5.png)

The user can set the 'Status' of an email after evaluating it:

![Screenshot 2020-05-30 at 19 39 57](https://user-images.githubusercontent.com/3966613/83336694-5ce86f00-a2ad-11ea-9674-51a9cd999a7c.png)

...as well as adding manual notes:

![Screenshot 2020-05-30 at 19 41 14](https://user-images.githubusercontent.com/3966613/83336710-8b664a00-a2ad-11ea-9740-7f142bc62301.png)

The view icon loads the email, giving the option of viewing plaintext or HTML renderings (HTML rendered inside a sandboxed iframe):

![Screenshot 2020-05-30 at 19 41 48](https://user-images.githubusercontent.com/3966613/83336723-b6509e00-a2ad-11ea-8554-0f098c73ea26.png)

I've had quite a few requests for this feature. Do others find it useful? Any other feature requests? I'm thinking of adding support to automatically triage emails via open source feeds / APIs. I think it'd probably make more sense to de-couple that out of gophish, and have another standalone application that queries emails via the API and updates notes / status on detected threats (rather than rolling in the tooling into gophish itself).

I'll also add a toggle button to enable/disable this functionality in the settings, as well toggling the display of the 'Reported Emails' menu item.

jordan-wright:
Hey @glennzw!

Thanks for sending this in. I really like this! It looks really similar to what I built with isthislegit. https://github.com/duo-labs/isthislegit

I’m personally just fine integrating this into Gophish. While it’s tangential to the core function, I think it adds significant value and making part of the single binary would be really nice.

Want to make a PR and we can start working through it?

[...]

PR #1894, opened 2020-07-13
Display of reported non-campaign emails

glennzw:
As per my RFC https://github.com/gophish/gophish/issues/1851 here is the pull request to add functionality to display emails that users have reported that are **not** gophish campaigns. 

I did this work a while ago, and merged it into the current gophish master today before creating this pull request; I hope I've done it correctly. Thanks for the patience with me getting around to this, it's sometimes hard to find the time for side projects!

I'm repeating myself, but to save people reading the RFC below are the details on this pull request.

----
I added the ability to display reported non-campaign emails on top of my [recent IMAP improvements](https://github.com/gophish/gophish/pull/1791). Any emails forwarded to your IMAP account that are not from gophish, and therefore require further investigation, would appear here. The emails are imported into the gophish db, as opposed to just being fetched on the fly and displayed. This is what it looks like:

Screenshot 2020-05-30 at 19 48 41

Attachments can be viewed via the dropdown:

Screenshot 2020-05-30 at 19 39 15

The user can set the 'Status' of an email after evaluating it:

Screenshot 2020-05-30 at 19 39 57

...as well as adding manual notes:

Screenshot 2020-05-30 at 19 41 14

The view icon loads the email, giving the option of viewing plaintext or HTML renderings (HTML rendered inside a sandboxed iframe):

Screenshot 2020-05-30 at 19 41 48

I've had quite a few requests for this feature. Do others find it useful? Any other feature requests? I'm thinking of adding support to automatically triage emails via open source feeds / APIs. I think it'd probably make more sense to de-couple that out of gophish, and have another standalone application that queries emails via the API and updates notes / status on detected threats (rather than rolling in the tooling into gophish itself).

I'll also add a toggle button to enable/disable this functionality in the settings, as well toggling the display of the 'Reported Emails' menu item.


glennzw:
A few things I still need to consider/do:

- [ ] Test MySQL support
- [ ] Remove debug statements (printing out SQL queries)
- [ ] Column sorting doesn't persist between auto refreshes
- [ ] Add an Enabled tickbox in the IMAP settings to enable/disable this feature (is this necessary? Maybe for privacy concerns people wouldn't want emails auto imported into gophish?)
- [ ] Domain filter (to ignore emails not from @companydomain.com)
- [ ] Inline images in reported emails don't seem to load in reported emails

controllers/api/reported.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"net/http"
+	"strconv"
+	"strings"
+
+	ctx "github.com/gophish/gophish/context"
+	"github.com/gophish/gophish/models"
```
vasugoyal241, line 11: Ghb

glennzw: Sorry I'm not sure what Ghb means?

PR #1902, opened 2020-07-20
Submit credentials to original origin after capturing them (#703)

vasugoyal241:
This commit implements the ability to submit credentials upstream after they are submitted to Gophish.ghpbhv


modestoartur:
This is a really good idea to merge!

analyserdmz:
This could evolve as a credentials-checker, with the ability to not log invalid credentials and keep only the valid ones. Just throwing an idea here.

Issue #1756, opened 2020-02-24
[Feature Request] Add option to disable certificate validation for IMAP


slegna-oryp:
What version of Gophish are you using?: 0.9.0 inside Docker

Brief description of the issue: 
When I try to add an account from our IMAP server I receive the following error: `x509: certificate signed by unknown authority` when testing the connection.

I don't think my root CA is bundled in the Debian truststore, which I also can't seem to add inside the container as a workaround. The option to add certificates to the truststore, or the ability to disable certificate validation would enable us to use the IMAP reporting feature.

glennzw:
I'm not sure if I want to add support for this at the moment as we're going to swap out the IMAP library soon, but I did create a branch for you that removes certificate validation:

https://github.com/glennzw/gophish/tree/AllowInsecureCerts

MatteoManzoni:
I had the same problem, resolved including ca-certificates package in the dockerfile

glennzw:
@MatteoManzoni @slegna-oryp 

This feature should be rolled out shortly: https://github.com/gophish/gophish/pull/1909

glennzw:
FYI this feature has been merged into master. Closing this out.

PR #1909, opened 2020-07-24
Added support to allow invalid IMAP certificates

glennzw:
A simple one, based on [requests](https://github.com/gophish/gophish/issues/1756). I've added a tickbox to allow users to permit invalid IMAP certificates.

Current behaviour, connection will fail when an invalid certificate is presented:

Screenshot 2020-07-24 at 18 19 41

But ticking the box, and hey presto:

Screenshot 2020-07-24 at 18 19 26

For testing purpose I created a certificate using openssl and created a socat tunnel from my localhost to a valid IMAP server:

```
openssl req -new -x509 -keyout test.key -out test.crt -nodes # Generate cert
cat test.key test.crt > test.pem

socat -v openssl-listen:993,cert=test.pem,verify=0,reuseaddr,fork ssl:imap.outlook.com:993,verify=0
```


db/db_mysql/migrations/20200116000000_0.9.0_imap.sql
```diff
@@ -1,7 +1,7 @@
 
 -- +goose Up
 -- SQL in section 'Up' is executed when this migration is applied
-CREATE TABLE IF NOT EXISTS `imap` (user_id bigint,host varchar(255),port int,username varchar(255),password varchar(255),modified_date datetime,tls boolean,enabled boolean,folder varchar(255),restrict_domain varchar(255),delete_reported_campaign_email boolean,last_login datetime,imap_freq int);
+CREATE TABLE IF NOT EXISTS `imap` (user_id bigint,host varchar(255),port int,username varchar(255),password varchar(255),modified_date datetime,tls boolean,ignore_cert_errors boolean,enabled boolean,folder varchar(255),restrict_domain varchar(255),delete_reported_campaign_email boolean,last_login datetime,imap_freq int);
```
jordan-wright, line 4: Can we create a new migration for these?

jordan-wright (changes requested):
In general, this LGTM! Thanks @glennzw!

The only request I'd have is to move the SQL modifications into new migration files since there are almost certainly folks who have already run this migration.

db/db_mysql/migrations/20200116000000_0.9.0_imap.sql
glennzw: Ahhh yes of course. Fixed, I think.

jordan-wright: It looks like this isn't fixed. Ideally, we would just have two brand new migration files - one for MySQL and one for SQLite. Both of these would just alter the existing table.

Right now, we have an `ALTER` migration for MySQL, but this migration here still exists, and the one for SQLite still does a full `CREATE TABLE` instead of just altering the existing one.

glennzw: Apologies Jordan, I definitely made the alterations as you describe, I think I must have botched my stashes and commits. Pushed again 🤞

jordan-wright:
Ok, promise not to get mad, but there's one last little tweak 😬 

Due to the way that filenames are resolved for migrations, there needs to be another "0" added to each of the migration files before the version number. For example:

`2020073000000_0.11.0_imap_ignore_cert_errors.sql`
should be
`20200730000000_0.11.0_imap_ignore_cert_errors.sql`

(Notice the extra trailing 0 before the `_`)

Otherwise, Gophish doesn't parse the time correctly and doesn't apply the migration.

When that's done, this should be good to merge 🎊

glennzw:
Not mad at all, I honestly love your eagle eyes and high QA levels! Renamed and pushed.

jordan-wright:
Great! This LGTM. Thanks for sending this in @glennzw.

As always, thanks for the great work! 💚

PR #1914, opened 2020-07-29
Update credentials in Readme.

xtrasimplicity:
The credentials were listed as admin/gophish, but this changed in v0.10.1, as per: https://docs.getgophish.com/user-guide/getting-started


jordan-wright:
LGTM! Thanks for sending this in. I was so focused on making sure the user guide was correct that I forgot we mention this in the README.

PR #1929, opened 2020-08-08
Support for custom events

glennzw:
This pull requests adds support for arbitrary campaign events. For example, you could craft an event of **'Secondary Link Clicked'** or **'Word Document Opened'** or **'Watched Training Video'** or **'Plugged in USB drive'** or anything else you like. You can also customise your events in terms of icon, color, label, and optional sub heading lines and icons. For example:

Screenshot 2020-08-08 at 22 16 54

There's nothing to setup server side for creating these new events, they are dynamically created by you client side with the new `/arbevent` endpoint. Obviously you'd need to insert them in the appropriate places. Here are the URLs that created the events in the above screenshot:

### Word Document Opened
`http://localhost/arbevent?rid=aIakOWF&title=Word Document Opened&label=label-clicked&icon=fa-file-word-o&color=%23F39C12&ua=0`

Breaking that URL down, we have the following parameters:

rid - the unique campaign RID
title - the title of the event to display in the event list and status
label - label color for the status
icon - the icon to use in the event list
color - color to use in the timeline graph and donuts
ua - either 0 or 1, whether or not to display the user agent. Defaults to 0.

### Word Macros Enabled
```http://localhost/arbevent?rid=aIakOWF&title=Word Macros Emabled&label=label-danger&icon=fa-file-word-o&color=%23ff0000&sub_text=Username: Bob&sub_icon=fa fa-user&sub_text=PC Name: DESKTOP-4C&sub_icon=fa fa-desktop```

Having a look at this URL, we have the extra parameters to display the text and icons under the main event heading:

sub_text - the sub text 
sub_icon - the sub icon

You can pass multiple values using the same parameter name, in the order you want them to appear (as is done in the above example). They are assembled into a pair of arrays server side, one for the text one for the icons.

  <hr>
  
As for the pie chart donuts, the regular 5 donuts will show, unless any arbitrary events have been submitted. Then each arbitrary event gets its own donut. For example:

Screenshot 2020-08-08 at 22 03 52

The arbitrary donuts are grouped into rows of 5 to match the original 5.

## Under the hood
Events are stored with a message of "Arbitrary Event" in the database to distinguish them from other events. Their URL payload is parsed to render them. For example, here's the database entry for the **'Word Document Opened'** event:

```
[...]

Issue #1667, opened 2019-11-18
Open File Tracking


bhaizlett123:
Hello,

I wanted to see if there is a way to track open files. Like if i include a word document in the email, i would like to see how many people actually opened the word doc. I know i could probably inject some DDE to open a url, but is there a way to configure gophish to track that?

bhaizlett123:
Just to add even if there is a way to append the rid parameter to something so when they open the file it tries to connect to that url and i can track that they opened the file?

jordan-wright:
Hi there!

I'm afraid that Gophish doesn't support attachment tracking at this time, and I don't have an ETA on when this support would be added.

Sorry we couldn't be of more help!

Issue #1606, opened 2019-09-26
New feature: track users who open an attachment


jrmarques666:
A great new feature would be an option to attach some file that when opened in a windows machine it connects to gophish and the action is recorded associated to the recipient id. Obviously this attachment must be customised for each recipient. Maybe we can use a Macro in a Word document. If so, we could also allow customisation of the text so to avoid suspicion.

jrmarques666:
I see this is a duplicate of https://github.com/gophish/gophish/issues/227. Sorry about that.

glennzw:
I too hadn't noticed that thread, and added my $0.02 to it.

jordan-wright:
Thanks for sending this in! This is indeed a duplicate of #227. Let's revisit that one soon and see if we can't build something into Gophish in line with @glennzw's suggestions.

Issue #1205, opened 2018-09-19
GoPhish URL in Word Docs, PDFs, Excel Docs etc.


W00glin:
First of all I just want to say thank you for such an incredible product. This is a truly excellent piece of software and I thoroughly enjoy using it.

I am not having an issue I just have a general inquiry about using GoPhish phishing links in word documents, excel spreadsheets pdfs etc. Essentially what I would like to do is take the phishing URL that is sent to users and then embed it into a document of some sort. I understand how to do this but what I am having issues with how to determine the correct URL to use. Unless I misunderstand and correct me if I am wrong but doesn't GoPhish generate a unique URL for each person the phishing email is being sent too? If that is the case is there a way for the GoPhish admins to determine that URL(s) and use them in other places? If it is not, could this be a consideration for future releases?

snooze6:
I agree that it would be useful to know which URL is sent to each user or even create new URLs without actually sending mails in order to create new attacks like @W00glin is suggesting.

jordan-wright:
Hey everyone!

I agree completely. Let me see what I can do. Right now, I figure I might have a checkbox when building a campaign to avoid sending emails. Then, I'll make sure that the results export will include the URL.

W00glin:
> 
> 
> Hey everyone!
> 
> I agree completely. Let me see what I can do. Right now, I figure I might have a checkbox when building a campaign to avoid sending emails. Then, I'll make sure that the results export will include the URL.

That would be excellent. Keep up the great work @jordan-wright

oded86w:
It's can be great to generate camping without send email .
And get link that not assign to email address .
And put the link in PDF/Word or maybe in desktop.ini , So we can put the files in DOK , And do a test of USB dropping .

PR #1936, opened 2020-08-15
Added support for templating attachments

glennzw:
This pull request adds support to apply template variables to attachments. This has two main use cases:
1. **Customize attachment** contents to targets e.g "Dear Glenn, this document is all about you."
2. The ability to **add tracking tokens to attachments** e.g. tracking pixel to a Word or PowerPoint document to tell when it's been opened, or of even more interest, if the user enabled macros when prompted.

There's no change to the front end. A user can simply choose to include the regular gophish template variables in certain files. I've included logic to only process a file once if no template variables are found.

I've used a switch statement to handle different attachment filetypes. At this stage I added support for:
 * Plain text / HTML documents: Just apply the`ExecuteTemplate() function`
 * Office documents (.docx, .xlsx, .pptx, .docm, .xlsm): Unarchive the file, and apply the `ExecuteTemplate` to every .xml or .rels file inside the archive, then rearchive those files.
 * Default: I chose to do nothing, but we could choose to apply the `ExecuteTemplate()` to all files that don't match the switch. This might break binary files though.

I've tested all the above file types and it works super well. The most comprehensive example is a macro enabled Word document that has a {{.TrackingURL}} as a linked image in the header/footer, and the {{.URL}} in a variable (e.g textbox) that's called from macro code. The result is that we can tell: 
 
1. Email was sent
2. Word document was opened (in this case 'Email opened')
3. Word macros enabled (in this case 'Link clicked') 

This will plug in nicely to my other pull request for custom events https://github.com/gophish/gophish/pull/1929 so we can get "Word Document opened" and "Macros enabled"

<hr>

## Notes on attachments

I found that in Office documents the curly braces in template variables weren't always being represented conveniently in the compressed XML. If you're having trouble open the `.docx` in vim (or decompress) and have a look in `word/document.xml` to make sure the literal `{{.FirstName}}` etc is appearing.

For macros, as noted in[ this previous attempt](https://github.com/gophish/gophish/pull/260) from some time ago, `.docm` files compile their macro code to a binary format. However the rest of the document is still just a compressed bunch of XML files. e.g:

```
" zip.vim version v28
" Browsing zipfile doc-with-macro.docm
" Select a file with cursor and press ENTER

[Content_Types].xml
_rels/.rels
word/_rels/document.xml.rels
word/document.xml
word/_rels/vbaProject.bin.rels
word/vbaProject.bin
word/media/image1.jpg
word/theme/theme1.xml
word/settings.xml
word/vbaData.xml
customXml/_rels/item1.xml.rels
customXml/itemProps1.xml
customXml/item1.xml
docProps/app.xml
docProps/core.xml
word/styles.xml
word/fontTable.xml
word/webSettings.xml
```

The macro code is compiled into the above `word/vbaProject.bin` file. We have two solutions here; either patch the binary with some placeholder variables string (e.g `url="AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"`) or place the URL variables in the Word document, and have them referenced by the macro code. 

The latter approach worked beautifully in my testing. Create a textbox called `urlbox` with contents `{{.URL}}` and the following macro code will hit that URL when the document opens:

```
Sub urlfetch()
 Dim shp As Shape

 For Each shp In ActiveDocument.Shapes
    If shp.Name = "urlbox" Then
        URL = shp.TextFrame.TextRange.Text
             Documents.Open (URL)
    End If
 Next
End Sub

Public Sub AutoOpen()
    urlfetch
End Sub
```

This is just PoC, it's a little kludgy as the webpage in `{{.URL}}` will be opened in a new Word window, but it's the quickest cross platform solution. A better option is the `Application.Documents.CanCheckOut (URL)` call but this sends a HEAD request (perhaps we can allow my new custom events to receive HEAD requests). My actual campaign Documents will include macro code to change the document into a 'gotcha' page, letting the user know that this was a campaign.

I can share my test files if needed: 

Screenshot 2020-08-15 at 17 40 18

Or add them to `https://github.com/gophish/payloads`

<hr>

The ability to track Office documents seems to be one of those most requested features (https://github.com/gophish/gophish/issues/880, https://github.com/gophish/gophish/issues/513, https://github.com/gophish/gophish/issues/352, https://github.com/gophish/gophish/pull/260, https://github.com/gophish/gophish/issues/227, https://github.com/gophish/gophish/issues/1667, https://github.com/gophish/gophish/issues/1606, https://github.com/gophish/gophish/issues/1205, https://github.com/gophish/gophish/issues/1176, https://github.com/gophish/gophish/issues/880, https://github.com/gophish/gophish/issues/352). I hope my solution solves this!


models/maillog.go
```diff
@@ -27,6 +32,10 @@ var MaxSendAttempts = 8
 // MailLog is exceeded.
 var ErrMaxSendAttempts = errors.New("max send attempts exceeded")
 
+// processAttachment is used to to keep track of which email attachments have templated values.
+// This allows us to skip re-templating attach
+var processAttachment = map[[20]byte]bool{} // Considered using attachmentLookup[campaignid][filehash] but given the low number of files current approach should be fine
```
jordan-wright, line 37: I'd prefer to err away from using a global map as a cache. While it wouldn't be immediately an issue, it takes up memory forever, not just when a campaign is being sent out.

Instead, just thinking out loud, but what we might be able to do is to modify the `models.Attachment` object to have an unexported `templateApplied` boolean field or something to indicate whether or not the templating has been applied to file already. This would be false by default then, after the `applyAttachmentTemplate` function has completed, we can set the flag so the attachment isn't processed again.

Additionally, I'd prefer to have `applyAttachmentTemplate` not be in `maillog.go`, but rather have it be a function off of an `Attachment` in `attachment.go`.

Ideally, it'd be something that kinda looks like this (I haven't tested this!)

in `models/attachment.go`:

```go
// Attachment contains the fields and methods for
// an email attachment
type Attachment struct {
	Id         int64  `json:"-"`
	TemplateId int64  `json:"-"`
	Content    string `json:"content"`
	Type       string `json:"type"`
	Name       string `json:"name"`

        templateApplied bool
}
...
func (a *Attachment) applyTemplate(ptx PhishingTemplateContext) (io.Reader, error) {
    if a.templateApplied {
        return strings.NewReader(a.Content), nil
    }
    // Do all your normal processing
    ...
    a.Content = processedAttachment
    a.templateApplied = true
    return strings.NewReader(a.Content), nil
}
```

Then, back in `maillog.go`, the call to `Attach` would look something like:

```go
msg.Attach(func(a *Attachment) (string, gomail.FileSetting, gomail.FileSetting) {
	h := map[string][]string{"Content-ID": {fmt.Sprintf("<%s>", a.Name)}}
	return a.Name, gomail.SetCopyFunc(func(w io.Writer) error {
		content := a.ApplyTemplate(ptx)
		_, err = io.Copy(w, content)
		return err
	}), gomail.SetHeader(h)
}(&a))
```

I _think_ something like this would work and would be cleaner, but give it a shot and let me know if you run into issues.

jordan-wright (changes requested):
Hey @glennzw!

Thanks for sending this in. I like this general approach of letting folks make their own attachments and Gophish will just be responsible for trying to apply the right template, rather than having Gophish try to be super smart about generating special files.

I've added some comments, but here are other TODO's that we'll need to consider for this:
* **Writing Tests** - I can probably help with this, but if you'd like to take a first pass, please go ahead. In general, we'll want to test at least one templated file, one non-templated file, multiple emails with the same attachment (to make sure we don't double-template it), etc.
* **Documentation** - Documentation is where this will shine. We want to be really clear about what we support, what we don't, what users need to do, and what Gophish will do for them. We'll undoubtedly get a **bunch** of questions about how to use this, so we'll want to have good answers pre-written to save ourselves some work.

Keep up the great work! And I'm thrilled to see this being done on a branch, since it means I can easily add commits for things like tests.

glennzw:
Awesome, I'm excited to make this a reality!

> * **Writing Tests** - I can probably help with this, but if you'd like to take a first pass, please go ahead. In general, we'll want to test at least one templated file, one non-templated file, multiple emails with the same attachment (to make sure we don't double-template it), etc.

For the test writing are you thinking automated code, or just creating a bunch of attachments and sending out campaigns manually to test them? I've done a fair bit of testing with the following files:

```
allvars.txt
doc-with-macro.docm
doc-with-template-vars.docx
doc-without-template-vars.docx
excel-with-macro.xlsm
excel-with-vars.xlsx
excel-without-vars.xlsx
html-file-with-template-vars.html
html-file-without-template-vars.html
powerpoint-with-vars.pptx
text-file-with-template-vars.txt
text-file-without-template-vars.txt
```

Basically each supported type with and without variables, being sent to multiple targets

One thing to note is that unlike writing an email or landing page template the user will only find out they used a `{{.BadVariable}}` when the campaign is launched, but the error is handled gracefully:

Screenshot 2020-08-16 at 13 18 06


> * **Documentation** - Documentation is where this will shine. We want to be really clear about what we support, what we don't, what users need to do, and what Gophish will do for them. We'll undoubtedly get a **bunch** of questions about how to use this, so we'll want to have good answers pre-written to save ourselves some work.

Definitely agree! I can get cracking on this, and addressing your notes below.

models/maillog.go
```diff
@@ -253,6 +267,129 @@ func (m *MailLog) Generate(msg *gomail.Message) error {
...
+		case ".exe":
+			// Todo. Perhaps we ignore the .exe and build our own, with a simple callback to the server
+			//  A special extension of 'exef' or some such might be useful in case users want to attach
+			//  an actual exe file. Does anyone email exe files in 2020 ?
+		default:
+			// We have two options here; either apply template to all files, or none. Probably safer to err on the side of none.
+			processedAttachment = string(decodedAttachment) // Option one: Do nothing
+			//processedAttachment, err = ExecuteTemplate(string(decodedAttachment), ptx) // Option two: Template all files
+		}
+		// Handle err from all the switch statement ExecuteTemplate functions
```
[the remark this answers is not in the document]
glennzw: The HTML/text error isn't handled, and I'd planned to have the pdf and exe to be similar, but for now I'll just remove this and add the error handling inside the txt/html.

models/maillog.go
glennzw: Good thinking!

I'm not sure if the `templateApplied` boolean will work though, as we have three states for an attachment:

1. Never been checked to see if it has variables
2. Been checked, doesn't have variables
3. Been checked, does have variables

We could use an `int` and have 0, 1, 2 states to represent this. Or we could use a *bool as the default value would be null. I'll explore this now.

glennzw: Oh, thinking about this some more, we can just default to a state that yes "yes check this file" and change it to the inverse if no variables are found the first time processing it (which is actually the logic I use).

glennzw: I pushed code having a go at this approach and ran tests against my set of attachment files.

glennzw: Running some debug statements to make sure files that we know don't have template variables aren't being reprocessed, and I notice the code isn't working.

It seems the `a.templateApplied = true` (in my case a.noTemplateVars) logic doesn't 'stick', as on the next iteration it's false.

Perhaps I'm not understanding unexported variables correctly.

jordan-wright: I haven't tested, but I think the issue is that we're not passing a pointer to the attachment in the call to `msg.Attach`. Take a look at the code snippet above and you'll see that I changed it to be a pointer (e.g. `func(a *Attachment)...` then passing `(&a)` to that function.).

> Oh, thinking about this some more, we can just default to a state that says "yes check this file" and change it to the inverse if no variables are found the first time processing it (which is actually the logic I use). e.g:

Yep! That's generally what I was thinking. I could be wrong, but this feels like simply an optimization to avoid parsing the entire attachment for every email. That's why I was calling it `templateApplied`, since it was just a flag to say that we've already processed it for templating.

jordan-wright: Eh, thinking more about it, I should clarify the above comment. I was wrong to say "we only need to process the file once" since if it _does_ have template variables then we'll need to process it every time so that the right template variables are parsed, right?

But if there are no template variables, then you'd be right just to short-circuit and return the content as-is. Let me think a little bit about whether this would change our approach.

glennzw: Yeah if there are template variables we'll need to process the attachment every time.

So first pass we assume there are template variables, and at the end we set the flag to confirm/deny this.

Then on subsequent passes we either return immediately as there are no variables, or we parse the attachment for the variables.

Understood on passing the pointer now, yes that makes sense. I'll fiddle some more while you ponder to see if we're doing it right.

jordan-wright:
> For the test writing are you thinking automated code, or just creating a bunch of attachments and sending out campaigns manually to test them?

Automated tests, like the ones you'll find in the various `_test.go` files (though admittedly I need more tests in the `models/` package 😉

There are multiple reasons for this:
* It gives examples on how the package can be used
* It helps us think through how things should look
* It helps us verify that the current implementation does what we want it to do in various scenarios
* It helps us ensure that we don't break the expected behavior in the future (I've saved myself multiple times by having good tests)

In this case, what you could probably do is make a `models/testdata/` folder since this is special in Go, then put the contents of your files in there. I'd probably just do at least one minimal-as-possible file with templates and one without templates for each _type_ of attachment (e.g. just a `docm`, probably not every office type).

Then the question is how to add the files. The easiest thing to do would be to just commit them to the repo, but I'm a bit nervous about committing macro-enabled docs to the Gophish repo since I don't want it to get accidentally flagged. Instead, you could just make something like a `docm_template.go` and `docm_notemplate.go` and add the base64 encoded file contents in those files as constants which can then be read into an attachment for your tests.

> One thing to note is that unlike writing an email or landing page template the user will only find out they used a {{.BadVariable}} when the campaign is launched, but the error is handled gracefully:

Out of curiosity, what's stopping us from doing the same kind of validation in attachments as we do in our templates today (ref [here](https://github.com/gophish/gophish/blob/master/models/template.go#L37) and [here](https://github.com/gophish/gophish/blob/master/models/template_context.go#L103:6))? We could make a `ValidateAttachment` function that acts the same way- just passing a sample context with the first-class template values (custom template vars are a different subject in https://github.com/gophish/gophish/pull/1813 that I still want to finish up sometime soon).

glennzw:
Ah yes great idea on the base64 blobs for safe testing- I'll get working on that once I've nailed down the main code. I haven't created automated testing code before so this will be a good learning opportunity.

>Out of curiosity, what's stopping us from doing the same kind of validation in attachments as we do in our templates today (ref here and here)? We could make a ValidateAttachment function that acts the same way- just passing a sample context with the first-class template values (custom template vars are a different subject in #1813 that I still want to finish up sometime soon).

Hmm 🤔  If I had a good reason I've forgotten what it was. Agreed, I don't see a reason why we can't have a `ValidateAttachment` function. I'll jump on that next.

models/maillog.go
glennzw: I suspect I'm missing something. I've copy/pasted your `msg.Attach()` code (just added the error handling) with appropriate by-reference logic now:

```go
// Attach the files
fmt.Println("Handling attachments for user " + r.FormatAddress())
for _, a := range c.Template.Attachments {
    msg.Attach(func(a *Attachment) (string, gomail.FileSetting, gomail.FileSetting) {
        h := map[string][]string{"Content-ID": {fmt.Sprintf("<%s>", a.Name)}}
        return a.Name, gomail.SetCopyFunc(func(w io.Writer) error {
            content, err := a.ApplyTemplate(ptx)
            if err != nil {
                return err
            }
            _, err = io.Copy(w, content)
            return err
        }), gomail.SetHeader(h)
    }(&a))
}
```


Then in `attachment.go`:

```go
type Attachment struct {
	Id          int64  `json:"-"`
	TemplateId  int64  `json:"-"`
	Content     string `json:"content"`
	Type        string `json:"type"`
	Name        string `json:"name"`
	vanillaFile bool   // Vanilla file, when true, has no template variables
}
```

(vanillaFile denotes a file *without* variables)

and then the `ApplyTemplate` function:

```go
func (a *Attachment) ApplyTemplate(ptx PhishingTemplateContext) (io.Reader, error) {

	fmt.Printf("File ID: %d\nFile name: %s\nTemplate boolean: %t\n", a.Id, a.Name, a.vanillaFile)
	a.vanillaFile = true  // Hard set for testing

      ......
```

The boolean doesn't seem to propagate between instances of the attachment between different emails, even with the above 'hard set':
```
Handling attachments for user "Test 1" <test1@mail.com>
File ID: 90
File name: doc-without-template-vars.docx
Template boolean: false

Handling attachments for user "Test 2" <test2@mail.com>
File ID: 90
File name: doc-without-template-vars.docx
Template boolean: false

Handling attachments for user "Test 2" <test2@mail.com>
File ID: 90
File name: doc-without-template-vars.docx
Template boolean: false
```

If I'm not mistaken we're looping through multiple attachments on a single email with the above code, so if the email had the identical attachment multiple times this would work (not that that's possible in gophish). Once we move on to a *new* email and we reach this code again we don't have a reference to the boolean.

glennzw: Yes, that thread seemed to be the right one to pull - this seems to work now. Iterating over `c.Template.Attachments` by reference:

```go
for i, _ := range c.Template.Attachments {
	a := &c.Template.Attachments[i]
	msg.Attach(func(a *Attachment) (string, gomail.FileSetting, gomail.FileSetting) {
		h := map[string][]string{"Content-ID": {fmt.Sprintf("<%s>", a.Name)}}
		return a.Name, gomail.SetCopyFunc(func(w io.Writer) error {
			content, err := a.ApplyTemplate(ptx)
			if err != nil {
				return err
			}
			_, err = io.Copy(w, content)
			return err
		}), gomail.SetHeader(h)
	}(a))
}
```

This is probably what you meant in your original message and I read it wrong.

glennzw:
Attachment validation:Done!  ✅

Screenshot 2020-08-16 at 22 31 57

glennzw:
> Automated tests, like the ones you'll find in the various `_test.go` files (though admittedly I need more tests in the `models/` package 😉

I had a go at writing some automated tests. I've added `attachment_test.go` which applies the `attachment.ApplyTemplate()` function to all the files in `testdata`. I followed your suggestion and turned the .docm Word macro into a base64 file.

The output from `go test` looks pretty good:
```
Checking attachment file -> html-file-with-vars.html
Checking attachment file -> html-file-without-vars.html
Checking attachment file -> text-file-with-vars.txt
Checking attachment file -> text-file-without-vars.txt
Checking attachment file -> word-file-with-vars.docm.b64
Checking attachment file -> word-file-without-vars.docx
```
Full output here: [gotest.txt](https://github.com/gophish/gophish/files/5086237/gotest.txt)

Having a file with `{{.BarVar}}` results in the test failing (just had a thought, maybe that should be a test too).

models/attachment.go
```diff
@@ -1,11 +1,149 @@
...
-	TemplateId int64  `json:"-"`
-	Content    string `json:"content"`
-	Type       string `json:"type"`
-	Name       string `json:"name"`
+	Id          int64  `json:"-"`
+	TemplateId  int64  `json:"-"`
+	Content     string `json:"content"`
+	Type        string `json:"type"`
+	Name        string `json:"name"`
+	vanillaFile bool   // Vanilla file has no template variables
```
jordan-wright, line 21: Could we think up a different name than `vanillaFile`? It's not really clear what this stands for (naming is hard!)

Maybe something like `isTemplate`?

models/attachment.go
```diff
@@ -1,11 +1,149 @@
...
+}
+
+// ValidateAttachment ensures that the provided attachment uses the supported template variables correctly.
+func (a Attachment) ValidateAttachment() error {
+
+	ptx := PhishingTemplateContext{
```
jordan-wright, line 27: For consistency, could we make sure we're using the same values as the existing [ValidateTemplate](https://github.com/gophish/gophish/blob/b01bd6cbc0bbf3109511d280b545708b3e2089d0/models/template_context.go#L101) function? I'm sure you can tell by now, but I quite like consistency where possible 😄 

I wish there were a way for us to reuse `ValidateTemplate` itself, but I think that might be too tricky.

models/attachment.go
```diff
@@ -1,11 +1,149 @@
...
+	Id          int64  `json:"-"`
+	TemplateId  int64  `json:"-"`
+	Content     string `json:"content"`
+	Type        string `json:"type"`
+	Name        string `json:"name"`
+	vanillaFile bool   // Vanilla file has no template variables
+}
+
+// ValidateAttachment ensures that the provided attachment uses the supported template variables correctly.
+func (a Attachment) ValidateAttachment() error {
```
jordan-wright, line 25: Let's call this `Validate` - it's cleaner and less repetitive.

models/attachment.go
```diff
@@ -1,11 +1,149 @@
...
+
+	_, err := a.ApplyTemplate(ptx)
+	return err
+}
+
+// ApplyTemplate parses different attachment files and applies the supplied phishing template.
+func (a *Attachment) ApplyTemplate(ptx PhishingTemplateContext) (io.Reader, error) {
+
+	var processedAttachment string
+	decodedAttachment, err := base64.StdEncoding.DecodeString(a.Content) // Decode the attachment
```
jordan-wright, line 49: I'd love for us to try and keep things as an actual reader the whole way through if possible. I can't guarantee it'd work, but it'd be worth the exploration. I'll drop some comments throughout on where things might need to be adjusted.

For example, here we could make this `base64.NewDecoder(base64.StdEncoding, strings.NewReader(a.Content)`

models/attachment.go
```diff
@@ -1,11 +1,149 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// If we've already determined there are no template variables in this attachment return it immediately
+	if a.vanillaFile == true {
+		return strings.NewReader(string(decodedAttachment)), nil
```
jordan-wright, line 56: And then we should be able to return our previous reader as-is.

models/attachment.go
```diff
@@ -1,11 +1,149 @@
...
+		fileExtension := filepath.Ext(a.Name)
+
+		switch fileExtension {
+
+		case ".docx", ".docm", ".pptx", ".xlsx", ".xlsm":
+			// Most modern office formats are xml based and can be unarchived.
+			// .docm and .xlsm files are comprised of xml, and a binary blob for the macro code
+
+			// Create a new zip reader from the file
+			zipReader, err := zip.NewReader(bytes.NewReader(decodedAttachment), int64(len(decodedAttachment)))
```
jordan-wright, line 71: This is the place I'm worried about re: keeping things a streaming reader, since it requires the length of our content. To that end, we _might_ be able to do something like writing to a `bytes.Buffer` or something to get that, but we should think about this a bit.

models/attachment.go
```diff
@@ -1,11 +1,149 @@
...
+				subFileExtension := filepath.Ext(zipFile.Name)
+				var tFile string
+				if subFileExtension == ".xml" || subFileExtension == ".rels" { // Ignore other files, e.g binary ones and images
+					// For each file apply the template.
+					tFile, err = ExecuteTemplate(string(contents), ptx)
+					if err != nil {
+						return nil, err
+					}
+					// Check if the subfile changed. We only need this to be set once to know in the future to check the 'parent' file
+					if tFile != string(contents) {
```
jordan-wright, line 102: If we only need to set this once, we should check if it's set in this condition, e.g.:

```go
if !fileContainedTemplateVars && tFile != string(contents) { ... }
```

jordan-wright: And out of curiosity, why don't we just set the (renamed) `a.isTemplate = true` here instead of having both `fileContainedTemplatesVars` and `a.isTemplate`?

models/attachment.go
```diff
@@ -1,11 +1,149 @@
...
+
+			zipWriter.Close()
+			processedAttachment = newZipArchive.String()
+
+		case ".txt", ".html":
+			processedAttachment, err = ExecuteTemplate(string(decodedAttachment), ptx)
+			if err != nil {
+				return nil, err
+			}
+			if processedAttachment == string(decodedAttachment) {
```
jordan-wright, line 136: In general, there's a whole bunch of calls to `string()` in this function which isn't _bad_ but I'll come back later and see if we could clean that up.

models/maillog.go
```diff
@@ -27,6 +25,10 @@ var MaxSendAttempts = 8
 // MailLog is exceeded.
 var ErrMaxSendAttempts = errors.New("max send attempts exceeded")
 
+// processAttachment is used to to keep track of which email attachments have templated values.
```
jordan-wright, line 28: I don't think we need this anymore, right?

jordan-wright (changes requested):
Hey @glennzw!

Took another quick-ish pass at this, things are looking good! This will be a huge feature to add to Gophish, so I appreciate you working on it.

I added some general feedback. Some minor things, and a bigger thread around whether or not we can keep things as a reader throughout the function to be more memory conscious which may or may not be possible.

Additionally, the tests are looking good! I'll review them more later. But you might make it where any binary format is base64 encoded. Right now it looks like the docx is still the standard format.

Keep up the great work and let me know if you have any questions!

glennzw:
Thanks @jordan-wright ! I'll get on these ASAP. Yeah I admit I felt a bit dirty with so many conversions between readers and strings and base64s 😅 I'll see what I can do and reach out if I get stuck!

models/attachment.go
glennzw: Yeah I was struggling on the name. We need something to indicate '_not a template file_' as we use the default condition of false in the logic, so it'd need to be `isNotTemplate`, which seems a bit negative. Maybe `isPlain` or `withoutTemplate`?

models/attachment.go
glennzw: Good call, consistency is key! I did notice the values in [ValidateTemplate](https://github.com/gophish/gophish/blob/b01bd6cbc0bbf3109511d280b545708b3e2089d0/models/template_context.go#L101) are missing `Position` (in BaseRecipient), `URL`, `TrackingURL`, `Tracker`, and `From` parameters. I assume these are automatically handled - but haven't checked thoroughly.

Either way, I'll update to match - and maybe in the future we can roll the attachment validation into `ValidateTemplate`.

models/attachment.go
glennzw: Been working on this today. So, Zip files require random access for reading, making it hard to stream bytes. The solution, as you suggested, seems to be to use a buffer. Here's a PoC I'm rolling into the code now:

https://play.golang.org/p/qCzzk-xqRA0

models/attachment.go
glennzw: That's some satisfying short circuiting there!

Hmm, I think I created `fileContainedTemplatesVars` first before implementing your more succinct `a.isTemplate` approach, and didn't merge the ideas.

glennzw: I realise the problem with your suggestion. 

It boils down to needing three states for a file;

0 --> File has never been checked, so we don't know if it has template variables (= apply template logic)
1 --> File has template variables (= apply template logic)
2 --> File does not have template variables (= return file)

To get around this, and use a regular boolean, I assume that a file *does* have template variables (thus the name `isVanilla` as this defaults to false - indicating template variables *are* present. This ensures files we haven't checked before are parsed, and if no template variables are found we can flip this default boolean. We have to do this kind of reverse logic as (AFAIK) we can only set the default state of a boolean to false.

When we load up a file the first thing we do is check if we have previously ascertained that is does not contain template variables, and can safely return it as is:

```
if a.vanillaFile == true {
   return decodedAttachment, nil
} else {
  ....
```

As we parse over a zip file's contents we check each file in the archive to see if its content changed after applying the template, and is therefore a templated file. We are interested in the inverse, however, as our initial assumption is that it *is* a templated file. So, if the content has remained the same for *every* file in the zip archive, we can then finally set `a.vanillaFile` to `true`, away from its default of `false`. This is why we need that temporary boolean `fileContainedTemplateVars`.

Of course the alternative is to have `a.isTemplate` as an `int` with the three states mentioned above. I'm not sure which approach is more readable, but given the trickiness of explaining the above, the tristate might be easier.

The other other alternative is that I'm talking nonsense and there's a trivial solution I'm not seeing 😅

models/maillog.go
glennzw: Oops, correct.

glennzw:
Hi @jordan-wright - I had a go at keeping things as Readers. See comments above about trickiness with zip files as we have to read the whole thing into memory. Also, with the [ExecuteTemplate](https://github.com/gophish/gophish/blob/22c7b9be14241ab38e6901a88e99b45740a3561f/models/template_context.go#L77) function requiring a string (and returning a string) it's a little tricky keeping things a Readers.

As for the naming convention of `isVanilla` see above discussion too around the tristate boolean. I suspect you'll have a more elegant solution in mind. 

Hopefully I'm in the right direction overall !

CodySchluenz:
How were you able to get the Word Document to send data back to goPhish without the user enabling macros? To check if the user opened the attachment

glennzw:
> How were you able to get the Word Document to send data back to goPhish without the user enabling macros? To check if the user opened the attachment

I added a linked image in the Word document (see [here](https://www.webworks.com/Documentation/Reverb/index.html#page/Preparing%20and%20Publishing%20Content/Preparing%20Word%20Files.3.21.htm)) and then changed the image source to be `{{.TrackingURL}} `.

CodySchluenz:
> > How were you able to get the Word Document to send data back to goPhish without the user enabling macros? To check if the user opened the attachment
> 
> I added a linked image in the Word document (see [here](https://www.webworks.com/Documentation/Reverb/index.html#page/Preparing%20and%20Publishing%20Content/Preparing%20Word%20Files.3.21.htm)) and then changed the image source to be `{{.TrackingURL}} `.

How did you get around word converting the {{.TrackingUrl}} to "%7b%7b.URL%7d%7d" ?

glennzw:
I edited the file in vim.

Alternative:

- Save docx with a real remote image, say `http://foo.com/cat.png`
- Close file
- Unzip .docx file - this will create a folder with various files as mentioned in "Notes on attachments" above.
- Edit the appropriate .rels file in a plain text editor and swap `http://foo.com/cat.png` for `{{.URL}}`

Once we wrap up the code part of this pull request I'll start on the documentation side.

rc4:
Hi @glennzw - not sure if this is any help to you at all, but a while ago I wrote a little [python script](https://github.com/rc4/gophish-macro-attachment) that supports the old OLE .doc format for getting a user's RID into a macro/image URL. It also (ab)uses the MHT document format, which can be renamed to .doc to have it automatically open in Word. MHT documents (even when renamed to .doc) will automatically fetch images when opened, allowing you to embed a tracking pixel and mark something as just "opened" without needing the user to enable macros (which send a POST so the user is marked as "submitted").

I have no knowledge of Go but if you'd like to take a look at what I'm doing in Python it might help to extend the functionality of your code. If you have any questions feel free to ask me - I spent a while messing around with these formats to get my script working.

glennzw:
> Hi @glennzw - not sure if this is any help to you at all, but a while ago I wrote a little [python script](https://github.com/rc4/gophish-macro-attachment) that supports the old OLE .doc format for getting a user's RID into a macro/image URL. It also (ab)uses the MHT document format, which can be renamed to .doc to have it automatically open in Word. MHT documents (even when renamed to .doc) will automatically fetch images when opened, allowing you to embed a tracking pixel and mark something as just "opened" without needing the user to enable macros (which send a POST so the user is marked as "submitted").
> 
> I have no knowledge of Go but if you'd like to take a look at what I'm doing in Python it might help to extend the functionality of your code. If you have any questions feel free to ask me - I spent a while messing around with these formats to get my script working.

Hi @rc4 ! Thanks so much for this, it's super interesting and useful. I'll see if I can leverage some of your tricks in my pull request.

I especially like the idea of smuggling data in the comments of a document

```ActiveDocument.BuiltInDocumentProperties("Comments")```

**Comment by theguybrush** [comment]
Hi @glennzw, I am very interested in these functionalities in GoPhish and I see that works are in progress =) . Is there an expected time for merge and release (aprox)?  
Thank you!

**Comment by glennzw** [comment]
> Hi @glennzw, I am very interested in these functionalities in GoPhish and I see that works are in progress =) . Is there an expected time for merge and release (aprox)?
> Thank you!

Thanks for the interest! As for being merged into master it's hard to tell, but @jordan-wright might have a better idea (largely dependent on his scarce and valuable time!).

In the meantime feel free to test this branch; I know a few people are already using it.

File: models/attachment.go (hunk 10 above)
Comment (eightoftwelve) [reply models/attachment.go]: i have successfully built the pull from attachment-template-support.  I downloaded your sample document and created a test campaign.  When I received the attachment (in the email),  All of the fields has been updated correctly.  If I paste them into chrome, the stats update.

My problem is that the document will not open and run the macros.  I get "Run-time error '5151':  Work was unable to read this document.  It may be corrupt.  Debug and Alt-F11 opens the code to the "Documents.open (URL)" highlighted yellow.

I have also ensured that the landing page is a blank page in case the landing page was causing the issue.

Any advise would be appreciated.

**Comment by eightoftwelve** [comment]
i have successfully built the pull from attachment-template-support. I downloaded your sample document and created a test campaign. When I received the attachment (in the email), All of the fields has been updated correctly. If I paste them into chrome, the stats update.

My problem is that the document will not open and run the macros. I get "Run-time error '5151': Work was unable to read this document. It may be corrupt. Debug and Alt-F11 opens the code to the "Documents.open (URL)" highlighted yellow.

I have also ensured that the landing page is a blank page in case the landing page was causing the issue.

Any advise would be appreciated.

**Comment by glennzw** [comment]
Hi @eightoftwelve, thanks for reaching out and sorry you're having trouble. In order to keep this pull request focussed would you mind creating an issue instead (https://github.com/gophish/gophish/issues)? In the issue could you attach the Word document, and info on the environment you're running gophish on (Windows/Linux, etc) and the environment you're opening the document on. Thanks!


> i have successfully built the pull from attachment-template-support. I downloaded your sample document and created a test campaign. When I received the attachment (in the email), All of the fields has been updated correctly. If I paste them into chrome, the stats update.
> 
> My problem is that the document will not open and run the macros. I get "Run-time error '5151': Work was unable to read this document. It may be corrupt. Debug and Alt-F11 opens the code to the "Documents.open (URL)" highlighted yellow.
> 
> I have also ensured that the landing page is a blank page in case the landing page was causing the issue.
> 
> Any advise would be appreciated.

**Comment by S10XtremeNLow** [comment]
Is there specific logging to the attachment templating? For example, some HTML files that I'm using don't inject the variables and others do.

File: models/attachment.go [hunk 12]
```diff
@@ -1,11 +1,151 @@
...
+					return nil, err
+				}
+				_, err = newZipFile.Write([]byte(tFile))
+				if err != nil {
+					zipWriter.Close()
+					return nil, err
+				}
+			}
+			// If no files in the archive had template variables, we set the 'parent' file to not be checked in the future
+			if fileContainedTemplatesVars == false {
```
jordan-wright, line 124: I'm still hoping that I can come up with a simpler option (e.g. looking at `sync.Once` as an alternative), but in the meantime, I think we can assign `a.vannillaFile` above and not use `fileContainedTemplatesVars`?

models/attachment.go
```diff
@@ -1,11 +1,151 @@
...
+				a.vanillaFile = true
+			}
+			zipWriter.Close()
+			return bytes.NewReader(newZipArchive.Bytes()), err
+			//processedAttachment = newZipArchive.String()
```
jordan-wright, line 129: Minor nit- can remove this.

models/attachment.go
```diff
@@ -1,11 +1,151 @@
...
+
+		case ".txt", ".html":
+			// Feels like a lot of Reader --> String --> Reader going on here
+			buf := new(strings.Builder)
```
jordan-wright, line 133: Worst case you can just call `ioutil.ReadAll(decodedAttachment)` which gives you a `[]byte`, then send `string()` to `ExecuteTemplate`.

models/testdata/text-file-without-vars.txt
```diff
@@ -0,0 +1,11 @@
+There are no variables in this file.
```
jordan-wright, line 1: Honestly, this is probably all you need for this test - no need for the rest of the long file. 😄

jordan-wright (changes requested):
Apologies for the delay @glennzw. I'm trying to get back to some of these outstanding PR's to hopefully get them merged.

I'm still hopeful that we can find a slightly simpler approach to the caching issue, and I wish we had a more streaming solution rather than a somewhat currently-unavoidable mixture of readers, strings, and []bytes. But until then we'll keep moving forward 😄

glennzw:
> Apologies for the delay @glennzw. I'm trying to get back to some of these outstanding PR's to hopefully get them merged.
> 
> I'm still hopeful that we can find a slightly simpler approach to the caching issue, and I wish we had a more streaming solution rather than a somewhat currently-unavoidable mixture of readers, strings, and []bytes. But until then we'll keep moving forward 😄

No worries @jordan-wright ! I'll have a look through your comments and give some thought to simplifying those ugly mixtures :)

models/attachment.go
```diff
@@ -1,11 +1,151 @@
...
+		case ".docx", ".docm", ".pptx", ".xlsx", ".xlsm":
+			// Most modern office formats are xml based and can be unarchived.
+			// .docm and .xlsm files are comprised of xml, and a binary blob for the macro code
+
+			// Zip archives require random access for reading, so it's hard to stream bytes. Solution seems to be to use a buffer.
+			// See https://stackoverflow.com/questions/16946978/how-to-unzip-io-readcloser
+			b := new(bytes.Buffer)
+			b.ReadFrom(decodedAttachment)
+			buf := b.Bytes()
+			zipReader, err := zip.NewReader(bytes.NewReader(buf), int64(len(buf))) // Create a new zip reader from the file
```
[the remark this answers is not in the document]
glennzw: I've been trying to figure this one out. Basically the problem is that zip archives require random access for reading, so it's hard to stream bytes (see [this post](https://stackoverflow.com/questions/16946978/how-to-unzip-io-readcloser)).

I did manage one piece of increased efficiency by getting the buffer length using `b.Len()`, and dropped the second buffer:

```
b := new(bytes.Buffer)
b.ReadFrom(decodedAttachment)
//buf := b.Bytes()
	
zipReader, err := zip.NewReader(bytes.NewReader(b.Bytes()), int64(b.Len())) 
```

But I can't find a way to use a solution like your proposed ```zipReader, err := zip.NewReader(b)```

I created a snippet on Go playground which decompresses a 'hello world' text file, in case that saves you time having a go: https://play.golang.org/p/g7VDSPXxvZU

models/attachment.go
glennzw: I can't remember why I used `fileContainedTemplatesVars`. Removed.

models/attachment.go
glennzw: Great idea, done!

models/testdata/text-file-without-vars.txt
glennzw: Done 😅

glennzw:
Hi @jordan-wright ! Bouncing back to you 😄 I suspect the two main sticking points I need your help on are:

1. Improved efficiency reading a zip file in memory (if possible)
2. A more elegant solution to the vanillaFile (my explanation of its use [here](https://github.com/gophish/gophish/pull/1936#discussion_r479667702))

models/attachment.go
```diff
@@ -1,11 +1,142 @@
...
+			if subFileExtension == ".xml" || subFileExtension == ".rels" { // Ignore other files, e.g binary ones and images
+				// For each file apply the template.
+				tFile, err = ExecuteTemplate(string(contents), ptx)
+				if err != nil {
+					zipWriter.Close() // Don't use defer when writing files https://www.joeshaw.org/dont-defer-close-on-writable-files/
+					return nil, err
+				}
+				// Check if the subfile changed. We only need this to be set once to know in the future to check the 'parent' file
+				if tFile != string(contents) {
+					a.vanillaFile = true
```
jordan-wright, line 105: So I have some other changes that I want to make in order to make this a bit simpler, but to start I wanted to get your take on this.

I added a line in the test that says:

```go
c.Assert(a.vanillaFile, check.Equals, strings.Contains(fname, "without-vars"))
```

And it seems to have failed on `word-file-with-vars.docm.b64`:

```
attachment_test.go:66:
    c.Assert(a.vanillaFile, check.Equals, strings.Contains(fname, "without-vars"))
... obtained bool = true
... expected bool = false
```

That said, I think this should be `a.vanillaFile = false`? But then that seems to break on `word-file-without-vars.docx.b64`. All this to say I think we should make sure we have adequate testing. Preferably, we'd have tests to both validate that templates are being applied, being applied correctly, and that processing a templated attachment twice indeed gives the right output.

models/attachment.go
```diff
@@ -1,11 +1,142 @@
...
+	case ".txt", ".html":
+		b, err := ioutil.ReadAll(decodedAttachment)
+		if err != nil {
+			return nil, err
+		}
+		processedAttachment, err := ExecuteTemplate(string(b), ptx)
+		if err != nil {
+			return nil, err
+		}
+		if processedAttachment == string(string(b)) {
```
jordan-wright, line 134:
```suggestion
		if processedAttachment == string(b) {
```

jordan-wright (changes requested):
Hey @glennzw! This is looking good. I have a few more tweaks I'm going to add to this branch soon, but beyond a couple of smaller items I think the next step is to refocus on testing. You'll see a note that an addition to the tests seems to have caught a bug so we should consider how we can add some more robust tests to make sure this is doing everything we want it to do.

models/attachment.go
glennzw: I notice that rolling back to the previous commit results in the tests passing:

Test cases:
```
c.Assert(a.vanillaFile, check.Equals, strings.Contains(fname, "without-vars"))
c.Assert(a.vanillaFile, check.Not(check.Equals), strings.Contains(fname, "with-vars"))
```

Revert:
```
git checkout 304aa3b6b7afd4e58ce8c90fced19776318cd578  #Attempt at keeping objects as Readers
go test -v -timeout 30s github.com/gophish/gophish/models

---- SNIP ----
Checking attachment file -> html-file-with-vars.html
Checking attachment file -> html-file-without-vars.html
Checking attachment file -> text-file-with-vars.txt
Checking attachment file -> text-file-without-vars.txt
Checking attachment file -> word-file-with-vars.docm.b64
Checking attachment file -> word-file-without-vars.docx.b64
---- SNIP ----

ok  	github.com/gophish/gophish/models	(cached)
```

So, something we've tweaked in the last commit 'Small fixes' has broken something.

You're correct, the `isVanilla` value for `word-file-with-vars.docm.b64` should be `false` - it seems the Word files have become inverted somehow (they are correctly reflected in the previous commit).

I'll do some digging and see if I can find the cause, and also give some thought to the additional tests you suggest.

glennzw: Aha, found it! I misread your point: 
>  That said, I think this should be `a.vanillaFile = false` ?

Yes you're correct, the assignment should be `a.vanillaFile = false` and the missing piece was that before we process the file we set `a.vanillaFile` to `true` (assume it's vanilla unless the processing proves otherwise).

We now pass both tests mentioned above:

```
go test -timeout 30s github.com/gophish/gophish/models

ok  	github.com/gophish/gophish/models	0.320s
```

glennzw: > All this to say I think we should make sure we have adequate testing. Preferably, we'd have tests to both validate that templates are being applied, being applied correctly, and that processing a templated attachment twice indeed gives the right output.

I'm working on this now. My plan is to create 'expected' output files for each file, so we can compare input and output. Finding some initial oddities with `.docx` files - with no templates, the before and after files look identical when decompressed, but their zip sizes are off by almost double. I think it's an artefact of different compression routines. I'll figure it out and report back.

glennzw: I created some 'expected' output files, that have been parsed by the templating mechanism and give the desired outcome (I've manually expected them):

```
html-file-with-vars.templated.html
html-file-without-vars.templated.html
text-file-with-vars.templated.txt
text-file-without-vars.templated.txt
word-file-with-vars.docm.templated.b64
word-file-without-vars.docx.templated.b64
```

The plain text files are obvious. The .b64 ones can be decoded and inspected like so:

```
base64 --decode word-file-with-vars.docm.templated.b64 > word-file-with-vars.docm 
```

Screenshot 2020-12-10 at 18 27 36

We then compare the output of applying the template to a file, with the expected outcome file:

```
c.Assert(templatedFile, check.Equals, expectedOutput)
```

Seems to run fine:

```
go test -timeout 30s github.com/gophish/gophish/models
ok  	github.com/gophish/gophish/models	0.343s
```

Next I'll have a go at creating a test file for applying the template twice to a file.

### Also, something else to bring up:

I noticed when creating these test files that when we unarchive a Word file, process it, and re-archive it the resultant file is much smaller. The reason seems to be that Microsoft only applies about 11% compression, whereas the go zip library will apply about 65% (for these test files at least). I don't think this is a big issue for us, but we might want to look into lowering the compression ratio to better match Microsoft's (MS does this to have faster opening/saving I assume). Also, we might want to return the original file rather than the uncompressed and recompressed one if no template variables are present. If we don't, the first email will have slightly smaller Word document. Not a biggie, but maybe better to be consistent and deliver the exact file the user attached.

I could achieve that by adding the following to the [end of the word processing code](https://github.com/gophish/gophish/pull/1936/files#diff-1ed3f192802af9040844eed631caaa322821342e401dc8d4804bc2ec11e0ae8aR123)
```
if a.vanillaFile == true {
    return base64.NewDecoder(base64.StdEncoding, strings.NewReader(a.Content)), nil // If no template variables were found, return the original file
}
```

glennzw: > processing a templated attachment twice indeed gives the right output

Hi @jordan-wright just checking what you mean by this. Either:

(a) Apply templating function twice to the same input, to make sure we get the same output both times; or
(b) Apply templating function twice in sequence, so, `ApplyTemplate(ApplyTemplate(input))` ?

I don't think we need (a) as we now test the generated output against expected output. For (b) I that's functionally equivalent to the existing test cases with no variables in them (but I could add this test if you like).

I've also added a PowerPoint and Excel test case.

vysecurity:
I just tried to attach a .html file, and it doesn't replace the variables...?

Miesvanderlippe:
Is this PR still alive? What needs to happen in order to release these features?

cgrossman1:
I just ran a successful campaign against almost 400 users using an image and macro in a Word document.  Thanks so much for this!   

I was wondering if this could also work for a custom exe file that will report back to GoPhish if the user executes it.  I can make an exe file that makes an HTTPS call, but would need GoPhish to embed the RID into the executable, the same we it's being done with the Word document.

glennzw:
> I just ran a successful campaign against almost 400 users using an image and macro in a Word document. Thanks so much for this!

Hi @cgrossman1 - that's amazing news, thanks so much for testing it out! I'm really glad it worked at scale.

> I was wondering if this could also work for a custom exe file that will report back to GoPhish if the user executes it. I can make an exe file that makes an HTTPS call, but would need GoPhish to embed the RID into the executable, the same we it's being done with the Word document.

Yes absolutely on the .exe file, I've also been thinking about how to do this. I think the easiest approach would be to sign the exe file, as explained in this research from Thinkst: https://blog.thinkst.com/2016/05/certified-canarytokens-alerts-from_25.html (and [sample code](https://github.com/thinkst/canarytokens/blob/13fad642b11159715c274dea9c1fae85fa47f116/sign_file.py) to achieve that). It'd be quite easy as we just replace the _TOKEN value from this code in the root-ca.conf. I'll give it a go once I release V1.0 of the attachment support.

glennzw:
> Is this PR still alive? What needs to happen in order to release these features?

Hi @Miesvanderlippe - thanks for the interest, yes it's still alive and just waiting for me to find the time to finish it up. I've been working locally on documentation and examples as I want those to be as thorough as possible so we're not overwhelmed with support tickets.

cgrossman1:
> > I just ran a successful campaign against almost 400 users using an image and macro in a Word document. Thanks so much for this!
> 
> Hi @cgrossman1 - that's amazing news, thanks so much for testing it out! I'm really glad it worked at scale.
> 
> > I was wondering if this could also work for a custom exe file that will report back to GoPhish if the user executes it. I can make an exe file that makes an HTTPS call, but would need GoPhish to embed the RID into the executable, the same we it's being done with the Word document.
> 
> Yes absolutely on the .exe file, I've also been thinking about how to do this. I think the easiest approach would be to sign the exe file, as explained in this research from Thinkst: https://blog.thinkst.com/2016/05/certified-canarytokens-alerts-from_25.html (and [sample code](https://github.com/thinkst/canarytokens/blob/13fad642b11159715c274dea9c1fae85fa47f116/sign_file.py) to achieve that). It'd be quite easy as we just replace the _TOKEN value from this code in the root-ca.conf. I'll give it a go once I release V1.0 of the attachment support.

Actually, I realized that we use the Outlook client which blocks EXE files by default.  So, what would work in our situation is to have a campaign that looks like a required update is needed (Adobe/Microsoft), with a link to a landing page that looks official, but has a link to an executable.  When the user downloads and runs the executable, it sends a message to GoPhish that the executable was run.  This is probably trickier, because the GoPhish web server would have to modify the executable to insert the RID on the fly.  I had a thought, and since we use Windows clients, I tried to extract the ADS information (Zone.Identifier) which actually contains the URL including the RID, but I had SmartScreen issues, and also kepts getting that info  wiped out after launching it.

cgrossman1:
Also, is it possible for it to work with PDF files?  I created a PDF with a fake CAPTCHA and when you click on the box, it runs a Javascript that goes to a URL.  It would be nice if this could replace the URL tag with the URL with the user's specific RID.  It's all in plain text, so it should be the same as doing this with an HTML file.

Thanks again for this awesome work!

cgrossman1:
> Also, is it possible for it to work with PDF files? I created a PDF with a fake CAPTCHA and when you click on the box, it runs a Javascript that goes to a URL. It would be nice if this could replace the URL tag with the URL with the user's specific RID. It's all in plain text, so it should be the same as doing this with an HTML file.
> 
> For reference:
> https://unit42.paloaltonetworks.com/phishing-trends-with-pdf-files/
> 
> Thanks again for this awesome work!

I added ".pdf" to line 126 of attachment.go, and was able to send a PDF I created with a link that goes to the proper landing page!

vysecurity:
Now we just need to make the tool log all file hashes sent, because the
files are dynamically made, the file hashes change.

cgrossman1:
@vysecurity I put {{.URL}} in the hyperlink, and GoPhish was able to modify the link in the attachment that was sent to the dynamically created URL.

electroxero:
Currently testing the branch using the Dockerfile but is not working as expected.

The build from source, cloning the repository is working btw.


Dockerfile line where there is the problem `RUN go get -v && go build -v`
Example of the error:
```
[+] Building 143.2s (17/28)
 => [internal] load build definition from Dockerfile                                                                                                                                   0.1s
 => => transferring dockerfile: 32B                                                                                                                                                    0.1s
 => [internal] load .dockerignore                                                                                                                                                      0.1s
 => => transferring context: 2B                                                                                                                                                        0.1s
 => [internal] load metadata for docker.io/library/debian:stable-slim                                                                                                                  0.9s
 => [internal] load metadata for docker.io/library/node:latest                                                                                                                         1.6s
 => [internal] load metadata for docker.io/library/golang:1.11                                                                                                                         1.7s
 => [build-js 1/6] FROM docker.io/library/node:latest@sha256:987be0a7ca165b6b19bf77ca812e980f603a10c1000bf25095cd89e2b146f9ad                                                          0.0s
 => [internal] load build context                                                                                                                                                      0.8s
 => => transferring context: 39.03kB                                                                                                                                                   0.8s
 => [stage-2  1/12] FROM docker.io/library/debian:stable-slim@sha256:457715c656bf1b14ae3790853c1a4bde13a7e740c510b9c029d38012be78d8c6                                                  0.0s
 => [build-golang 1/4] FROM docker.io/library/golang:1.11@sha256:e972c78795b22d5cfab02ac410aa2305fcc036319a7af51065d1af583cd3ec04                                                      0.0s
 => CACHED [build-js 2/6] RUN npm install gulp gulp-cli -g                                                                                                                             0.0s
 => CACHED [build-js 3/6] WORKDIR /build                                                                                                                                               0.0s
 => CACHED [build-golang 2/4] WORKDIR /go/src/github.com/gophish/gophish                                                                                                               0.0s
 => [build-golang 3/4] COPY . .                                                                                                                                                        1.0s
 => [build-js 4/6] COPY . .                                                                                                                                                            1.1s
 => ERROR [build-golang 4/4] RUN go get -v && go build -v                                                                                                                            139.4s
 => [build-js 5/6] RUN npm install --only=dev                                                                                                                                         83.5s
 => [build-js 6/6] RUN gulp                                                                                                                                                           19.3s
------
 > [build-golang 4/4] RUN go get -v && go build -v:
#14 0.455 go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
#14 0.455       ignoring go.mod;
#14 0.455       see 'go help modules'
#14 0.572 github.com/sirupsen/logrus (download)
#14 2.302 Fetching https://golang.org/x/sys/unix?go-get=1
#14 2.598 Parsing meta tags from https://golang.org/x/sys/unix?go-get=1 (status code 200)
#14 2.598 get "golang.org/x/sys/unix": found meta tag get.metaImport{Prefix:"golang.org/x/sys", VCS:"git", RepoRoot:"https://go.googlesource.com/sys"} at https://golang.org/x/sys/unix?go-get=1
#14 2.598 get "golang.org/x/sys/unix": verifying non-authoritative meta tag
#14 2.598 Fetching https://golang.org/x/sys?go-get=1
#14 2.747 Parsing meta tags from https://golang.org/x/sys?go-get=1 (status code 200)
#14 2.748 golang.org/x/sys (download)
#14 19.36 github.com/NYTimes/gziphandler (download)
#14 20.22 Fetching https://golang.org/x/crypto/bcrypt?go-get=1
#14 20.38 Parsing meta tags from https://golang.org/x/crypto/bcrypt?go-get=1 (status code 200)
#14 20.38 get "golang.org/x/crypto/bcrypt": found meta tag get.metaImport{Prefix:"golang.org/x/crypto", VCS:"git", RepoRoot:"https://go.googlesource.com/crypto"} at https://golang.org/x/crypto/bcrypt?go-get=1
#14 20.38 get "golang.org/x/crypto/bcrypt": verifying non-authoritative meta tag
#14 20.38 Fetching https://golang.org/x/crypto?go-get=1
#14 20.55 Parsing meta tags from https://golang.org/x/crypto?go-get=1 (status code 200)
#14 20.55 golang.org/x/crypto (download)
#14 23.64 github.com/PuerkitoBio/goquery (download)
#14 24.79 github.com/andybalholm/cascadia (download)
#14 25.50 Fetching https://golang.org/x/net/html?go-get=1
#14 25.66 Parsing meta tags from https://golang.org/x/net/html?go-get=1 (status code 200)
#14 25.66 get "golang.org/x/net/html": found meta tag get.metaImport{Prefix:"golang.org/x/net", VCS:"git", RepoRoot:"https://go.googlesource.com/net"} at https://golang.org/x/net/html?go-get=1
#14 25.66 get "golang.org/x/net/html": verifying non-authoritative meta tag
#14 25.66 Fetching https://golang.org/x/net?go-get=1
#14 25.80 Parsing meta tags from https://golang.org/x/net?go-get=1 (status code 200)
#14 25.80 golang.org/x/net (download)
#14 32.37 github.com/emersion/go-imap (download)
#14 34.94 Fetching https://golang.org/x/text/encoding?go-get=1
#14 35.13 Parsing meta tags from https://golang.org/x/text/encoding?go-get=1 (status code 200)
#14 35.13 get "golang.org/x/text/encoding": found meta tag get.metaImport{Prefix:"golang.org/x/text", VCS:"git", RepoRoot:"https://go.googlesource.com/text"} at https://golang.org/x/text/encoding?go-get=1
#14 35.13 get "golang.org/x/text/encoding": verifying non-authoritative meta tag
#14 35.13 Fetching https://golang.org/x/text?go-get=1
#14 35.32 Parsing meta tags from https://golang.org/x/text?go-get=1 (status code 200)
#14 35.32 golang.org/x/text (download)
#14 50.78 Fetching https://golang.org/x/text/transform?go-get=1
#14 50.93 Parsing meta tags from https://golang.org/x/text/transform?go-get=1 (status code 200)
#14 50.93 get "golang.org/x/text/transform": found meta tag get.metaImport{Prefix:"golang.org/x/text", VCS:"git", RepoRoot:"https://go.googlesource.com/text"} at https://golang.org/x/text/transform?go-get=1
#14 50.93 get "golang.org/x/text/transform": verifying non-authoritative meta tag
#14 50.94 github.com/emersion/go-sasl (download)
#14 51.80 github.com/emersion/go-message (download)
#14 53.12 github.com/emersion/go-textwrapper (download)
#14 54.63 bitbucket.org/liamstask/goose (download)
#14 58.02 github.com/go-sql-driver/mysql (download)
#14 60.62 github.com/kylelemons/go-gypsy (download)
#14 61.88 github.com/lib/pq (download)
#14 63.31 github.com/mattn/go-sqlite3 (download)
#14 73.12 github.com/ziutek/mymysql (download)
#14 74.33 github.com/gophish/gomail (download)
#14 76.25 github.com/jinzhu/gorm (download)
#14 83.87 github.com/jinzhu/inflection (download)
#14 84.62 github.com/oschwald/maxminddb-golang (download)
#14 87.43 github.com/jordan-wright/email (download)
#14 88.23 github.com/gorilla/csrf (download)
#14 89.02 github.com/gorilla/securecookie (download)
#14 89.88 github.com/pkg/errors (download)
#14 90.79 github.com/gorilla/sessions (download)
#14 91.66 Fetching https://golang.org/x/time/rate?go-get=1
#14 91.81 Parsing meta tags from https://golang.org/x/time/rate?go-get=1 (status code 200)
#14 91.81 get "golang.org/x/time/rate": found meta tag get.metaImport{Prefix:"golang.org/x/time", VCS:"git", RepoRoot:"https://go.googlesource.com/time"} at https://golang.org/x/time/rate?go-get=1
#14 91.81 get "golang.org/x/time/rate": verifying non-authoritative meta tag
#14 91.81 Fetching https://golang.org/x/time?go-get=1
#14 91.96 Parsing meta tags from https://golang.org/x/time?go-get=1 (status code 200)
#14 91.96 golang.org/x/time (download)
#14 92.53 github.com/gorilla/mux (download)
#14 93.43 github.com/gorilla/handlers (download)
#14 94.23 github.com/felixge/httpsnoop (download)
#14 94.97 github.com/jordan-wright/unindexed (download)
#14 95.69 Fetching https://gopkg.in/alecthomas/kingpin.v2?go-get=1
#14 96.32 Parsing meta tags from https://gopkg.in/alecthomas/kingpin.v2?go-get=1 (status code 200)
#14 96.32 get "gopkg.in/alecthomas/kingpin.v2": found meta tag get.metaImport{Prefix:"gopkg.in/alecthomas/kingpin.v2", VCS:"git", RepoRoot:"https://gopkg.in/alecthomas/kingpin.v2"} at https://gopkg.in/alecthomas/kingpin.v2?go-get=1
#14 96.32 gopkg.in/alecthomas/kingpin.v2 (download)
#14 97.57 github.com/alecthomas/template (download)
#14 98.30 github.com/alecthomas/units (download)
#14 99.22 golang.org/x/sys/internal/unsafeheader
#14 99.22 golang.org/x/net/html/atom
#14 99.22 golang.org/x/text/encoding/internal/identifier
#14 99.23 golang.org/x/text/internal/utf8internal
#14 99.23 github.com/lib/pq/oid
#14 99.23 golang.org/x/crypto/blowfish
#14 99.26 golang.org/x/text/transform
#14 99.26 golang.org/x/sys/unix
#14 99.29 github.com/emersion/go-sasl
#14 99.29 golang.org/x/net/html
#14 99.29 github.com/emersion/go-message/textproto
#14 99.29 github.com/emersion/go-textwrapper
#14 99.30 github.com/NYTimes/gziphandler
#14 99.30 golang.org/x/crypto/bcrypt
#14 99.35 github.com/gophish/gophish/context
#14 99.38 golang.org/x/text/internal/tag
#14 99.38 golang.org/x/text/encoding
#14 99.40 golang.org/x/text/runes
#14 99.42 github.com/gophish/gophish/auth
#14 99.48 github.com/go-sql-driver/mysql
#14 99.53 golang.org/x/text/internal/language
#14 99.53 github.com/emersion/go-imap/utf7
#14 99.53 golang.org/x/text/encoding/internal
#14 99.53 github.com/emersion/go-message
#14 99.55 github.com/kylelemons/go-gypsy/yaml
#14 99.59 # github.com/go-sql-driver/mysql
#14 99.59 ../../go-sql-driver/mysql/nulltime.go:36:15: undefined: sql.NullTime
#14 99.59 github.com/lib/pq/scram
#14 99.61 golang.org/x/text/encoding/charmap
#14 99.62 github.com/emersion/go-imap
#14 99.63 # github.com/emersion/go-message
#14 99.63 ../../emersion/go-message/charset.go:24:9: undefined: errors.As
#14 99.63 ../../emersion/go-message/encoding.go:27:9: undefined: errors.As
#14 99.64 golang.org/x/text/encoding/japanese
#14 99.71 golang.org/x/text/encoding/korean
#14 99.88 golang.org/x/text/encoding/simplifiedchinese
#14 100.0 github.com/andybalholm/cascadia
#14 100.0 golang.org/x/text/encoding/traditionalchinese
#14 100.2 github.com/sirupsen/logrus
#14 100.3 golang.org/x/text/encoding/unicode
#14 100.3 golang.org/x/text/internal/language/compact
#14 100.5 golang.org/x/text/language
#14 100.5 github.com/lib/pq
#14 100.5 github.com/emersion/go-imap/commands
#14 100.6 github.com/PuerkitoBio/goquery
#14 100.6 github.com/emersion/go-imap/responses
#14 100.6 github.com/gophish/gophish/logger
#14 100.6 golang.org/x/text/encoding/ianaindex
#14 100.7 github.com/mattn/go-sqlite3
#14 100.8 github.com/gophish/gophish/config
#14 100.8 github.com/ziutek/mymysql/mysql
#14 100.8 github.com/emersion/go-imap/client
#14 100.8 golang.org/x/text/encoding/htmlindex
#14 100.8 github.com/gophish/gomail
#14 100.9 github.com/gophish/gophish/webhook
#14 101.0 github.com/jinzhu/inflection
#14 101.1 github.com/oschwald/maxminddb-golang
#14 101.2 github.com/jordan-wright/email
#14 101.2 github.com/jinzhu/gorm
#14 101.2 github.com/gophish/gophish/mailer
#14 101.3 github.com/gorilla/securecookie
#14 101.3 github.com/ziutek/mymysql/native
#14 101.4 github.com/pkg/errors
#14 101.5 github.com/gorilla/sessions
#14 101.6 github.com/gorilla/csrf
#14 101.6 golang.org/x/time/rate
#14 101.6 github.com/gorilla/mux
#14 101.7 github.com/felixge/httpsnoop
#14 101.7 github.com/jordan-wright/unindexed
#14 101.8 github.com/alecthomas/template/parse
#14 101.8 github.com/gophish/gophish/middleware/ratelimit
#14 101.8 github.com/alecthomas/units
#14 101.9 github.com/ziutek/mymysql/godrv
#14 102.2 github.com/gorilla/handlers
#14 102.2 github.com/alecthomas/template
#14 102.4 gopkg.in/alecthomas/kingpin.v2
------
executor failed running [/bin/sh -c go get -v && go build -v]: exit code: 2
```

### Step to reproduce

```
git clone https://github.com/gophish/gophish.git
cd gophish
git checkout attachment-template-support
docker.exe build --tag gophish:dev . 
```

blshkv:
FYI, there is also https://github.com/onvio/gophish/commits/master. You might want to merge both if onvio fork has done something better

vonjk:
Hello. After pulling the latest version of gophish, try the template attachment function, but cannot upload documents in docx format.
image

H4xl0r:
> This pull request adds support to apply template variables to attachments. This has two main use cases:
> 
> 1. **Customize attachment** contents to targets e.g "Dear Glenn, this document is all about you."
> 2. The ability to **add tracking tokens to attachments** e.g. tracking pixel to a Word or PowerPoint document to tell when it's been opened, or of even more interest, if the user enabled macros when prompted.
> 
> There's no change to the front end. A user can simply choose to include the regular gophish template variables in certain files. I've included logic to only process a file once if no template variables are found.
> 
> I've used a switch statement to handle different attachment filetypes. At this stage I added support for:
> 
> * Plain text / HTML documents: Just apply the`ExecuteTemplate() function`
> * Office documents (.docx, .xlsx, .pptx, .docm, .xlsm): Unarchive the file, and apply the `ExecuteTemplate` to every .xml or .rels file inside the archive, then rearchive those files.
> * Default: I chose to do nothing, but we could choose to apply the `ExecuteTemplate()` to all files that don't match the switch. This might break binary files though.
> 
> I've tested all the above file types and it works super well. The most comprehensive example is a macro enabled Word document that has a {{.TrackingURL}} as a linked image in the header/footer, and the {{.URL}} in a variable (e.g textbox) that's called from macro code. The result is that we can tell:
> 
> 1. Email was sent
> 2. Word document was opened (in this case 'Email opened')
> 3. Word macros enabled (in this case 'Link clicked')
> 
> This will plug in nicely to my other pull request for custom events #1929 so we can get "Word Document opened" and "Macros enabled"
> 
> ## Notes on attachments
> I found that in Office documents the curly braces in template variables weren't always being represented conveniently in the compressed XML. If you're having trouble open the `.docx` in vim (or decompress) and have a look in `word/document.xml` to make sure the literal `{{.FirstName}}` etc is appearing.
> 
> For macros, as noted in[ this previous attempt](https://github.com/gophish/gophish/pull/260) from some time ago, `.docm` files compile their macro code to a binary format. However the rest of the document is still just a compressed bunch of XML files. e.g:
> 
> ```
> " zip.vim version v28
> " Browsing zipfile doc-with-macro.docm
> " Select a file with cursor and press ENTER
> 
> [Content_Types].xml
> _rels/.rels
> word/_rels/document.xml.rels
> word/document.xml
> word/_rels/vbaProject.bin.rels
> word/vbaProject.bin
> word/media/image1.jpg
> word/theme/theme1.xml
> word/settings.xml
> word/vbaData.xml
> customXml/_rels/item1.xml.rels
> customXml/itemProps1.xml
> customXml/item1.xml
> docProps/app.xml
> docProps/core.xml
> word/styles.xml
> word/fontTable.xml
> word/webSettings.xml
> ```
> 
> The macro code is compiled into the above `word/vbaProject.bin` file. We have two solutions here; either patch the binary with some placeholder variables string (e.g `url="AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"`) or place the URL variables in the Word document, and have them referenced by the macro code.
> 
> The latter approach worked beautifully in my testing. Create a textbox called `urlbox` with contents `{{.URL}}` and the following macro code will hit that URL when the document opens:
> 
> ```
> Sub urlfetch()
>  Dim shp As Shape
> 
>  For Each shp In ActiveDocument.Shapes
>     If shp.Name = "urlbox" Then
>         URL = shp.TextFrame.TextRange.Text
>              Documents.Open (URL)
>     End If
>  Next
> End Sub
> 
> Public Sub AutoOpen()
>     urlfetch
> End Sub
> ```
> 
> This is just PoC, it's a little kludgy as the webpage in `{{.URL}}` will be opened in a new Word window, but it's the quickest cross platform solution. A better option is the `Application.Documents.CanCheckOut (URL)` call but this sends a HEAD request (perhaps we can allow my new custom events to receive HEAD requests). My actual campaign Documents will include macro code to change the document into a 'gotcha' page, letting the user know that this was a campaign.
> 
> I can share my test files if needed:
> 
> Screenshot 2020-08-15 at 17 40 18
> 
> Or add them to `https://github.com/gophish/payloads`
> 
> The ability to track Office documents seems to be one of those most requested features (#880, #513, #352, #260, #227, #1667, #1606, #1205, #1176, #880, #352). I hope my solution solves this!

@glennzw 
could you share the payload files please

roxanne-tampus:
Hi @glennzw and @jordan-wright! Would like to know if it is possible to have a pdf attachment that was converted from html format when template variables is filled after starting a campaign?

example html template => invoice_html.html
```
Hi {{.FirstName}}. This is your official receipt after ordering from eBay. 
....
....
If you think you didn't make this transaction please <a class="btn btn-primary" href="{{.URL}}">Login</a>!
```

when campaign is started the content is expected to be
Hi John!
....
....
If you think you didn't make this transaction please <a class="btn btn-primary" href="{{.URL}}">Login</a>!

then convert this html to pdf and send to recipient as attachment > invoice.pdf

PR #1940, opened 2020-08-19
Implement SSRF Mitigations

### LINKED ISSUE #1908: Server Side Request Forgery via Import Site feature (related to closed issue #1026)
### What version of Gophish are you using?:
Gophish v0.10.1

### Brief description of the issue:
This issue has previously been raised [here #1026](https://github.com/gophish/gophish/issues/1026). It appears there were actually two seperate issues reported originally:

1) Lack of Authentication on endpoints (/import/site , /import/email and /import/group).
2) Server Side Request Forgery (SSRF)

The fix for the authentication was implemented but the SSRF is still exploitable and appears that there is no fix for this (it is functionality acting as intended).  By nature, the import site functionality needs to be able to make requests to any external IP address or domain name in order to clone the site. So an allow listing approach is not really feasible. Perhaps a regex to block requests to localhost, internal IP address ranges and AWS endpoints, but this just seems like a never ending battle? 🤷‍♂️

The account for the person who originally reported the issue appears to have been deleted but as per their issue:
> It is possible to access the "Import Site" endpoint of the "Create Landing Page" ~~area without cookie/API key, AKA without authentication~~. ~~As such unauthenticated~~ users who can access the admin portal can use it to perform server side request forgery attacks. Note: this is only really an issue if the admin interface is listening on anything other than 127.0.0.1, or if it is running on 127.0.0.1 on a multiuser machine accessed by non gophish admins.

I have struck out the text which is no longer relevant to reflect the authentication updates which have been made. The endpoints now require authentication but it is still possible to abuse the SSRF vulnerability to achieve the following:
- Attacks against the server itself (target other services running on localhost)
- Internal port scanning
- Attacks against other back-end systems e.g:
   - Abusing trust relationship between gophish and mail server
   - Querying AWS metadata service

__Note:__ AWS has added support for IMDSv2 to mitigate SSRF attacks against AWS metadata service. Also I have checked if various protocols and URL schemas (ssh, smb, file, gopher, dict, etc) and had no luck with any of these being supported.

### What are you expecting to see happen?
Expecting to not be able to port scan localhost, internal IP address ranges, and access AWS metadata service.

### What are you seeing happen?
It is possible to query the AWS metadata service and port scan localhost / internal IP address ranges. For example, it is possible to fetch the EC2 instance identity document (`http://169.254.169.254/latest/dynamic/instance-identity/document`) by inserting this as the endpoint to attempt to clone in the "Import Site" feature.
 
ssrf-req
ssrf-resp

jordan-wright:
This fixes #1908 by creating a *net.Dialer which restricts outbound connections to only allowed IP ranges. This implementation is based on the blog post at https://www.agwa.name/blog/post/preventing_server_side_request_forgery_in_golang

To keep things backwards compatible, by default we'll only block connections to 169.254.169.254, the link-local IP address commonly used in cloud environments to retrieve metadata about the running instance. For other internal addresses (e.g. localhost or RFC 1918 addresses), it's assumed that those are available to Gophish.

To support more secure environments, we introduce the `allowed_internal_hosts` configuration option where an admin can set one or more IP ranges in CIDR format. If addresses are specified here, then all internal connections will be blocked except to these hosts.

I've done various testing, but could use a second set of eyes here to make sure I didn't break anything.

There's various bits about this approach I don't really like. For example, since various packages all need this functionality, I had to make the RestrictedDialer a global singleton rather than a dependency off of, say, the admin server. Additionally, since webhooks are implemented via a singleton, I had to introduce a new function, `SetTransport`.

Finally, I had to make an update in the gomail package to support a custom net.Dialer.

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,7 +3,8 @@
 		"listen_url": "127.0.0.1:3333",
 		"use_tls": true,
 		"cert_path": "gophish_admin.crt",
-		"key_path": "gophish_admin.key"
+		"key_path": "gophish_admin.key",
+		"allowed_internal_hosts": ["127.0.0.1"]
 	},
 	"phish_server": {
 		"listen_url": "0.0.0.0:80",
--- a/config.json
+++ b/config.json
@@ -3,8 +3,7 @@
 		"listen_url": "127.0.0.1:3333",
 		"use_tls": true,
 		"cert_path": "gophish_admin.crt",
-		"key_path": "gophish_admin.key",
-		"allowed_internal_hosts": ["127.0.0.1"]
+		"key_path": "gophish_admin.key"
 	},
 	"phish_server": {
 		"listen_url": "0.0.0.0:80",
--- a/config/config.go
+++ b/config/config.go
@@ -9,11 +9,12 @@ import (
 
 // AdminServer represents the Admin server configuration details
 type AdminServer struct {
-	ListenURL string `json:"listen_url"`
-	UseTLS    bool   `json:"use_tls"`
-	CertPath  string `json:"cert_path"`
-	KeyPath   string `json:"key_path"`
-	CSRFKey   string `json:"csrf_key"`
+	ListenURL            string   `json:"listen_url"`
+	UseTLS               bool     `json:"use_tls"`
+	CertPath             string   `json:"cert_path"`
+	KeyPath              string   `json:"key_path"`
+	CSRFKey              string   `json:"csrf_key"`
+	AllowedInternalHosts []string `json:"allowed_internal_hosts"`
 }
 
 // PhishServer represents the Phish server configuration details
--- a/controllers/api/import.go
+++ b/controllers/api/import.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/PuerkitoBio/goquery"
+	"github.com/gophish/gophish/dialer"
 	log "github.com/gophish/gophish/logger"
 	"github.com/gophish/gophish/models"
 	"github.com/gophish/gophish/util"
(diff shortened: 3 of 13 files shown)


jordan-wright:
cc @glennzw - would you mind taking this for a quick spin to make sure it works as intended before I merge it?

glennzw:
> cc @glennzw - would you mind taking this for a quick spin to make sure it works as intended before I merge it?

Sure thing I'm going to check this out tomorrow!

glennzw:
Hi @jordan-wright - I took this update for a ride and it looks good to me. Test cases:

Current master:
- Able to import from localhost and another webserver on my local network

This branch:
- Unchanged config.json: Able to access localhost and other networked machines
- `"allowed_internal_hosts": []` : Able to access localhost and other networked machines (is this expected?)
- `"allowed_internal_hosts": ["127.0.0.1"]` : Only able to access local interface, no other machines on my network
- `"allowed_internal_hosts": ["192.168.0.11"]` : Unable to access localhost, able to connect to webserver on network
- `"allowed_internal_hosts": ["192.168.0.0/24"]` : Able to access any server on the subnet, but no others

<hr>

### Other observations:

Noticed the additional logging output, e.g:
`current allowed: []*net.IPNet{(*net.IPNet)(0xc0001e1050)}`

After a failed attempt to import a site, a successful attempt doesn't clear the previous error message:

Screenshot 2020-08-20 at 14 51 31

jordan-wright:
Thanks for the feedback @glennzw - it looks like all expected behavior.

And re: your other observations:

* You caught me, I was hoping I could remove that debug line before you found it 😛 It's removed now
* I think that's a UI bug that I'll fix in a different patch to keep this one focused to just the ssrf mitigations.

glennzw:
Haha, I thought maybe you left the debug statement in for me to find so you'd know I did a proper review ;-)

For the UI bug; I added the `errorFlashFade` function when doing the IMAP work, so messages are only displayed for N seconds. Might be useful for when we address this later.

https://github.com/gophish/gophish/blob/6df62e85fd60f0931d3c8bfdb13b436a961bc9b6/static/js/src/app/gophish.js#L14

PR #1967, opened 2020-09-14
Added functionality to display last user login

glennzw:
This pull request adds a column to the `Users & Groups` section to display the last successful login of each user:

lastlogin

I'm not sure if the database update logic should be in `models.user`. It didn't feel right putting in `auth` as there are no other database interactions there.

### DIFF
--- a/models/user.go
+++ b/models/user.go
@@ -55,18 +55,6 @@ func GetUserByUsername(username string) (User, error) {
 	return u, err
 }
 
-// UpdateLastLogin updates the timestamp for the last successful login for a user
-func UpdateLastLogin(username string) error {
-	u := User{}
-	err := db.Where("username=?", username).First(&u).Error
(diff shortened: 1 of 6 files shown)


models/user.go
```diff
@@ -53,6 +55,18 @@ func GetUserByUsername(username string) (User, error) {
 	return u, err
 }
 
+// UpdateLastLogin updates the timestamp for the last successful login for a user
```
jordan-wright, line 58: What do you think about removing this method and instead just update `LastLogin` then call `models.PutUser(&u)` back in the `Login` handler?

glennzw: Ah, much more elegant approach, good call @jordan-wright! I've made that change and committed.

jordan-wright:
LGTM - thanks @glennzw!

Issue #1971, opened 2020-09-16
Gophish becomes unresponsive while delivering emails to recipients and gophish server becomes unavailable 


bhagoria:
What version of Gophish are you using?: Gophish v0.10.1

Brief description of the issue: Gophish becomes unresponsive when people start clicking on the link in numbers and emails still being delivered. During this time people who click on the phishing link observe page can not be displayed. Also, some slowness in email sending.

What are you expecting to see happen? : Gophish console remains up and running, no performance issues, emails being delivered without any issue, people who click on the phishing link should get on the landing page.

What are you seeing happen?
When I target 10K people in a batch (targetting more than 40K in 4 different batches) and launch the campaign and the emails start being delivered, I notice that the Gophish console becomes so slow, unable to refresh, people who click on the phishing link start getting page can not be displayed. And, Gophish service keeps running above 100%.

I would anticipate that the issue can be reproduced when people start clicking on the link in numbers and emails still being delivered.

I am using phishing domain running on https. 

Please provide any terminal output that may be relevant below:

I am sharing the Gophish error that can help us to look into it. 

2020/09/15 09:16:50 http: TLS handshake error from 117.X.X.X:32698: read tcp 81.X.X.X:443->117.X.X.X:32698: i/o timeout
2020/09/15 09:16:50 http2: server: error reading preface from client 157.47.X.X:51726: read tcp 81.x.x.x:443->157.X.45.1:51726: i/o timeout
2020/09/15 09:18:33 http: TLS handshake error from 34.93.X.X:38438: read tcp 81.x.x.x:443->34.X.X.38:38438: read: connection reset by peer
 Microsoft Outlook 16.0.5017; ms-office; MSOffice 16)\""
time="2020-09-15T09:22:49+02:00" level=warning msg="451 4.7.0 Timeout waiting for client input" code=451 email="\"Tonese\" <toniasdddd"
time="2020-09-15T09:23:17+02:00" level=warning msg="write tcp 81.x.x.x:33974->81.X.X.X:25: write: broken pipe" 

Please provide as many steps as you can to reproduce the problem:
 * Step 1 - Take 12 K targets in one batch, launch the campaign with a phishing link running on https with redirect enabled. While emails being delivered and people start clicking on the phishing link. You will notice so much traffic on the phishing server that would make Gophish unresponsive.
 16 GB ram is there on the phishing server, which does not seem like a root cause of this issue.

glennzw:
[...]

Issue #1534, opened 2019-08-08
Site Becomes Unresponsive When Attempting Large (10k) Campaign


JoaoBernardoo:
Gophish ver. 7.1

We are attempting to run a large campaign of about 10k users at once, but when we attempt this the front-end of the website locks up and is unusable. Is there a recommended spec for the machine we run Gophish on, for campaigns of this size? I have checked previous posts with similar subject matter but, they are from 2017 and I would like to know if anything has changed since then. Thanks for any help!

jordan-wright:
Hi @JoaoBernardoo,

Thanks for reaching out! I have it as a "todo" to better handle server-side pagination for API requests and performance in general. The problem you're running into is that Gophish is setting up the campaign details in the database, loading the campaign JSON response, and trying to send that back to the browser. With such a large campaign, this has the potential to lock up the browser as you're seeing.

I'm always open to pull requests, but in the meantime you may consider leveraging our API for to manage large campaigns. We have a [Python API client](https://docs.getgophish.com/python-api-client/) that should make this pretty easy to get up and running.

Sorry I couldn't be of more help!

Issue #816, opened 2017-10-31
Lazy loading on Group Edit Page


bananabr:
When the number o members in a group is huge (~3000), as the group details/edit page loads all members to memory and leaves pagination and sorting to the frontend, some browsers may freeze or show hangup warnings.

I think that for groups larger than 100 members, some kind of lazy loading must be implemented. The same problem applies to the campaign details page.

jordan-wright:
Yes, this is a known issue. The long term fix is to leverage Datatables' ajax loading capabilities, but this hasn't been implemented yet since I've been working on other feature requests.

I'm happy to review pull requests if you or anyone else would like to send one up.

bananabr:
Altough I have web development experience, I am not familiar with Go and the framework gophish uses. But if you can give me some directions on how to setup a development environment, I can work on this.

jordan-wright:
Sure thing!

As far as setting up a development environment, you'll first need to set up Gophish from source: https://gophish.gitbooks.io/user-guide/content/installation.html

Then, you'll just need to `npm install` to get all the frontend dependencies.

The bad part about this issue is that there's a lot of moving parts. You'll have to implement the ajax handlers (I'd probably make `ajax.go` in the `controllers/` package or I'd consider seeing if I could hook together the API to the format datatables expects and just add parameters like `limit` and `offset` to the API to avoid split brain work with entire Ajax handlers).

Then you'd have to implement the logic in the models package to handle loading the right amount of results at the correct offsets.

Finally, you'd have to implement the front end JS work on the various pages.

I totally understand if such a big feature seems daunting. This is one of those things that I'd be able to get to eventually, but right now I'm finishing up a refactor of how we send emails, then I have a couple of other things in line before I'd be able to get to it.

Reedtechno:
Not sure if I should create a separate issue for this or not, but we are having a similar issue but with the Users & Groups page loading all groups on page load.  My teammate has been developing a tool to sync AD security groups into GoPhish and has done so effectively.  Now we have 1500+ groups and it takes 2.5 minutes for the page to load...

If this should be a separate issue, let me know and I can create one.

Issue #29, opened 2015-09-26
Implement API Paging


jordan-wright:
With a test use case of approx. 15,000 targets, the frontend is a bit sluggish.

I believe this can be fixed by implementing API pagination, making all requests asynchronous, etc.

jordan-wright:
Moving this to 0.2 since this will introduce changes to the API.

PR #2019, opened 2020-10-24
Pagination Support! 🥳

glennzw:
## Pagination Support for Large User Groups

I'm super excited with this pull request as I think it adds a feature that's been asked after for quite a while, pagination: https://github.com/gophish/gophish/issues/1971 https://github.com/gophish/gophish/issues/1534 https://github.com/gophish/gophish/issues/816 https://github.com/gophish/gophish/issues/29. This will allow us to have large user groups (>10k) without the browser hanging.

The current datatables configuration downloads the entire user group into the browser, so for groups >10k or >20k (depending on end user RAM) the browser becomes unresponsive. To rectify this I've enabled server side processing for datatables. So far I've only done this for the 'Users & Groups' menu item, but I wanted to get feedback before continuing to implement it in the campaigns section. Click the screenshot below to see a video of me happily browsing 30,000 users:

[Paginationt](https://www.youtube.com/watch?v=z8Oy2DnAZss)

### Client Side
The main client side change was to enable pagination via the `serverSide` parameter, and specifying the data endpoint and data format (see [Server-side processing with DataTables](https://datatables.net/manual/server-side)):

```javascript 
"serverSide": true,
        "ajax": {
                    url: "/api/groups/" + id,
                    'beforeSend': function (request) {
                        request.setRequestHeader("Authorization", "Bearer " + user.api_key);
                    }
                },
        columns: [
            { data: 'first_name', render: escapeHtml},
            { data: 'last_name', render: escapeHtml },
            { data: 'email', render: escapeHtml },
            { data: 'position',  render: escapeHtml },
            { data: null,
                render: function ( data, type, row ) {
                    return '<span style="cursor:pointer;"><i class="fa fa-trash-o" id="' + data.id + '"></i></span>';
                }
            }]
```

Now the datatable sends a request to the server including various URL parameters specifying the start point, number of rows requested, any search terms, and sorting info. I managed to add these requirements into the existing backend API whilst maintaining backward compatibility. Here's a simple example request to fetch some data:

```
GET /api/groups/85?draw=2&start=0&length=10
```
(a full request can be seen [here](https://github.com/gophish/gophish/files/5432027/derp.txt), it's quite verbose)

### Server Side

~I modified `/controllers/api/group.go` Group [function](https://github.com/gophish/gophish/blob/master/controllers/api/group.go#L68) to parse the new datatable parameters, and pass them on to the `models.GetGroup` function (allowing us to specify start, length, search, and order.~
I added a new function `models.GetDataTable` to parse the new datatable parameters and return a result with the appropriate filters. The `draw` parameter is present in all datatable requests (it ensures data is re-assembled in order) so I use this is a flag to determine if we want the new datatable format to be returned, or the old API format. Here's a sample return of the difference:

Original API response to `/api/groups/:id`:
```json
{
  "id": 8,
  "name": "IT Staff",
  "modified_date": "2020-10-24T09:51:26Z",
  "targets": [
    {
      "id": 157284,
      "email": "AlexStokes@einrot.com",
      "first_name": "Alex",
      "last_name": "Stokes",
      "position": "Front-end mechanic"
    },
    {
      "id": 156483,
      "email": "AmyBradley@gustr.com",
      "first_name": "Amy",
      "last_name": "Bradley",
      "position": "Front-end mechanic"
```

New pagination response to `/api/groups/8?draw=7&start=0&length=10&search[value]=mechanic` ([full request](https://github.com/gophish/gophish/files/5432708/req.txt)):
```
{
  "draw": 7,
  "recordsTotal": 999,
  "recordsFiltered": 5,
  "data": [
    {
      "id": 157284,
      "email": "AlexStokes@einrot.com",
      "first_name": "Alex",
      "last_name": "Stokes",
      "position": "Front-end mechanic"
    },
    {
      "id": 156483,
      "email": "AmyBradley@gustr.com",
      "first_name": "Amy",
      "last_name": "Bradley",
      "position": "Front-end mechanic"
    },
........
```

You'll notice other bits and pieces, but this was the main idea to support pagination.

### Group Handling
Next up was modifying the way groups and users are handled. At the moment, editing a group or creating group is handled by holding the entire group in the browser's datatable, and submitting the whole datatable to the server. So, if you have a group with 10,000 users and you want to delete one, you must load all 10k records, select the one to delete, and then submit 9,999 back to the server. 

I added three new API definitions to `controllers/api/server.go`(and gophish.js):

```go
router.HandleFunc("/groups/{id:[0-9]+}/rename", as.GroupRename)
router.HandleFunc("/groups/{id:[0-9]+}/target/add", as.GroupTarget)
router.HandleFunc("/groups/{id:[0-9]+}/target/delete", as.GroupTarget)
```

I also modified the way CSV uploads are handled to suit pagination. At the moment if you upload a CSV file, the file is sent to the server, the server processes it and sends it back in JSON, which is loaded into the datatable, and when you hit save the same JSON is sent back to the server to be saved. So if you have an existing group with 10k users, and want to add another 1k the following happens:
 * Load 10k users into browser datatable from server
 * Upload 1k CSV users
 * Server responds with 1k JSON users
 * Hit Save, and send 11k JSON users to server

For me this often results in the browser hanging. My modification allows a CSV file to be uploaded to the server and inserted directly into the database.

### Known issues

- We can delete the last user on a group, and then the group is marked as having 0 users. Current gophish does not allow this, but I'm not sure if this will be an issue.
- We can't upload multiple CSV files when creating a new group (but you can upload one, save the group, and then upload multiple files). This is because when creating a new group multiple POST requests are sent to the endpoint to create a new group- the first request will create a new group with the supplied name (we use the `name` parameter in the file upload) but the subsequent POSTs will get an error of 'group name exists'. I haven't come up with a solution yet, but am thinking on it.
- CSV uploads of large files hammer the db and gophish (gophish becomes unresponsive). CSV files of 1,000 rows or less load super quick. Need to investigate. Also, once we have >20k users in a group adding additional CSV chunks becomes progressively slower - I think this may need some INDEX investigations.


jordan-wright:
🎉🎉🎉🎉 woohoo! Thanks @glennzw!

I owe you PR reviews. My apologies it's taken so long to get them to you- I've been awfully busy with other projects atm.

Let me see if I can get this reviewed this weekend to make sure we're on the same page regarding general implementation and we can go from there. 

Super exciting!

glennzw:
Hi @jordan-wright! You're welcome and no worries, I know how busy you are!

This pull request was a big learning experience, so there may be rough edges.

Also, I just pushed a fix to properly display the number of results from a search filter, so please make sure you're on the latest commit before you review :)

glennzw:
Some insight on the new commit `Updated how targets are added to a group, improving efficiency`. I was previously adding each target with one db transaction (very slow, just PoC), and have now rolled them into transactions. I borrowed your idea @jordan-wright to use an email cache to look for duplicates. This can be seen in the new function, [AddTargetsToGroup](https://github.com/gophish/gophish/pull/2019/commits/3692fcc680f63bad7f04c51af6928695b0e46b8a#diff-a85e32631c4edd1cd267b723ee96d19215dd1d6b5ee56518b86103610c56a801R360)

One thing I've noticed is that uploads become very slow as the number of targets in a group increases. I created numerous 1,000 user CSV files and uploaded them consecutively via curl. Initially inserting 1,000 users is fairly quick (<10s) but as we creep up it eventually takes almost a minute:

CSVGraph copy

([csv_1k_time_insert.txt](https://github.com/gophish/gophish/files/5453868/csv_1k_time_insert.txt))


I think this is a database issue as if I try re-insert the same .csv file the operation completes in <1s (the cache working very quickly).

Anyway, I'm just mentioning this as a note to our future selves. I think we'd get better utility out the database if we dropped the `GroupTargets` table and put up with the possibility of having duplicate email addresses in the `targets` table (adding `group_id` field to the `targets` table, and maybe having a duplicate key of `email, group_id`).

a-maccormack:
This is great! Just to add to this:

It seems that the changes introduced in this PR caused a failure in `util/util_test.go`. Specifically, the `TestParseCSVEmail` testcase was failing due to the new implementation. Here's a fix to the testcase:

```go
func TestParseCSVEmail(t *testing.T) {
	expected := models.Target{
		BaseRecipient: models.BaseRecipient{
			FirstName: "John",
			LastName:  "Doe",
			Email:     "johndoe@example.com",
		},
	}

	csvPayload := fmt.Sprintf("%s,%s,<%s>", expected.FirstName, expected.LastName, expected.Email)
	r, err := buildCSVRequest(csvPayload)
	if err != nil {
		t.Fatalf("error building CSV request: %v", err)
	}
        
        // start of proposed change
	got, _, err := ParseCSV(r)
        // end of proposed change
	if err != nil {
		t.Fatalf("error parsing CSV: %v", err)
	}
	expectedLength := 1
	if len(got) != expectedLength {
		t.Fatalf("invalid number of results received from CSV. expected %d got %d", expectedLength, len(got))
	}
	if !reflect.DeepEqual(expected, got[0]) {
		t.Fatalf("Incorrect targets received. Expected: %#v\nGot: %#v", expected, got)
	}
}
```

PR #2060, opened 2020-12-05
Added functionality to lock accounts (+bug fix)

glennzw:
A little update that's been useful for me, so maybe it'll be useful for y'all too. This pull request adds functionality to lock accounts:

Create new user or edit existing user:
newuser

Result when logging in with the correct password whilst account is locked:
locked

I also fixed a bug whereby the "Require the user to set a new password" was not being saved to the database when creating new users (it only works when editing an existing user). I'm surprised we hadn't noticed this one yet 😅

I also tweaked the UI such that when you edit a user the "Require the user to set a new password" status is set in the modal from the database (so you know if they'e changed it).

**Note:** I started adding functionality to ensure at least one admin account is unlocked in the [EnsureEnoughAdmins()](https://github.com/gophish/gophish/blob/master/models/user.go#L67) function, but I wanted to check if you thought that was necessary @jordan-wright

### DIFF
--- a/controllers/route.go
+++ b/controllers/route.go
@@ -387,7 +387,7 @@ func (as *AdminServer) Login(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		if u.AccountLocked == true {
-			as.handleInvalidLogin(w, r, "Acount Locked")
+			as.handleInvalidLogin(w, r, "Account Locked")
 			return
 		}
 		u.LastLogin = time.Now().UTC()
--- a/controllers/route.go
+++ b/controllers/route.go
@@ -304,9 +304,9 @@ func (as *AdminServer) nextOrIndex(w http.ResponseWriter, r *http.Request) {
 	http.Redirect(w, r, next, 302)
 }
 
-func (as *AdminServer) handleInvalidLogin(w http.ResponseWriter, r *http.Request) {
+func (as *AdminServer) handleInvalidLogin(w http.ResponseWriter, r *http.Request, message string) {
 	session := ctx.Get(r, "session").(*sessions.Session)
-	Flash(w, r, "danger", "Invalid Username/Password")
+	Flash(w, r, "danger", message)
 	params := struct {
 		User    models.User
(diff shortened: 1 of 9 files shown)


controllers/route.go
```diff
@@ -376,14 +376,18 @@ func (as *AdminServer) Login(w http.ResponseWriter, r *http.Request) {
...
 		// Validate the user's password
 		err = auth.ValidatePassword(password, u.Hash)
 		if err != nil {
 			log.Error(err)
-			as.handleInvalidLogin(w, r)
+			as.handleInvalidLogin(w, r, "Invalid Username/Password")
+			return
+		}
+		if u.AccountLocked == true {
+			as.handleInvalidLogin(w, r, "Acount Locked")
```
jordan-wright, line 390:
```suggestion
			as.handleInvalidLogin(w, r, "Account Locked")
```

Small typo

jordan-wright (changes requested):
This largely LGTM. I added a note about a small typo. One thing I'd ask is if we could add a test to make sure this works as intended. I believe you can find inspiration in `route_test.go`, and it should be a matter of making a new test with the same `setupTest()`, then modifying the user to be locked, then maybe just checking if our message appears in the resulting HTML.

controllers/route.go
glennzw: Good eyes, my bad.

glennzw:
Thanks @jordan-wright ! Ah good  point on the test case, I must remember to include these. I've added logic to add a second user with a locked account and  a new test case `TestAccountLocked()`:


```
Running tool: /usr/local/opt/go/libexec/bin/go test -timeout 30s -run ^TestAccountLocked$ github.com/gophish/gophish/controllers

ok  	github.com/gophish/gophish/controllers	0.287s
```

jordan-wright:
LGTM! Thanks @glennzw!

glennzw:
Cool!

I noticed the minified `users.min.js` wasn't updated in the merge though.

PR #2089, opened 2020-12-30
Add tooltip to the elements of the target table

mitprasoon:
Tooltip has been added to the `First Name`, `Last Name`, `Email` and `Position` columns of the targets table in the group modal while creating and editing a group.
Right now if any of the above-mentioned elements is more than 15 characters in length it is truncated and suffixed with an ellipsis. This at times creates confusion while viewing groups that have targets with
similar first name, last name, email, etc.
With this commit, a small pop-up box with the complete value of the element will appear whenever that element is hovered over by the mouse.
Screenshot 2020-12-30 at 9 03 15 PM

### DIFF
--- a/static/js/src/app/groups.js
+++ b/static/js/src/app/groups.js
@@ -56,8 +56,16 @@ function edit(id) {
     targets = $("#targetsTable").dataTable({
         destroy: true, // Destroy any other instantiated table - http://datatables.net/manual/tech-notes/3#destroy
         columnDefs: [{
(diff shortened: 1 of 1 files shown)


mitprasoon:
Hi @jordan-wright,

Your thoughts on this PR? This is not a major change but greatly increases the usability while creating groups.

Thanks

blacklist-arcc:
Would love to see this feature in gophish!

PR #2110, opened 2021-02-04
Support for SMTPUTF8 (and homoglyph attacks)

olivierlemoal:
Support for SMTPUTF8 will help Gophish users to perform homoglyph attacks.
However, this need both the sender SMTP server and the recipient mail server to support SMTPUTF8 (successfully tested with Office 365 and Gmail). Following Gophish test mail is usurping google.com domain (using the ọ char, which is not converted to punnycode) :

result

It depends on PR https://github.com/gophish/gomail/pull/3 for SMTP implementation.

PR add following option in Sender profile (default is unchecked) : 

image

If support is not enabled on sender SMTP server : 

error

### DIFF
--- a/db/db_mysql/migrations/20210205103800_0.11.0_use_smtputf8.sql
+++ b/db/db_mysql/migrations/20210205103800_0.11.0_use_smtputf8.sql
@@ -1,0 +1,8 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE `smtp` ADD COLUMN use_smtputf8 BOOLEAN;
+
+-- +goose Down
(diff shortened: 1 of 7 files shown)


glennzw:
Hi @olivierlemoal, thanks so much for taking the time to do this, it looks like a pretty cool feature! One quick request, could you remove the minified `sending_profiles.min.js` file please?

I'll find some time to take this and the related https://github.com/gophish/gomail/pull/3 for a spin

olivierlemoal:
Hi @glennzw, I just updated the commit without the minified js.

txr13:
Has there been any further testing on this commit? I also need SMTPUTF8 support.

olivierlemoal:
Closing as no follow- up

PR #2138, opened 2021-03-07
Updated the Ansible Playbook

philomath213:
Hello
I have made some changes to Ansible Playbook.

### Changes

- Some linting on Ansible tasks file.
- Ensure Gophish binary is executable.
- Download the latest release using GitHub API.
- Print logs file after installing Gophish in order to get the initial temporary password.
- Update Ansible Playbook README.

Have a nice day!

### DIFF
--- a/ansible-playbook/README
+++ b/ansible-playbook/README
@@ -1,4 +1,4 @@
-Tested on Ubuntu 16.04.4.
+Tested on Ubuntu 20.04 LTS.
 
 Installs Postfix (to listen on localhost only) and the latest Linux gophish binary.  setcap is used to allow the gophish binary to listen on privileged ports without running as root.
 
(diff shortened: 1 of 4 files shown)


glennzw:
Hi @philomath213, thanks so much for sending this through! I just ran a test to deploy to an Ubuntu 20.04 LTS server and got the following error:

```
TASK [gophish : Install specified packages.] ***************************************************************************************************************************************************************
fatal: [206.189.x.x]: FAILED! => {"changed": false, "msg": "No package matching 'python-pip' is available"}
```

Presumably because Python2 is EOL. I modified `roles/gophish/vars/main.yml` to replace `python-pip` with `python3-pip`, which seems to work now.

Could you verify this fix on your side, and make the change in your commit if this seems right?

Also something I noticed; the playbook outputs the password in the log with the task `TASK [gophish : display log file]`, but then the machine reboots and a new random password is set (so the displayed password is no longer valid and you'd need to go manually check the `gophish.log` file). Not sure if there's an elegant solution to this, but just pointing it out.

philomath213:
Hi,
You are right python-pip is no longer available in recent Ubuntu distros since Python2 is EOL, replacing `python-pip` with `python3-pip` will resolve this issue.

Regarding "gophish" password, I didn't find any adequate solution to get the password other than displaying the log file after setting up "gophish".
We can also add another role that just runs "display log task", so after restarting gophish host, we can run it in order to get the log.

glennzw:
Looks good to me, thanks! 

Let's leave the password issue for now - if it crops up in the issues we can rethink it, but I do like your idea.

isalsaadi07 (approved):

PR #2162, opened 2021-04-08
Implement functionality to support custom RId - Ability to specify both the character set and length

sadreck:
Hi,

I tried to follow all the conventions based on the existing code, but if something is off do let me know and I'll update it.

I've been working on this feature to specify the character set and length of `rid` when creating a campaign:

gophish1

And when the campaign is created, it will appear as:

gophish2

Things I have implemented are:

* For existing tests/functionality, when the character set/length aren't defined it falls back to the default one `[A-Z][a-z][0-9]{7}`.
* Cleans up the character set to remove duplicate characters.
* Compares possible combinations with total recipients to ensure it doesn't get stuck in an infinite loop trying to generate a unique ID.

Thanks,
Pavel


glennzw:
Hi Pavel, thanks so much for taking the time to submit this. Could I ask what you think the demand for this feature would be? I suspect perhaps if savvy users are just reporting any email with "rid=xxxxxxx" in the URL, but I haven't come across that complaint before.

One area that this might cause problems is with reporting; for example the IMAP reporter scans emails for the current rid format.

Overall I think it's a cool idea, but I'm worried (a) it might add unnecessary complexity to gophish and (b) will break the IMAP reporter and any other reporting plugins (e.g Outlook).

sadreck:
Hi,

Like you said the main reason is that the value of `rid` is easily recognisable and to increase the phishing difficulty. I think that real phishing is getting much better than just "look for grammar errors"!

For instance, using Apache rewrite rules you can already mask `rid` in:

* The path name `/download/rid/file.doc`
* Change `rid` parameter into `user` like `/index?user=rid`
* Include `rid` in the subdomain rather than the path like `rid.totallynotphishing.com`. In this instance the default functionality will not work as `rid` would be case-insensitive breaking any visits to `AbCdEfG.totallynotphishing.com` as it would interpret them as `abcdefg.totallynotphishing.com`.

Having the ability to generate only numeric or lowercase and/or different length values, would help a lot in the above scenarios.

It will probably break reporting tools if this is feature is used, but this could be covered in the documentation - or we could add a global setting "Enable Custom RId" and only then this option will become visible.

Issue #1915, opened 2020-07-29
Question/Discussion/Feature: Mark fake/test logins as "false positive"


blacklist-arcc:
What version of Gophish are you using?: 
0.10.1

Brief description of the issue: 
In some situations users detect the fake-landingpage and produce fake/test logins which are counted as valid submissions. The changes in the overall statistics due to fake submissions are significant in small campaigns and change the overall picture.

What are you expecting to see happen? :
A possible solution could be to mark the affected login data as false positive, which will not be counted in the overall statistics. They could also be marked as reported at the same time.

I have not found any related issue to this topic so i want to start a small discussion here. What are your thoughts on this?

Best regards!

jordan-wright:
Hey there!

I can appreciate the desire to do this. It's possible that I'll do something like this in the future, but for now there's a pretty long list of things in front of it.

I'll keep this issue open until I figure out the longterm 👍 or 👎 on this.

Thanks for sending this in!

blacklist-arcc:
Thanks for your fast response. I think we can help to get this feature into gophish. Maybe you can provide your thoughts in which way this can be implemented. We will try to create a PR ;) I think it's nearly impossible without database changes.

JOBERO81:
Hello, I am in training as a programmer.
I would like to try to make this issue as a final project for the training.
I haven't worked with Go yet, but I am learning it for this project.
Even though I don't know if I will get such high quality code reproduced yet, I am very excited about it.
If you have any hints or advice for me I would be very happy about it.
Of course it would be a dream if I could make it so that it can be included in the program later.
Hoping that it is okay for you that I dare to do it as a beginner.
best wishes

blacklist-arcc:
@jordan-wright: @JOBERO81 is working with me on this project. I would be happy if you write a few more thoughts about it so we can start on the right path

glennzw:
Hi @blacklist-arcc and @JOBERO81; if you're still interested in pursuing this feel free to submit a PR. There's some documentation on source code here: https://pkg.go.dev/github.com/gophish/gophish

I'd imagine this solution would require:
 * Database change (adding a new field 'FalsePositiveSubmission' or some such. This would need to be tied to the event.
 * Front end changes to add a toggle button, similar to the current reported button
 * API/backend changes to handle the request from the UI and update the database

blacklist-arcc:
[...]

PR #2208, opened 2021-05-19
implement mark submitted data as false positive, reference #1915

FLX-0x00:
A button was added near the replay button for submitted data in order to mark an event as false positive. Additional API functions  and db modification has beed added. If all submitted data events for one RID are flagged as false positive the status will be changed to "Clicked Link" automatically. With an adiitional checkbox you can mark the user as reported optionally.

ToDo
- minify js & css
- review code =)

### DIFF
--- a/controllers/api/campaign.go
+++ b/controllers/api/campaign.go
@@ -135,3 +135,18 @@ func (as *Server) CampaignComplete(w http.ResponseWriter, r *http.Request) {
 		JSONResponse(w, models.Response{Success: true, Message: "Campaign completed successfully!"}, http.StatusOK)
 	}
 }
+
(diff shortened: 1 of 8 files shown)


MoellerTom:
<h3>New Button that was added to the Timeline</h3>

False Positive button

<h3>Swal that is shown depending on the report-status of the current record</h3>

Swal if reported
Swal if not reported

<h3>Event that was marked as False Positive will be shown as usual but with an Alert that it was marked</h3>

marked Event Alert

<h3>If all events (for that recipient) where Data was submitted are flagged as False Positive the status for the recipient will be changed and the count will be corrected</h3>

Status after all events are flagged
results

blacklist-arcc:
@glennzw wanted to kindly ask if there are plans to implement/review the code

PR #2213, opened 2021-05-27
Security fix API key

ArminBerberovic:
If the user opened several tabs and logs out in one, it is still possible to issue api calls from the other tabs, i.e. add, delete or modify landing pages, users, etc. This behaviour ca be prevented by using the login cookie to authorize API calls. This patch implements this by translating the login cookie to the API key on the server side. It also prevents the transfer of the API key from the server to the client in most places (except settings page) and modifies the client requests so they don't contain the API key.


eickeh:
Hi there, are there plans to merge this any time soon?

glennzw:
Hi @ArminBerberovic and @eickeh - thanks for sending this through, it's always great to have multiple eyes on the security front! As a starting point before I go through the code in detail could you please remove the minified js files? We like to rebuild those from the unpacked versions. Thanks again!

eickeh:
Hi @glennzw, thanks for getting back to this PR! I have no access but @ArminBerberovic will change it next week probably.

ArminBerberovic:
Hey @glennzw, thanks for coming back to us. I've removed the minified js files

joao-intelliway:
You need to remove the TestBearerToken function from the tests

PR #2233, opened 2021-06-15
Add feature to export group targets as CSV

mitprasoon:
A new feature has been added through which targets in a group can be
exported as CSV. For this a new button has been added to the group page
for every group. On clicking this button for a group, group targets will
be downloaded as CSV file.

This will enable an admin to download group targets from one system and
import them on other to create a group.


vonjk:
I tried this feature,  which is very useful, but if Position is Chinese, the export CSV will be garbled。。

joao-intelliway (approved):

PR #2389, opened 2022-01-29
Fix sending profile form (fixes #1986)

### LINKED ISSUE #1986: Password suggestions in Search box for Sending Profiles
Running latest gophish.

Whenever I go to the 'Sending Profiles' tab Chrome tries to autofill username suggestions into the Search box:

Screenshot 2020-09-23 at 20 26 57

This is super annoying, as even when I delete the autofilled username it pops back in. It's only happening on the search box for 'Sending Profiles', and I can't see anything in the HTML element that would entice chrome to fill this particular field.

Doing some Googling I think this might be a Chrome bug: https://support.google.com/chrome/thread/6787468?hl=en

I wanted to check in here though and see if anyone else runs into this, or has any possible fixes.

marksteward:
Wrap all inputs on the New Sending Profile page in a `<form>`, as bootstrap and HTML5 advise. This fixes the annoying behaviour in Chrome where it tries to autofill the search box as a username for the password field. Other pages are also missing forms, but they don't have this problem because they don't have password fields.

I've used `<label class="input-group">`, which is strictly wrong, but the correct `<legend>` would be much larger. Bootstrap 4 will make it easier to make the legend a sensible size.

As a side-effect, these changes stop enter from adding a custom email header (as the button is no longer the submit for the form). Instead, a user must tab into the Add Custom Header button before pressing enter/space. I think this is less surprising, as enter for a popup form typically submits the entire form. I can change this if necessary.

### DIFF
--- a/static/js/src/app/sending_profiles.js
+++ b/static/js/src/app/sending_profiles.js
@@ -307,7 +307,7 @@ $(document).ready(function () {
         dismissSendTestEmailModal()
     })
     // Code to deal with custom email headers
-    $("#headersForm").on('submit', function () {
+    $("#addCustomHeader").on('click', function () {
         headerKey = $("#headerKey").val();
         headerValue = $("#headerValue").val();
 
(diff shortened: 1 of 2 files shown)


glennzw:
Thanks very much @marksteward, I've finally figured out a way to test this out (tricky with self signed certs etc) and it looks good; no more usernames in the Search box. I'll go ahead and merge this.

Thanks again for taking the time to submit a fix to this really annoying bug!

PR #2482, opened 2022-05-26
Fix resource selection during campaign copy

29vivek:
Clear the selection of resource (template, page, profile) whenever the original resource is deleted and there is
only one currently available resource present in the DB while copying a campaign. Without this fix, the only
available resource is shown as the original resource, instead of showing [Deleted].


glennzw:
Great catch @29vivek, I can confirm I see the issue on my side and that your fix works. Please could you remove the `.min.js` version as we like to create those on our side - and I'll get this merged in!

Thank you for your patience with us getting to this.

29vivek:
Hi @glennzw, I've removed the `.min.js` version. Thanks!

PR #2507, opened 2022-06-20
Update README.md

Parrishjm:
Very small update to change go get... to go install... per go documentation.


glennzw:
Ah thanks @Parrishjm - I'd actually just made this change from an issue that was opened by someone else.

Either way hopefully the documentation is correct now!

PR #2599, opened 2022-09-14
Fix new records being added on completing a campaign

29vivek:
There were new records with name '[Deleted]' being added when a campaign was completed. This used to happen when the resource associated with a campaign (template, page, profile) was deleted before marking the campaign as completed. The save gorm call used to upsert these values and ended up adding rogue records.

### DIFF
--- a/models/campaign.go
+++ b/models/campaign.go
@@ -661,7 +661,8 @@ func CompleteCampaign(id int64, uid int64) error {
 	// Mark the campaign as complete
 	c.CompletedDate = time.Now().UTC()
 	c.Status = CampaignComplete
-	err = db.Where("id=? and user_id=?", id, uid).Save(&c).Error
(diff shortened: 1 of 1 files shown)


glennzw:
Hi @29vivek - good catch, thanks for submitting this. I'm trying to reproduce the bug on my side. Where did you see the `[Deleted]` records?

29vivek:
The `[Deleted]` records are populated in the DB. They aren't visible in the frontend since they aren't associated to a user.

glennzw:
Ah I see - in each table for the given template, etc. I see after deleting an email template and then marking the campaign as Complete this row at the bottom:

```sql
sqlite> SELECT * FROM templates ORDER BY ID DESC limit 1;
327|0|[Deleted]||||0001-01-01 00:00:00+00:00|
```

29vivek:
Yes, this is the one!

glennzw:
LGTM, thanks @29vivek!

PR #2631, opened 2022-10-19
Add configurable list of filetypes that allow substitutions in attachments

mbharanya:
Currently it is possible to add templated microsoft office files as attachments to simulate macro viruses. 
In our use case we also need to be able to send other plain-text executable files such as `.bat` and `.ps1` as an attachment with working template rendering. As this use case differs for each user, this pull request will add the following new configuration to the `config.json` file:
```json
{
...
	"attachments": {
		"plain_text_file_list": [
			".txt",
			".html",
			".ics",
			".ps1",
			".bat",
			".vbs",
			".sh",
			".py"
		]
	}
...
}
```
This allows the user to set a list of allowed files where templates are rendered in attachment files.
As this is my first PR to this repo I am glad for any feedback, please let me know if any of it does not match the current style or naming scheme.

### DIFF
--- a/config.json
+++ b/config.json
@@ -19,5 +19,17 @@
 	"logging": {
 		"filename": "",
 		"level": ""
+	},
+	"attachments": {
+		"plain_text_file_list": [
+			".txt",
+			".html",
+			".ics",
+			".ps1",
+			".bat",
+			".vbs",
+			".sh",
+			".py"
+		]
 	}
 }
--- a/config/config.go
+++ b/config/config.go
@@ -26,6 +26,11 @@ type PhishServer struct {
 	KeyPath   string `json:"key_path"`
 }
 
+// Attachments represents the handling of attachments in emails
+type Attachments struct {
+	PlainTextFileList []string `json:"plain_text_file_list"`
+}
+
 // Config represents the configuration information.
 type Config struct {
 	AdminConf      AdminServer `json:"admin_server"`
@@ -37,6 +42,7 @@ type Config struct {
 	TestFlag       bool        `json:"test_flag"`
 	ContactAddress string      `json:"contact_address"`
 	Logging        *log.Config `json:"logging"`
+	Attachments    Attachments `json:"attachments"`
 }
 
 // Version contains the current gophish version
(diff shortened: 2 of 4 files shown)


mbharanya:
As an example this powershell script will be substituted with the correct user-specific URL and issue a Web-Request to the web server, which can then be included in statistics:
```ps1
Invoke-Webrequest  {{.URL}}&executed=ps1
```

is transformed to:

```ps1
Invoke-Webrequest  http://localhost?rid=dj9YRIs&executed=ps1
```

mbharanya:
@glennzw I hope it's okay to mention. Did you get the chance to take a look at this? Much appreciated!

glennzw:
Hi @mbharanya - thanks so much for taking the time to submit this PR! It's a pretty good idea. Let me give it some thought; I'm wondering if the `config.json` is the best place to add this.

Also, beyond the list you added could you imagine any further filetypes? If that list seems exhaustive it might just be worth hard coding them.

Another approach would be to have an option when attaching a file as to whether or not it should be templated.

mbharanya:
Thank you for looking at the PR @glennzw
I also thought about providing an exhaustive list in the code, like it is done for the office files. But I think this usecase may vary a lot from user to use. Also as new file types arise, putting it in the config.json will allow flexibility. Some other options might be  program-specific files or a lot of other ones (https://fileinfo.com/filetypes/executable). 

Also the way it is currently implemented, users won't need to add anything to the `config.json` file, as it will fall back to the current behaviour, replacing no text.

>Another approach would be to have an option when attaching a file as to whether or not it should be templated.

This is also a good idea for sure. Which option would you prefer?

mbharanya:
@glennzw Did you have the chance to look at this (and my other PR #2651) so far? It would be really helpful for us, if you could consider merging it to the main branch, as we rely on this feature internally at my company. As I said, if you have any concrete feedback, I'd be happy to implement it :)

joao-intelliway:
I agree that hard coding is the best approach. If new file types are created anyone can create an issue or pull request to add them.

compr00t:
@glennzw Any progress on merging this soon?

PR #2637, opened 2022-10-20
fix Chinese attachment filename

lsr00ter:
fix the bug of Chinese attachment filename encoding error.

### DIFF
--- a/models/maillog.go
+++ b/models/maillog.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"encoding/base64"
 	"crypto/rand"
 	"errors"
 	"fmt"
@@ -358,6 +359,7 @@ func addAttachment(msg *gomail.Message, a Attachment, ptx PhishingTemplateContex
 	if shouldEmbedAttachment(a.Name) {
 		msg.Embed(a.Name, copyFunc)
 	} else {
+		a.Name = "=?UTF-8?B?" + base64.StdEncoding.EncodeToString([]byte(a.Name)) + "?="
 		msg.Attach(a.Name, copyFunc)
 	}
 }


joao-intelliway:
Hello, can you add a test to your pull request?

XiaoliChan:
Can confirm this commit fix the chinese attachment filename issue.

Before this patch, as you can see the attachment is not doing base64 encoding, this will be showing wired strings in some email clients.

- Email client (showing wired strings):
image

- EML file (showing correct strings):
image

After this patch, the filename will be doing base64 encoding stuff, then decode in HTML web client.

- Email client (showing correct strings):
image

- EML file (showing base64 encoded strings):
image

Alazyworm233:
Hello, after modifying the maillog.go file, I recompiled the program. Unfortunately, a new bug occurred when sending attachments with Chinese names. What could be the reason for this?

XiaoliChan:
> Hello, after modifying the maillog.go file, I recompiled the program. Unfortunately, a new bug occurred when sending attachments with Chinese names. What could be the reason for this? 
Not really sure what happened

Alazyworm233:
The "=?UTF-8?B?" is lost.

XiaoliChan:
> The "=?UTF-8?B?" is lost. 

Looks like it lost. XD

Alazyworm233:
So,why???????

XiaoliChan:
>  So,why???????

Can you try with `163`, and `Gmail`? probably something has been filtered by `Tencent enterprise e-mail`

Alazyworm233:
QQ mailbox
What solutions are there?

XiaoliChan:
> QQ mailbox What solutions are there?

It should be worked on `QQ mailbox`, but your screenshot looks like is `Tencent enterprise email`.

Not sure what happened with that...

Alazyworm233:
I don't know either, but I noticed that the 'utf-8' was missing in the email.

Alazyworm233:
I think I know where the problem is now. Tencent's email seems to have a limitation on the length of file names, which causes excessively long file names to be truncated.

Pearce-cmd:
> I think I know where the problem is now. Tencent's email seems to have a limitation on the length of file names, which causes excessively long file names to be truncated.

I also encountered the same problem. If you delete the numbers in the file name, you can display Chinese normally, but I don’t know why.

awei84:
> > 我想我现在知道问题出在哪里了。腾讯的邮件似乎对文件名长度有限制，导致过长的文件名被截断。
> 
> 我也遇到了同样的问题。如果删除文件名中的数字，可以正常显示中文，但不知道为什么。

image

awei84:
image
I seem to have solved this problem before, what about you trying this?

awei84:
> > I think I know where the problem is now. Tencent's email seems to have a limitation on the length of file names, which causes excessively long file names to be truncated.
> 
> I also encountered the same problem. If you delete the numbers in the file name, you can display Chinese normally, but I don’t know why.

I seem to have solved this problem before, what about you trying this?

PR #2667, opened 2022-11-22
Updated Dockerfile

socketz:
Suppressed npm warnings and fixed CMD to ENTRYPOINT.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN npm install gulp gulp-cli -g
 
 WORKDIR /build
 COPY . .
-RUN npm install --only=dev
+RUN npm install --loglevel=error --only=dev
 RUN gulp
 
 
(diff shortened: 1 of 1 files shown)


glennzw:
Hi @socketz - thanks for taking the time to submit this. Could you elaborate on the difference between using ENTRYPOINT over CMD?

socketz:
Hi @glennzw, CMD commands are ignored by daemon when there are parameters stated within the docker run command while ENTRYPOINT instructions are not ignored but instead are appended as command line parameters by treating those as arguments of the command.

You can read more differences in official documentation:

https://docs.docker.com/engine/reference/builder/#cmd

https://docs.docker.com/engine/reference/builder/#entrypoint

Regards!

joao-intelliway (approved):
https://phoenixnap.com/kb/docker-cmd-vs-entrypoint

Issue #2624, opened 2022-10-12
 501 Error: Bad sender address syntax


syh4ck:
Hey Guys Im facing an issue with the latest version of gophish, 

The sending profile MAIL from Input : `First Name <test@example.com>`, 

Even after entering in above format I'm getting 
 501 Error: Bad sender address syntax

Need help ?

glennzw:
I've seen this a few times recently; I'm guessing you're using Outlook / O365? Can you try with just `test@example.com` and let me know if that works?

Do older versions (e.g 0.11) work?

syh4ck:
Hi Glennzw,

Both approaches I tried didn't work. I finally got the previous version and am now using it.

glennzw:
Would you be able to help me get to the bottom of this issue in the newer versions?

If you could share which email provider you're using and I'll try replicate it on my side. I think it's probably got to do with this PR: https://github.com/gophish/gophish/pull/2334

Further discussion [here](https://github.com/gophish/gophish/pull/2334), where Mailgun seems to be the culprit.

righter83:
I've got the same on my own postfix mailserver:

`warning: Illegal address syntax from mail[2a03:4000:3e:XXX in MAIL command: <XXX XXX <XXXX@XXXX>>`

drego85:
Same problem here with last version 0.12.1:

```
level=warning msg="501 <Support Test <support@EXAMPLE-DOMAIN.com>>: \"@\" or \".\" expected after \"Support\"" code=501 email="\"Andrea D\" <andrea.d@xxxx.net>"
level=error msg="501 <Support Test <support@EXAMPLE-DOMAIN.com>>: \"@\" or \".\" expected after \"Support\""
```

Therefore it is not possible to specify the textual name sender of the email.

troplolBE:
I got the same issue on my side and did some research. The issue was introduced in the #2334 PR. I was able to work around it by just changing the code and use the old `send` function instead of the new `SendCustomFrom` function.

I did some debug and I saw that the smtp_from variable in the picture here under does not contain the surrounding `<>` characters. It must been added afterwards but I haven't found where.

https://github.com/gophish/gophish/blob/cec2da512890e4e5fe893aaf2a223a501408370d/mailer/mailer.go#L171

Hope this helps.

troplolBE:
Hi @glennzw @syh4ck 

[...]

PR #2669, opened 2022-11-23
Modified "SMTP From" field to avoid SMTP server errors

troplolBE:
This pull request has been opened following the MAIL FROM issues discussed in the #2624 issue. This pull request modifies the html field  `SMTP From` to stop suggesting the users he can add more than one email address. It also implements a check to be sure that the content of the field is an email address by making use of a regex. This modifications also ensures we are complying with [RFC 5321](https://www.ietf.org/rfc/rfc5321.txt)

End users should use the Sender Envelope in the Email template instead.

### DIFF
--- a/models/maillog_test.go
+++ b/models/maillog_test.go
@@ -284,7 +284,7 @@ func (s *ModelsSuite) TestMailLogGenerateOverrideTransparencyHeaders(ch *check.C
 	smtp := SMTP{
 		Name:        "Test SMTP",
 		Host:        "1.1.1.1:25",
-		FromAddress: "Foo Bar <foo@example.com>",
+		FromAddress: "foo@example.com",
 		UserId:      1,
 		Headers: []Header{
 			Header{Key: "X-Gophish-Contact", Value: ""},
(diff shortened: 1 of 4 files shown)


troplolBE:
I will put the PR in draft because I first need to adapt the test suite to suit this new change

glennzw:
Thanks so much for taking the time to bug hunt and submit this @troplolBE! Looks like a good approach. I was wondering if we should use `mail.ParseAddress()` to validate the address rather than relying on regex but it looks like that will accept the `First Last <first.last@foo.com>` that we're trying to avoid.

I'll take this for a proper spin but LGTM so far!

troplolBE:
Hey @glennzw 

I was thinking of using that function too but the problem is that it will validate `First Last <first.last@foo.com>` so we would need to grab the email part but to me we can definitely use the default function. Let me know if you want me to change this !

glennzw:
No I think your regex is suitable actually - I've taken your fix for a spin and all looks great. Thanks so much @troplolBE!

PR #2683, opened 2022-12-04
Changing the default phishing page with html file

Aporlorxl23:
Hi,

I changed the code so we can use an HTML file whenever we want instead of the "404 page not found" we usually see. The user adds the codes to the index.html file and can use it as the default phishing page. As an extra, an empty file called gophish/index.html needs to be created.

 
 

Thanks,
Eren Şimşek

### DIFF
--- a/controllers/phish.go
+++ b/controllers/phish.go
@@ -8,7 +8,9 @@ import (
 	"net"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


joao-intelliway:
Why you didn't used?
`http.ServeFile(w, r, URL)`

Issue #1843, opened 2020-05-21
QR Codes


dryahweh:
Hi Jordan!

Thanks for all the effort you put into GoPhish! I have a question related to using QR Codes with GoPhish.

Some of my clients are asking if we can use QR Codes instead of URLs. We have tried using Dynamic QR Codes to register victims scanning the code, yet the result does not provide me the email that scans nor the IP (although the IP is not as important as the email). Normally GoPhish autogenerates a link with an RID to track the target who clicked the link.

Is there a way of using a QR to replace the URL in this scenario? This would result in the event being picked up by the GoPhish listener without the user clicking, but scanning.

Thanks in advance for the help!

S0larflare:
I can’t see any particular issue. On the face of it I think it would be pretty simple, just use something like this:

https://github.com/skip2/go-qrcode/blob/master/README.md

And instead of including the URL include the generated qrcode. I’m not sure how widely used it would be, but I think it would be a reasonably simple addition to GoPhish

Sent from my iPhone

On 21 May 2020, at 16:16, dryahweh <notifications@github.com> wrote:

﻿

Hi Jordan!

Thanks for all the effort you put into GoPhish! I have a question related to using QR Codes with GoPhish.

Some of my clients are asking if we can use QR Codes instead of URLs. We have tried using Dynamic QR Codes to register victims scanning the code, yet the result does not provide me the email that scans nor the IP (although the IP is not as important as the email). Normally GoPhish autogenerates a link with an RID to track the target who clicked the link.

Is there a way of using a QR to replace the URL in this scenario? This would result in the event being picked up by the GoPhish listener without the user clicking, but scanning.

Thanks in advance for the help!

glennzw:
Agreed with S0larflare. You could:

 * Create a campaign, in your emails insert an image with src `http://<youserver>/qr/{{.rid}}.jpg`
 * API into the campaign grabbing all the URLs
 * Foreach email use the QR library above to create `{{.rid}}.jpg`
 * Copy all the generated QR images into your server's webroot

Only small snag is some email clients may not display remote images. Alternative plan is to use the API to insert campaign emails with base64 embedded QR images.

jordan-wright:
Thanks for reaching out! There's no way to generate QR codes in Gophish emails right now. More manual approaches like the one @glennzw suggested work. If you were wanting to generate the QR code in a landing page, it'd be different since you could use a Javascript library [this one.](https://davidshimjs.github.io/qrcodejs/)

[...]

PR #2809, opened 2023-03-24
Add unicode QR code as URL_QR variable in mail template

dmarushkin:
Hi team!

Please consider small change that add ability to add in email template text unicode QR code instead of URL in clear text.

It inspired by this thread https://stackoverflow.com/questions/19673303/formatting-qr-codes-in-html-emails-using-the-pre-tag/75771797 and tested in assessment with "Check our new mobile ChatGPT Internal Support Bot in Telegram" scenario, where internal outlook blocked external image link with QR.

Best regards and thanks for great app!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/PuerkitoBio/goquery v1.5.0
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d // indirect
-	github.com/dmarushkin/go-qrcode/v2 v2.0.0-20230616071929-fef537719631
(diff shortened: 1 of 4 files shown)


glennzw:
Hi @dmarushkin, thanks for taking the time to submit this, it's a cute idea. I'm hesitant to role it into core gophish though as it's one more external library for quite a limited use case. See my solution here for inserting a QR code image:

https://github.com/gophish/gophish/issues/1843#issuecomment-634730316

dmarushkin:
Hi Glenn!
I've tried external QR image links but that was not shoot in my case, because corp outlooks block loading images from external sources for untrusted senders. So I've come to solution with generating text QR in gophish, actually for now I've found that generating qr with div blocks looks much better than unicode symbols.

That's a good point about one more external library, and I can bring QR generating logic to a function inside code of gophish, if you don't mind...

PR #2816, opened 2023-03-29
General updates & mysql8 support without custom DB settings 

gavinelder:
👋 This PR is pretty large and touched a couple of items.

Mainly DEP updates everywhere possible outside of the Email Sending which needs a whole bunch of re-factor.

Support mysql8 without DB opts moving from `time.Time` -> `*time.Time` to allow NULL values & updating the DB tables to allow default NULL along with some migration scripts which will need some testing.

Would appreciate someone looking at the migrations in general given we don't have a polluted DB.


joao-intelliway (approved):
The code updates seems fine and the campaign creation really has the logic error

PR #2864, opened 2023-05-16
Security fix api key rebased

eickeh:
There is an [old pull request here](https://github.com/gophish/gophish/pull/2213) from me and a former colleague of mine with the same intention. As I can't reach him anymore and don't have access to his repo, here comes an updated one. 

Original PR text:

> If the user opened several tabs and logs out in one, it is still possible to issue api calls from the other tabs, i.e. add, delete or modify landing pages, users, etc. This behavior can be prevented by using the login cookie to authorize API calls. This patch implements this by translating the login cookie to the API key on the server side. It also prevents the transfer of the API key from the server to the client in most places (except settings page) and modifies the client requests so they don't contain the API key.

My additions:
- Reintroduced checking of the Bearer Token as this probably would have broke the api
- Rebased everything on the current master branch

Please tell me if I can do anything further. Would love to see this upstream :tada:


eickeh:
Did anyone look at this yet? It's actually not much code =)

PR #2850, opened 2023-05-02
adding environement var DB_NAME to run.sh

rpsteinbrueck:
Adding environment variable DB_NAME to run.sh so that Gophish Container can be set up with mysql/mariadb.

db_name has to be changed in config to mysql for mysql connection to work.

PR #2878, opened 2023-06-01
Addition to docker run.sh

sn0b4ll:
Added db_name to run.sh in order to be able to switch to mysql via env var. This is needed because I run the docker-container in an evironment which needs the config to be set via environment and I, of course ;), want to use mysql for production.


sn0b4ll:
Closing because #2850 does the same, sorry for the duplicate.

PR #2882, opened 2023-06-05
General update for this project

qy-zh:
Hi there,
my employer sponsored quite some time to update this project, which we would like to share now with you.
As you can see from our repository CI passes and most of the security issues have been addressed.
In my humble opinion, the project should be up-to-date now.
I appreciate any feedback and it would be great if you accept my PR - of course after checking it thoroughly.
Best,
Qiyue Zhang


jordan-wright:
Hi there!

This is a _significant_ number of changes in a single PR. So much so that it would be very difficult to adequately review them.

Please consider opening atomic PRs that make a small number of changes at a time so that we can more easily review these and see about getting these merged.

Please thank your employer on our behalf for sponsoring this work!

Issue #2946, opened 2023-08-16
Suggestion: Show number of potential emails to send before starting campaign


FLX-0x00:
I think it would be nice to show the number of emails that will be send by the phishing campaign right before it starts.

Just right in this window. Maybe gophish can calculate the right number from the email groups.

![image](https://github.com/gophish/gophish/assets/26464774/fb28881a-360a-48be-ab1f-52c4cd290755)

glennzw:
That's a nice idea, thanks @FLX-0x00 ! Even better if it's all done in Javascript.

FLX-0x00:
There are so many times when you want to test the email with a few users and not double check the selected group and then hundreds of emails fly away :D

WigzyDev:
Great idea! I have also launched campaigns to the wrong amount of people while testing before.  Submitted a PR to add this.

FLX-0x00:
Hopefully the PR will merged. We have implemented a false positive button switcher to mark submissions as false positive, but no luck with merging since 2021 :/ 

no hard feelings - we use a custom compiled version and we will merge this PR in our internal version =) Thanks @WigzyDev  for your awesome contribution

WigzyDev:
> Hopefully the PR will merged. We have implemented a false positive button switcher to mark submissions as false positive, but no luck with merging since 2021 :/
> 
> no hard feelings - we use a custom compiled version and we will merge this PR in our internal version =) Thanks @WigzyDev for your awesome contribution

All good. That's unfortunate that PR's aren't actively being merged. We also went the route of the custom compiled internal version with a bunch of updates.

PR #2964, opened 2023-09-06
Added total number of targets to campaign launch modal. enhancement #2946

w1gs:
This PR adds the total amount of targets to the launch campaign confirmation modal. Adds enhancement described at  #2946 .
Example: 
image


FLX-0x00:
The min.js should not be part of a PR as reviewing this will cost too much time and there is too much potential for backdoors. The minifiy process will be done by the maintainers

WigzyDev:
> The min.js should not be part of a PR as reviewing this will cost too much time and there is too much potential for backdoors. The minifiy process will be done by the maintainers

That makes sense. Thanks for letting me know. Is there a standard on formatting? Didn't see anything in the contributing.md and vscode applied my user settings formatting. The actual code for this PR is only a few lines in the unminified JS but the whole file was auto formatted.

FLX-0x00:
If the maintainer wants to use special formatting, the .editorconfig file should be in the repo or a linting pipeline as a check. Can't see any of this here that indicates a general formatting convention. Thanks again for your work on this project!

FLX-0x00:
maybe keep the pr open. just in case :)

Issue #2978, opened 2023-09-20
Avoid google repot domain


thomasisaacsibille:
**Thanks for reaching out! We're happy to help resolve issues as quickly as possible.**

**Please use this template when creating a new issue. If you do not follow this template format, your issue may be closed without comment.**

Before filing a new issue, please use the search bar at the top of the browser to search for similar issues. Also, please make sure you have read our [documentation](https://getgophish.com/documentation/), which covers how to use Gophish.

What version of Gophish are you using?: 0.10.1

Brief description of the issue: 
We just started phishing with Gophish. But a few minutes after Gmail received the email. Some users report email as spam ( which is very good) and Google blocks the gophish domain. 
We could install Gophish with wildcard DNS (*.example.com) and generate a random subdomain for all emails. 
This will avoid blocking phishing campaign when a user report as phishing.

glennzw:
Hi @thomasisaacsibille, great question. You _should_ be able to randomise the subdomain with the per-user {{.RId}} parameter in the URL like so:


<img width="228" alt="image" src="https://github.com/gophish/gophish/assets/3966613/20e3e569-9d97-405f-8788-4438316e8006">


However this isn't working for me, although I can use all other fields simple fields, e.g:


<img width="522" alt="image" src="https://github.com/gophish/gophish/assets/3966613/94960f4c-8953-4529-9c46-271fa3758d90">

...which results in:


<img width="777" alt="image" src="https://github.com/gophish/gophish/assets/3966613/76d65c7c-feac-4866-a0c0-bc25b83dc2be">

When using the {{.RId}} parameter in the subdomain I get the error:

<img width="618" alt="image" src="https://github.com/gophish/gophish/assets/3966613/96f5a026-31d3-4f48-a55b-b7f84534e1aa">

I'd be interested to hear anyone else's guesses on this. My one suspicion is that perhaps the RId parameter hasn't been created yet as it's the only dynamic one.

If so, a solution for you could be to randomize the "position" field. The easiest way would be to load up all your users in Excel and give the position column random values. A formula you could use for that would be:

`=CHAR(RANDBETWEEN(97,122))&CHAR(RANDBETWEEN(97,122))&CHAR(RANDBETWEEN(97,122))&CHAR(RANDBETWEEN(97,122))&CHAR(RANDBETWEEN(97,122))`

The CSV file would then look like so:

[...]

PR #2979, opened 2023-09-20
feat: allow uniq identifier into phish url

thomasisaacsibille:
This PR is for isse #2978 

When set * into URL, it will be replace by RId in order to avoid google URL protection.


thomasisaacsibille:
@glennzw Any chance to check this PR ?
Thanks

PR #3062, opened 2024-01-02
Add option to read version from environment

flyn-org:
Gophish currently requires a file named VERSION in its working directory. This adds the option to read the version from the environment instead.


flyn-org:
After some time, the OpenWrt folks suggested another approach. I will propose another patch later, depending on the outcome with OpenWrt.

Issue #212, opened 2016-03-16
Cannot import site due to proxy server in use


plex3r:
When I try to import a site it fails, I assume this is because I have to use a proxy to connect to the internet?

jordan-wright:
Do you see any specific error messages that could indicate the cause for the error?

plex3r:
I didn't see any errors when trying to do this it just times out, what I ill try is connecting to a direct internet connection and see if it then succeeds.

![landing_page_failure](https://cloud.githubusercontent.com/assets/4057454/13841260/b666cf84-ec1f-11e5-835b-2ea0c814199c.PNG)
![gophish_terminal](https://cloud.githubusercontent.com/assets/4057454/13841263/b9651772-ec1f-11e5-99f3-9b22318dfb3e.PNG)

jordan-wright:
Hey @plex3r - just checking in, any luck with this?

jordan-wright:
Hey @plex3r - just checking in one more time, any luck with this?

plex3r:
Hi Jordan,

Sorry haven't got chance to test yet.

I'll do it today.

plex3r:
@jordan-wright just tried it on an adsl connection and it works fine so it's got to be something to do with going through a proxy server at work to access the internet.

Hope this helps.

jordan-wright:
You bet! I might go ahead and close this for now. Proxy issues can be a huge pain to diagnose.

Since you were able to get it working, it at least looks like your gophish installation is good to go.

Thanks for the update!

plex3r:
@jordan-wright I've no idea what might be involved in support web proxies but I'd guess a lot of companies that are going to use this as an internal tool will have them and yes they may be able to use an alternative connection if it's a big issue to overcome.

jordan-wright:
I completely agree. In this case, it looks like something in the proxy was preventing outbound requests, which is odd, but will be difficult to track down from our end.

Did you see the same behavior trying to import HTTP sites, or only HTTPS sites?

Also, you might try setting the "HTTP_PROXY" environment variable to your companies proxy and see if that helps?

rfdevere:
This used to be an issue on every product/software I ran, issues getting it from my work machine to an external IP. If data policies permit it where you are just run things off a VPS and SSH onto it. You have 0 issues with getting access, although at the expensive of having your data 'out there'

jordan-wright:
Going to go ahead and close this for now. We will continue evaluating solutions if we encounter other issues.

[...]

PR #3076, opened 2024-01-15
Add support for HTTP_PROXY into the import functionality

Oneiroi:
As noted in #212 import functionality fails when an egress proxy is in use, this pull requests addresses this by adding in support for detection of an environment variable `HTTP_PROXY` if this is detected this value will be used as a HTTP proxy.


Oneiroi:
Moving into a branch and resubmitting ...

PR #3077, opened 2024-01-15
Add HTTP_PROXY support for import

Oneiroi:
As noted in #212 import functionality fails when an egress proxy is in use, this pull requests addresses this by adding in support for detection of an environment variable HTTP_PROXY if this is detected this value will be used as a HTTP proxy.

### DIFF
--- a/controllers/api/import.go
+++ b/controllers/api/import.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
+	"os"
 	"strings"
 
 	"github.com/PuerkitoBio/goquery"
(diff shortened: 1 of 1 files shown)


jordan-wright (changes requested):
I'm generally fine with adding proxy support, but I think there's a cleaner way we can do this!

WDYT about updating this PR to be a single line addition to the `http.Transport{}`:

```go
tr := &http.Transport{
	DialContext: restrictedDialer.DialContext,
	Proxy: http.ProxyFromEnvironment,
	TLSClientConfig: &tls.Config{
		InsecureSkipVerify: true,
	},
}
```

The `http.ProxyFromEnvironment` function is how the default transport "just works" with `HTTP_PROXY` (and other!) environment variables.

Oneiroi:
Closing in favour of #3336

Issue #3117, opened 2024-03-26
Complete campaign endpoint is a GET instead of a POST


Caetan:
What version of Gophish are you using?: 
Version 0.1.0

Brief description of the issue: 
The endpoint for complete a campaign is a GET instead of a POST, even thougth the comment stats it's a POST

What are you expecting to see happen? :
https://github.com/gophish/gophish/blob/8e79294413932fa302212d8e785b281fb0f8896d/static/js/src/app/gophish.js#L105
```
return query("/campaigns/" + id + "/complete", "GET", {}, true)
```

I guess GoLang code should be also modified

 Cheers!

ttinkywinky45:
**Hi 👽 ! If you are interested in ⚙️Advanced Phishing⚙️ and the ✨Evilginx3✨ tool then I offer you Google Gmail and Custom Phishlets (Paypal, Cryptos...) with a ✅Free Tutorial✅ which arrives if the 📈Telegram reaches 50 Subscribers📈
⏳ GOOGLE GMAIL PHISHLET 2024 WITH PROOF ⏳
 ->
⛔https://www.youtube.com/watch?v=nA47UknIE6A⛔
---
🔓 Ask your questions and a free tutorial from 50 subs on telegram !!!
->
📩https://t.me/Noremorse1011📩**

![evilginx_background](https://github.com/kgretzky/evilginx2/assets/144027798/0e448f68-ecdc-49a4-a4f4-23dbed7cb971)

brite107:
Can I be assigned please? This would be a great first issue.

AndrusGerman:
Open a pull request to fix this @Caetan

Hosted-EU:
There’s A New Phishing Tool Out There.
It Can Generate URLs For Phishing Pages.
And Able To Steal Cookies from Phishing Pages To Bypass 2FA 🔥

Watch Video On How It Works : https://youtube.com/watch?v=BJv3jFuJJOw

PR #3224, opened 2024-08-18
feat(server)!: the endpoint complete campaigns is switched to the post method

AndrusGerman:
## Reason
Solve problem [3117](https://github.com/gophish/gophish/issues/3117) which is to change the endpoint from GET to POST of the route that completes the campaign 

## Changes
* `controllers/api/campaign.go`:  Changed the enpoint method from POST to GET   in  `campaigns/{id:[0-9]+}/complete`
* `static/js/src/app/gophish.js`:  Changed the enpoint method from POST to GET   in `query("/campaigns/" + id + "/complete", "POST", {}, true)`

* Compiled the corresponding JavaScript with Gulp (static/js/dist/app/autocomplete.min.js, static/js/dist/app/campaign_results.min.js, static/js/dist/app/settings.min.js...)

### DIFF
--- a/controllers/api/campaign.go
+++ b/controllers/api/campaign.go
@@ -126,7 +126,7 @@ func (as *Server) CampaignComplete(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	id, _ := strconv.ParseInt(vars["id"], 0, 64)
 	switch {
-	case r.Method == "GET":
+	case r.Method == "POST" || r.Method == "GET":
 		err := models.CompleteCampaign(id, ctx.Get(r, "user_id").(int64))
 		if err != nil {
 			JSONResponse(w, models.Response{Success: false, Message: "Error completing campaign"}, http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


jordan-wright:
For backwards compatibility, I think we should support `GET` and `POST` in the server-side handler, then change the frontend to use `POST` (like you have here).

Then, the release notes for an upcoming release we could say that the `GET` endpoint is deprecated and will be removed in the following release. That way, folks that use the API will have a release to update their code.

Also, if you could please remove the minified JS files, it'd be appreciated! We'll build those ourselves since they're tough to review.

AndrusGerman:
I will be making the corresponding changes 🫡

AndrusGerman:
> For backwards compatibility, I think we should support and in the server-side handler, then change the frontend to use (like you have here).`GET``POST``POST`
> 
> Then, the release notes for an upcoming release we could say that the endpoint is deprecated and will be removed in the following release. That way, folks that use the API will have a release to update their code.`GET`
> 
> Also, if you could please remove the minified JS files, it'd be appreciated! We'll build those ourselves since they're tough to review.

This should be enough 😊 @jordan-wright

PR #3245, opened 2024-09-23
Update workflow actions and Go versions

jordan-wright:
This PR:

* Updates the versions of various actions used by the CI and release workflows
* Updates the release workflow to use Go version 1.22
* Updates the test matrix to use Go versions 1.21, 1.22, and 1.23

It also updates the CI workflow to run when pull requests are created or changed. This will help give feedback when formatting or tests are broken during a PR.

As a good example of why this is useful, you'll see that I needed to run `gofmt` to get this to pass! We should have caught that earlier and now we'll catch it moving forward.


masumbillah000092-ai:
<!DOCTYPE html>
<html>
<head>
<style>
.box{
  width:200px;
  height:200px;
  background:linear-gradient(45deg, #ff4d4d, #4d79ff);
  border-radius:20px;
  margin:50px auto;
}
</style>
</head>
<body>

<div class="box"></div>

</body>
</html> 
     **</html** >

masumbillah000092-ai:
@zjh

PR #3336, opened 2025-02-28
V0.12.1 http proxy sqlite3

Oneiroi:
- adds support for http_proxy per conversation in #212 and #3077 
- bumps node & golang dependency versions (removes go-sqlite3 v2.0.3 as this has been retracted replacing with latest available version).

Replaces #3077


Oneiroi:
Closing encountered some functional issues e.g. login is no longer functional; debugging

*UPDATE 2025-03-05* this is due to running with `use_tls: false` when this is run the Secure flag is not set (which makes sense), unfortunately modern browsers will reject the setting of the authentication cookie when SameSite=None is passed without the Secure flag.

Changing to this use SameSite=Lax instead permits functionality in both use_tls configurations (true or false); whilst this does not have the same overall security as SameSite=Strict this may be a future improvement for now, this permits functionality

PR #5809, opened 2025-07-06
Send_Interval

pranav-r13:
Added send_interval variable, lets user explicitly mention the time interval between 2 consecutive emails in minutes.


FLX-0x00:
min.js files should not be included in PRs.

PR #9279, opened 2025-10-03
Issue 3134 Solution:Add domain to template context

gbw30:
Added the strings import and a new Domain field so phishing templates can reference the recipient’s email domain alongside existing context data. 
Populate the new field by splitting the recipient email before returning the template context, ensuring {{.Domain}} resolves during email and landing page rendering.
Extended the template context unit test to cover the expected domain value.


AndrusGerman:
Te amo y no te conozco 🙂‍↕️💛

